PR #10, opened 2022-07-19
Applitools Test

brocollie08:


.circleci/config.yml
```diff
@@ -188,17 +202,78 @@ jobs:
...
+          command: |
+            wget -q https://github.com/bazelbuild/bazelisk/releases/download/v1.11.0/bazelisk-linux-amd64 -O bazelisk && \
+            chmod +x bazelisk && \
+            sudo mkdir /opt/bazelisk-v1.11.0 && \
+            sudo mv bazelisk /opt/bazelisk-v1.11.0 && \
+            sudo ln -s /opt/bazelisk-v1.11.0/bazelisk /usr/local/bin/bazel
+
+      # This is done because the outputs from the prev build can't be used due to lack of flag
+      # .bazelrc doesn't currently support string interpolation
+      - run: bazel build --define=APPLITOOLS_KEY=${APPLITOOLS_KEY} //android/demo:android_instrumentation_test
```
brocollie08, line 230: @sugarmanz I know you asked for this to be removed, I think I can potentially do something like `bazel test --define...` and it would still build it, but I don't think it's any less work than explicitly building. 

Doing that also would require me to move the install step into the script right before the test, and I'd have to go all the way to the root before finding a path to the APKs for installation because the workspace is different within the script execution (doable but slightly questionable?)

.circleci/config.yml
```diff
@@ -188,17 +202,78 @@ jobs:
...
+      - run:
+          name: Create avd
+          command: |
+            sdkmanager "system-images;android-29;default;x86"
+            echo "no" | avdmanager --verbose create avd -n test -k "system-images;android-29;default;x86"
+
+      - run:
+          name: Launch emulator
+          command: |
+            emulator -avd test -delay-adb -verbose -gpu swiftshader_indirect -no-snapshot -noaudio -no-boot-anim
```
sugarmanz, line 241: 🥇

.circleci/config.yml
sugarmanz: > I think I can potentially do something like bazel test --define... and it would still build it, but I don't think it's any less work than explicitly building.

Not having `--define=APPLITOOLS_KEY=${APPLITOOLS_KEY}` on the `bazel test` command actually causes the test module to be rebuilt.

Screen Shot 2022-07-21 at 6 24 47 PM

So, it should be on both if the commands are done separately.

> Doing that also would require me to move the install step into the script right before the test, and I'd have to go all the way to the root before finding a path to the APKs for installation because the workspace is different within the script execution (doable but slightly questionable?)

I think this is what you were saying about needing the `$BUILD_WORKSPACE_DIRECTORY`. However, I think that's probably an incorrect usage. The script should be giving all the context including the APK paths (I think as part of runfiles?) w/ the `data` params to actually perform the installation dynamically from where it is called.

WORKSPACE
```diff
@@ -136,6 +136,24 @@ maven_aar(
     artifact = "com.eclipsesource.j2v8:j2v8:6.1.0",
 )
 
+maven_aar(
```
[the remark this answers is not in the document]
brocollie08: similar `@aar` like j2v8

android/demo/androidtest.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+adb shell am instrument -w com.intuit.player.android.reference.demo.test/androidx.test.runner.AndroidJUnitRunner
```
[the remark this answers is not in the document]
brocollie08: this has to do with what i mentioned in my comment when i opened the PR, the working directory inside this script is some tmp directory and i'd have to go to the all the way to the root and find the project -> correct path for bazel outputs. Maybe this is relevant to what you posted in slack yesterday

android/demo/src/androidTest/java/com/intuit/player/android/reference/demo/test/ApplitoolsMockTest.kt
```diff
@@ -1,64 +0,0 @@
...
-import com.intuit.player.jvm.utils.mocks.getFlow
-import org.junit.Rule
-import org.junit.Test
-import org.junit.runner.RunWith
-import org.junit.runners.Parameterized
-import java.net.URLEncoder
-
-@LargeTest
-@RunWith(Parameterized::class)
-class ApplitoolsMockTest(val mock: Mock<*>?) : ApplitoolsTest() {
```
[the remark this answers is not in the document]
brocollie08: we talked about removing this a while ago since this only took a screenshot for each mock as it renders and does nothing else, most of those screenshots are duplicates, + increasing the CI time

settings.xml
```diff
@@ -0,0 +1,20 @@
+<settings
```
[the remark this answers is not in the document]
brocollie08: i had to specify this to pull from jcenter because that was the place where i found the applitools stuff

WORKSPACE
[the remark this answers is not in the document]
sugarmanz: Screen Shot 2022-07-21 at 7 51 17 PM
Screen Shot 2022-07-21 at 7 51 42 PM

`maven_install` wasn't able to handle pulling j2v8 b/c the packaging type was wrong in the POM. I believe the packaging type should be correct in the applitool modules.

android/demo/BUILD
```diff
@@ -48,4 +48,39 @@ java_library(
     neverlink = True
 )
 
+applitools_config(
```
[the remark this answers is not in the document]
brocollie08: i'm not sure `$@` is an environment variable, `@` is a pre-defined genrule variable to reference the `outs` file.
I can spend time looking at how to get rid of the `--define` if we really hate it, a lot of the comments in here root from the existence of it

android/demo/BUILD
```diff
@@ -48,4 +48,39 @@ java_library(
...
+    name = "applitools_config",
+    package = "com.intuit.player.android.reference.demo"
+)
+
+sh_binary(
```
sugarmanz, line 56: This is unnecessary, right?

android/demo/scripts/configgen.sh
```diff
@@ -0,0 +1 @@
+
```
sugarmanz, line 1: remove

.circleci/config.yml
```diff
@@ -56,17 +64,23 @@ jobs:
...
       - run: bazel build --remote_header=x-buildbuddy-api-key=${BUILDBUDDY_API_KEY} --config=ci --config=skip-ios -- //...
-     
+
       - save_cache:
-          paths:  
+          paths:
             - node_modules
           key: v1-dependencies-{ { checksum "package.json" }}
 
+      - save_cache:
```
sugarmanz, line 73: I don't think we need to be manually persisting cache anymore with #21

.circleci/config.yml
```diff
@@ -170,17 +184,68 @@ jobs:
...
 
+  android_test:
+    executor: android
+    steps:
+      - checkout
+
+      - attach_workspace:
+          at: ~/player
+
+      - restore_cache:
```
sugarmanz, line 195: That would go for restoring the cache as well

.circleci/config.yml
```diff
@@ -170,17 +184,68 @@ jobs:
...
+          background: true
+
+      - run:
+          name: Wait for emulator to start
+          command: |
+            circle-android wait-for-boot
+
+      - run: bazel test --define=APPLITOOLS_KEY=${APPLITOOLS_KEY} --define=CIRCLE_BUILD_NUM=${CIRCLE_BUILD_NUM} --remote_header=x-buildbuddy-api-key=${BUILDBUDDY_API_KEY} --config=skip-ios --config=ci -- //android/demo:android_instrumentation_test
+
+      - save_cache:
```
sugarmanz, line 229: And saving again

sugarmanz (approved):

PR #15, opened 2022-07-20
Add missing package.json attrs

### LINKED ISSUE #13: Add missing `package.json` attributes to published packages
The published modules contain the basics of what we need to work, but we should add the rest of the metaData:

- Link to git repo
- Contributors (auto-generated from .all-contributorsrc/git log/etc)
- Issues Link
- Docs site
- License

adierkens:
Fixes #13


WORKSPACE
```diff
@@ -9,11 +9,17 @@ workspace(
...
 
-http_archive(
+# http_archive(
+#     name = "rules_player",
+#     strip_prefix = "rules_player-0.3.0",
+#     urls = ["https://github.com/player-ui/rules_player/archive/refs/tags/v0.3.0.tar.gz"],
+#     sha256 = "d92136eb30489f8f337d0c93e93ea45cf865965605f1a16d8fa4d4528a64356e"
+# )
+
+git_repository(
```
KetanReddy, line 19: Any reason in particular to switch from `http_archive` to `git_repository`?

tools/package-template.json
```diff
@@ -0,0 +1,11 @@
+{
```
sugarmanz, line 1: Some of this data I'd like to put in the POM too. Does it make sense to have the package json template be the source of truth? I almost think we could put it in a `.bzl` struct that you can do `json.encode()` on to get this effective file out of. Thoughts?

adierkens: Yeah that works, we can probably defined a `constants.bzl` with a lot of this info

adierkens: Now have a `genrule` to create the base template: https://github.com/player-ui/player/pull/15/files#diff-f354acffd857666c4ea24e57eb1f38b3e9e996ac4a69a3941230c553ec04719eR10

PR #22, opened 2022-07-22
bring over latest jvm code

sugarmanz:


brocollie08 (approved):

sugarmanz:
@adierkens I think this in the `CODEOWNERS` file "incorrectly" added you and @KetanReddy.. not sure if there is a way to specify a folder _only_ at the root

adierkens:
> @adierkens I think this in the `CODEOWNERS` file "incorrectly" added you and @KetanReddy.. not sure if there is a way to specify a folder _only_ at the root

Ah yeah -- we can either remove the initial `/` (to make it `docs/`) or move it closer to the top of the file so it's one of the last things it tries to match

PR #25, opened 2022-07-25
Fix validation namespace collision in compiled module

adierkens:
The namespace only conflicts when using the compiled output (which uses the `Validation` name to the local interface)


KetanReddy (approved):

KetanReddy:
Weird that this didn't cause any issues before.

adierkens:
> Weird that this didn't cause any issues before.

I think this was brought to light b/c of the switch to rollup and `rollup-plugin-dts`

PR #31, opened 2022-07-27
read sha as batch id

sugarmanz:


android/demo/src/androidTest/java/com/intuit/player/android/reference/demo/test/base/ApplitoolsTest.kt
```diff
@@ -23,6 +23,8 @@ abstract class ApplitoolsTest {
         Eyes().apply {
             componentsProvider = AndroidXComponentsProvider()
             configuration = Configuration().apply {
+                appName = "Android Reference Assets"
+                addProperty("platform", "android")
                 batch = batchInfo
                 apiKey = ApplitoolsConfig.APPLITOOLS_KEY
```
brocollie08, line 29: you gonna try removing this in this PR?

sugarmanz: applitools support and i agreed, that since what we have works, it's not entirely worth it to figure out why we can't just pull from the environment.. i thought that maybe it had to do with those env vars not existing within the emulator, but not entirely sure.. regardless, i was a little skeptical anyways

PR #48, opened 2022-08-02
Use .bazelrc.local for CI builds

### LINKED ISSUE #46: Use `.bazelrc.local` for CI bazel configs
Write out a `.bazelrc.local` during the circle build setup that defaults to using the `--config=ci` and the bazel cache write token

adierkens:
Closes #46

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,6 +48,10 @@ jobs:
     steps:
       - checkout
 
+      - run: |
+          echo "common --remote_header=x-buildbuddy-api-key=${BUILDBUDDY_API_KEY}" >> .bazelrc.local
+          echo "common --config=ci" >> .bazelrc.local
+
       - restore_cache:
           keys:
             - v1-dependencies-{ { checksum "package.json" }}
@@ -63,7 +67,7 @@ jobs:
       - attach_workspace:
           at: ~/player
 
-      - run: bazel build --remote_header=x-buildbuddy-api-key=${BUILDBUDDY_API_KEY} --config=ci --config=skip-ios -- //...
+      - run: bazel build --config=skip-ios -- //...
 
       - save_cache:
           paths:
@@ -131,7 +135,7 @@ jobs:
 
       - run: |
           BUNDLE_TARGETS=$(bazel query "attr(name, '_Bundles', //...)" --output label 2>/dev/null | tr '\n' ' ')
-          bazel build --remote_header=x-buildbuddy-api-key=${BUILDBUDDY_API_KEY} --config=ci $BUNDLE_TARGETS
+          bazel build $BUNDLE_TARGETS
 
       - run: bazel shutdown
       - run: cd xcode && bundle exec pod install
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -52,6 +52,19 @@ jobs:
           keys:
             - v1-dependencies-{ { checksum "package.json" }}
 
+      - run: |
+          echo "build --remote_header=x-buildbuddy-api-key=${BUILDBUDDY_API_KEY}" >> .bazelrc.local
```
adierkens, line 56: @sugarmanz Do you know of a way to specify args for any bazel cmd? Instead of having to split it up into `build`, `run`, `test`, etc?

sugarmanz: I think you can do:

```suggestion
          echo "common --remote_header=x-buildbuddy-api-key=${BUILDBUDDY_API_KEY}" >> .bazelrc.local
```

scripts/release.sh
```diff
@@ -20,7 +20,7 @@ elif [ "$RELEASE_TYPE" == "release" ] && [ "$CURRENT_BRANCH" == "main" ]; then
 fi
 
 for pkg in $PKG_NPM_LABELS ; do
-  bazel run --config=release -- ${pkg}.publish --access public --tag ${NPM_TAG}
+  bazel run -- ${pkg}.publish --access public --tag ${NPM_TAG}
```
sugarmanz, line 23: Is it expected that the `--config=release` flag will always be remembered? I'm picturing a scenario where someone wants to publish locally and forgets that flag. Or, do we want to just say publishing is _always_ done on CI? Maybe even have some condition at the top of this script that checks for some CI-only env var and fails our if not? Seems like maybe overkill, but just a thought I had.

.circleci/config.yml
```diff
@@ -52,6 +52,11 @@ jobs:
           keys:
             - v1-dependencies-{ { checksum "package.json" }}
 
+      - run: |
+          echo "common --remote_header=x-buildbuddy-api-key=${BUILDBUDDY_API_KEY}" >> .bazelrc.local
+          echo "common --config=ci" >> .bazelrc.local
+          echo "test --define=APPLITOOLS_API_KEY=${APPLITOOLS_API_KEY} --define=APPLITOOLS_BATCH_ID=${CIRCLE_SHA1}" >> .bazelrc.local
```
sugarmanz, line 58: I think this, as-is, is problematic since it would apply those flags on _every_ `test` invocation, causing some rebuild/reanalysis since those flags weren't defined during `build`. So, either we'd need to apply these flags on `common` or leave them on the specific Android test command so that is the only place that eats the perf impact. I'm trying to think of a reason to not have it on `common`, the only thing I can think of is whether or not that would invalidate remote caching? Or maybe that'd just be on the first time and subsequent builds with that flag would be able to pull from the cache properly? Thoughts?

scripts/release.sh
adierkens: I could see people running individual commands if needed, but I wouldn't envision anyone running the entire `release` script locally. 

Enforcing that from a CI env var isn't a bad idea

.circleci/config.yml
adierkens: I think that could also mess with the ability to share the cache between local and remote, since none of the remote builds would be sharable for pulling down to your local environment

adierkens: Moving it back to the android call might be the safest

hborawski:
Can we close this? I believe this was covered in #97

PR #50, opened 2022-08-04
enhance error state deserialization & stacktrace logging in player view model

sugarmanz:

### DIFF
--- a/jvm/core/src/main/kotlin/com/intuit/player/jvm/core/player/state/PlayerFlowState.kt
+++ b/jvm/core/src/main/kotlin/com/intuit/player/jvm/core/player/state/PlayerFlowState.kt
@@ -99,8 +99,13 @@ internal class ErroneousState(override val node: Node) :
 
     override val flow: Flow get() = node.getSerializable("flow", Flow.serializer()) ?: Flow()
 
-    override val error: PlayerException get() = node.getSerializable("error", PlayerException.serializer())
-        ?: PlayerException(node.getString("error") ?: "unable to determine error")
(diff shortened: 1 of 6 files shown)


jvm/core/src/main/kotlin/com/intuit/player/jvm/core/player/state/PlayerFlowState.kt
```diff
@@ -99,8 +99,13 @@ internal class ErroneousState(override val node: Node) :
 
     override val flow: Flow get() = node.getSerializable("flow", Flow.serializer()) ?: Flow()
 
-    override val error: PlayerException get() = node.getSerializable("error", PlayerException.serializer())
-        ?: PlayerException(node.getString("error") ?: "unable to determine error")
+    override val error: PlayerException get() = when (val rawError = node["error"]) {
+        is PlayerException -> rawError
+        is Exception -> PlayerException(rawError.message ?: "", rawError)
```
brocollie08, line 104: out of curiosity, when do these two cases come into play

brocollie08 (approved):

jvm/core/src/main/kotlin/com/intuit/player/jvm/core/player/state/PlayerFlowState.kt
sugarmanz: It's more for completeness than anything. I actually think that Graal works this way and is able to spit up the original exception, which may or may not be a `PlayerException`, but that could be wrong. If `Node` internals change in the future with some introspection for `Node` children that are error-like and try to deserialize automatically, this will account for that.

PR #51, opened 2022-08-04
added templates to content navigation, fixed the link to reference as…

mercillo:
added templates to content navigation, fixed the link to reference assets from the asset plugin

Documentation Updates


adierkens (approved):
Looks like we need to change some circle configs to be able to build forks

Issue #47, opened 2022-08-02
Docs Infra Improvements


adierkens:
- Add version switch to docs (`latest`, `v0`, `next`) 
- Fix path stamping for non-next builds ([this](https://github.com/player-ui/player/blob/main/docs/site/BUILD#L54))

PR #54, opened 2022-08-06
Add subpath support for published docs

adierkens:
Fixes: #47

Re-organizes how the docs site is merged together. Now creates a `.tar.gz` of the full merged site, stamps that result to a new `.tar.gz`, and uploads that to github-pages. 

The merged `.tar.gz` is stamped twice depending on it's output directory (`/latest`, `/next`, `/<version>`)

### Version Selection

Adds a menu selection for `latest`, `next` and any released version

CleanShot 2022-08-08 at 11 05 32

### DIFF
--- a/docs/site/components/Navigation.tsx
+++ b/docs/site/components/Navigation.tsx
@@ -212,6 +212,7 @@ export const VersionSelector = () => {
       rootProps={{
         width: 'auto',
         display: 'flex',
+        flexShrink: '0',
       }}
       value={router.basePath || 'latest'}
       onChange={(e) => {
@@ -263,7 +264,13 @@ export const TopNavigation = () => {
             onClick={mobileNavDisclosure.onOpen}
           />
           <Link passHref href="/">
-            <CLink py="2">
+            <CLink
+              display={{
+                base: 'none',
+                md: 'block',
+              }}
+              py="2"
+            >
               <Image alt="Player Logo" height="48px" src={logoSrc} />
             </CLink>
           </Link>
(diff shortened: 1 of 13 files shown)


docs/site/components/Navigation.tsx
```diff
@@ -169,6 +170,65 @@ export const GitHubButton = () => {
...
+
+  React.useEffect(() => {
+    const send = async () => {
+      const response = await fetch(
+        'https://api.github.com/repos/player-ui/player-ui.github.io/contents/'
+      );
+
+      const data = await response.json();
+      const versions = data
+        .filter((d) => d.type === 'dir' && d.name.match(/^v\d$/))
```
KetanReddy, line 189: Shouldn't this regex match any name that is a valid semvar version? Or are we organizing the docs folders differently?

adierkens: The docs are published by major -- so `v0`, `v1`,  etc 
I should fix it to handle `vXX` though

adierkens: The release folder is determined from 
https://github.com/player-ui/player/blob/900e3c6c11fc45127cd6252da40476e44ef5b345/scripts/release.sh#L57

KetanReddy: Yeah, minor/patch versions were going to be my followup question Also not sure where we landed on publishing docs for release candidates and canaries.

adierkens: Anything merged to `main` (effectively an RC) goes under `next` -- we don't have anything for PR canaries yet

PR #57, opened 2022-08-11
add local auto plugin to run bazel pod push after release

hborawski:
need to point to final zip location since `pod trunk push` resolves the ruby podspec into a json format for their CDN (not documented), so the whole purpose of having it be ruby files is negated

add a local script to push after the zip has been added to the release


sugarmanz (approved):
Looks good to me!

PR #76, opened 2022-08-30
Add algolia search to /latest/ site

adierkens:


docs/site/components/AlgoliaSearch.tsx
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable import/no-unassigned-import */
+import React from 'react';
+import { DocSearch } from '@docsearch/react';
+
+import '@docsearch/css';
+
+export const AlgoliaSearch = () => {
+  return (
+    <DocSearch
+      appId="OX3UZKXCOH"
```
sugarmanz, line 10: should this be an env var?

adierkens: It can definitely be converted it to one; I didn't see it changing often, and this (including the apiKey) could be extracted from the JS bundle anyway

KetanReddy (approved):

sugarmanz (approved):

PR #77, opened 2022-08-31
Cleanup Android dependencies

sugarmanz:


plugins/reference-assets/android/deps.bzl
```diff
@@ -1,20 +1,17 @@
...
-    "androidx.constraintlayout:constraintlayout:2.0.4",
-]
+maven_main = []
 
 main_exports = [
     "//android/player",
 ]
 
-main_deps = main_exports + parse_coordinates(maven) + [
+main_deps = main_exports + parse_coordinates(maven_main) + [
```
brocollie08, line 11: since `maven_main` doesn't have anything in it right now, maybe remove the `parse_coordinates`, i'm not against keeping the `maven_main` though

android/player/deps.bzl
```diff
@@ -2,15 +2,17 @@ load("//jvm/dependencies:versions.bzl", "versions")
...
-    "androidx.core:core-ktx:1.3.2",
-    "androidx.constraintlayout:constraintlayout:2.0.4",
-    "androidx.navigation:navigation-fragment-ktx:2.3.3",
-    "androidx.navigation:navigation-ui-ktx:2.3.3",
-    "androidx.transition:transition:1.4.1",
+    # Lifecycle
+    "androidx.lifecycle:lifecycle-runtime-ktx:%s" % versions.androidx.lifecycle,
+    "androidx.lifecycle:lifecycle-viewmodel-ktx:%s" % versions.androidx.lifecycle,
+
+    # Default fallback
```
brocollie08, line 14: what's constraintlayout gotta do with default fallback? you mean the fallbackView? is that the only place we use it..?

jvm/dependencies/versions.bzl
```diff
@@ -13,9 +13,34 @@ versions = struct(
...
+        lifecycle = "2.4.0",
+        navigation = "2.3.3",
+        test = struct(
+            core = "1.4.0",
+            espresso = "3.3.0",
+            junit = "1.1.3",
+        ),
+        transition = "1.4.1",
+    ),
+    dagger = "2.35.1",
```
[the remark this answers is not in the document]
sugarmanz: dagger is needed to load the grab common repository -- at least it was when i initially integrated. it'd be cool to get off that repo eventually

plugins/reference-assets/android/deps.bzl
sugarmanz: This is more for the patternization of `deps.bzl`. If someone wants to add another dependency, all they need to do right now is add `"androidx.core:core-ktx:%s" % versions.androidx.core" and that will automatically be pulled into the `maven_install` _and_ converted to the corresponding `@maven//:androidx_core_core_ktx` Bazel reference.

If we remove the `parse_coordinates`, then they'd need to do:
> converted to the corresponding `@maven//:androidx_core_core_ktx` Bazel reference.
manually, or reintroduce the `parse_coordinates` method.

android/player/deps.bzl
sugarmanz: I would love to not have a dependency on `constraintlayout` in the base android player dep, but we are using it to build the default fallback view:

https://github.com/player-ui/player/blob/main/android/player/src/main/res/layout/default_fallback.xml#L2

PR #80, opened 2022-09-08
Enhancement/Parser Rewrite for plugins

mercillo:
Changed the parsers to be built around the plugins. The plugins should determine what the nodetypes are for template, applicability and switches. Based on that nodetype, they should parse their own respective nodes.


KetanReddy:
In general I think this should be fine, it does look like there are some cases not covered by tests. Can we get a few more tests just to cover those?

mercillo:
> In general I think this should be fine, it does look like there are some cases not covered by tests. Can we get a few more tests just to cover those?

added some more unit tests!

hborawski:
applitools diff looks to be a view presentation timing diff, i added a delay in iOS before calling applitools, i can do it for android as well

core/view/src/__tests__/view.test.ts
```diff
@@ -1,4 +1,4 @@
-import { LocalModel, withParser } from '@player-ui/data';
+import { DataModelWithParser, LocalModel, withParser } from '@player-ui/data';
```
adierkens, line 1: Looks like this is imported but not used?

core/view/src/parser/__tests__/parser.test.ts
```diff
@@ -1,7 +1,38 @@
...
+      schema: new SchemaController(),
+      data: {
+        format: (binding, val) => val,
+        formatValue: (val) => val,
+        model,
+      },
+    };
+    new TemplatePlugin(options).applyParserHooks(parser);
+    new ApplicabilityPlugin().applyParser(parser);
+    new SwitchPlugin(options).applyParser(parser);
```
adierkens, line 34: Why is `TemplatePlugin` the only one that has a different API? `applyParserHooks` vs. `applyParser`

core/view/src/parser/index.ts
```diff
@@ -146,36 +77,54 @@ export class Parser {
...
+      );
+      if (parsedNode) {
+        return parsedNode;
+      }
     }
 
-    if (hasSwitch(obj)) {
-      return this.parseSwitch(obj, options);
-    }
+    interface Obj {
```
adierkens, line 94: Is there a more descriptive name then `Obj` as a type we can use?

adierkens: Is there a specific reason this is defined inside of the `parseObject` method instead of a top-level interface?

core/view/src/parser/index.ts
```diff
@@ -251,35 +225,71 @@ export class Parser {
...
+                  },
+                ],
+              };
             }
           }
         } else if (localValue && typeof localValue === 'object') {
-          if (hasApplicability(localValue)) {
-            const applicabilityNode = this.parseApplicability(
+          const determineNodeType =
+            this.hooks.determineNodeType.call(localValue);
```
adierkens, line 242: I've noticed that `determineNodeType` alternates between being called w/ a _value_ (`localValue` here) and also w/ just the key name ([here](https://github.com/player-ui/player/pull/80/files#diff-96a718b00b84d926bdfa59f06fbc95d051ee90f066adb5b3bc742ffdf3299092R148))

core/view/src/parser/index.ts
```diff
@@ -146,36 +77,54 @@ export class Parser {
...
-    let value: any;
-    const children: Node.Child[] = [];
+      value: any;
+    }
 
-    /** Parse a nested child and add it to the parent */
-    const parseLocalObject = (objToParse: unknown, path: string[] = []) => {
+    const parseLocalObject = (
+      currentValue: any,
+      objToParse: unknown,
```
adierkens, line 102: What's the difference between `currentValue` and `objToParse`?

core/view/src/parser/index.ts
```diff
@@ -44,6 +43,13 @@ export class Parser {
     onCreateASTNode: new SyncWaterfallHook<
       [Node.Node | undefined | null, object]
     >(),
+
+    determineNodeType: new SyncBailHook<[object | string], NodeType>(),
+
+    parseNode: new SyncBailHook<
+      [object | null, Node.ChildrenTypes, ParseObjectOptions, NodeType | null],
```
adierkens, line 50: In what scenario would the determinedNodeType be `null`?

adierkens, line 50: For hook types with longer definitions, you can name the arguments to keep track a bit easier:

```suggestion
      [
        obj: object | null,
        nodeType: Node.ChildrenTypes, 
        parseOptions: ParseObjectOptions, 
        determinedNodeType: NodeType | null
        ],
```

mercillo:
looks like the codecov patch is hitting the compiler.ts file again which can be ignored haha

core/view/src/parser/index.ts
mercillo: there are cases when the current value is just a value with nothing to parse, so it doesnt contain a children element on the return

PR #93, opened 2022-10-21
Core package organization, and renaming to `ReactPlayer`

### LINKED ISSUE #87: Simplify package layout for JS libraries
In large projects, managing the transitive Player libraries (binding, expression, etc) can be frustrating when `yarn` or `npm` chooses to pull in a different version. 

Moving to _just_ exposing `@player-ui/player` and `@player-ui/react` as the public facing packages (and bundling the rest of core) should help reduce that churn.

adierkens:
The majority of this PR merges most of the sub-packages inside the `core/` folder under the `core/player` package. Most of the API of the `core/player` remains unchanged (as many of these modules were exported under `@player-ui/player` anyway). 

The big motivation for reducing the number of transitive dependencies is to reduce the churn for updating versions. Many times the transitive dependencies don't get properly resolved/shared without manual editing of the lockfiles generated by `yarn`/`npm`. This should help to reduce that variability down to just `@player-ui/player` and `@player-ui/react`. 

This PR also does the same thing for `@player-ui/react`, which now includes the export for the `ReactAsset` component to render in the player tree. In order to avoid conflict w/ the `Asset` interface from `@player-ui/types`, the `Asset` component was renamed to `ReactAsset`, along with the entire `WebPlayer` to `ReactPlayer` to better describe it's platform dependence. `@player-ui/react` no longer requires the use of `react-dom` 

--- 
Fixes #87


sugarmanz:
I know I'm late to this PR, but it probably still makes sense to keep separate `BUILD` targets for the sub-modules to help maintain incremental build performance.

PR #96, opened 2022-10-26
Update documentation

KVSRoyal:
* Prevent infinite load of PlatformTabs. These were trying to use `router.isReady` which shouldn't be used outside of `useEffect` ([link](https://nextjs.org/docs/api-reference/next/router#router-object)). This causes a progress bar to appear until the user navigates to a different page and then navigates back (e.g. for the [Asset Provide Plugin](https://player-ui.github.io/latest/plugins/asset-provider)), so remove that.
* The validations objects link under the [Cross-field Validation section in Content/Assets & Views](https://player-ui.github.io/latest/content/assets-views#cross-field-validation) was broken. Add a link to a relevant validations page.
* Use reliable local images for the Player logo and single/simple flow example.
* Update the contributing documentation with the info I gathered while working on this so others don't need to.


hborawski:
@KVSRoyal when you get a chance, rebase against `main`, fork build should work after #97

docs/site/pages/content/assets-views.mdx
```diff
@@ -45,7 +45,7 @@ The `id` of the views are used in the navigation section to reference a specific
 
 ## Cross-field validation
 
-The other special property of a `view` vs. an `asset` is the addition of a `validation` property on the view. These contains [`validation` objects](../guides/validation) that are used for validations crossing multiple fields, and are ran on user navigation rather than data change.
+The other special property of a `view` vs. an `asset` is the addition of a `validation` property on the view. These contain [`validation` objects](../plugins/common-types#validations) that are used for validations crossing multiple fields, and are ran on user navigation rather than data change.
```
hborawski, line 48: i think this should be referencing https://github.com/player-ui/player/blob/main/docs/site/pages/content/schema.mdx#validation rather than validations from a plugin

CONTRIBUTING.md
```diff
@@ -8,9 +8,15 @@ For small bug-fixes, documentation updates, or other trivial changes, feel free
 
 If the changes are larger (API design, architecture, etc), [opening an issue](https://github.com/player-ui/player/issues/new/choose) can be helpful to reduce implementation churn as we hash out the design.
 
-## Building and Testing Locally
+## Requirements
+* [npm >= 8.19.2](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
+* [yarn >= 1.22.19](https://yarnpkg.com/)
+* [Swift >= 5.2](https://www.swift.org/download/)
+* [Android NDK = 19.2.5345600](https://github.com/android/ndk/wiki/Unsupported-Downloads#r19c). You'll need to add `ANDROID_NDK_HOME` to your environment manually.
```
brocollie08, line 15: We might need to explicitly call out the incompatibility with NDK version >21

hborawski: do we need bazel 6.x to bump up the android ndk?

docs/site/pages/content/assets-views.mdx
KVSRoyal: It doesn't seem like I can do a relative link for this (`../content/schema#validation`). Is it fine if I use the full link: `https://github.com/player-ui/player/blob/main/docs/site/pages/content/schema.mdx#validation`?

CONTRIBUTING.md
KVSRoyal: How's this?

> * [Android NDK >= 19.2.5345600, <= 21](https://github.com/android/ndk/wiki/Unsupported-Downloads#r19c). Any version > 22 will not work, period. You'll need to add `ANDROID_NDK_HOME` to your environment manually.

brocollie08: I looked around, it looks like it's not supported yet.. there's [this](https://github.com/bazelbuild/bazel/pull/14077) PR but it's been up for a while and according to [this](https://github.com/bazelbuild/bazel/issues/12889) thread, that will no longer be the solution going forward, and there's supposed to be a `rules_android_ndk` that supports NDK r25. I don't know if that'll require bazel 6.x atm

docs/site/pages/content/assets-views.mdx
hborawski: can you just do `./schema#validation` since its also in the `content` folder?

KVSRoyal: Oh yeah, sorry. My bad. That also doesn't work.

CONTRIBUTING.md
brocollie08: https://github.com/ahumesky/rules_android_ndk
this specifically says version 25, so i'm assuming 22, 23, 24 don't work lol

hborawski (approved):

docs/site/components/mdx-components.tsx
```diff
@@ -101,10 +101,6 @@ const PlatformTabs = (props: React.PropsWithChildren<unknown>) => {
     CodeTabsNameMap.has(c.props.mdxType.toLowerCase())
   );
 
-  if (!router.isReady) {
```
KetanReddy, line 104: Good catch! I am trying to remember the reason for including this catch but it looks like its not needed anymore.

PR #98, opened 2022-10-27
Test PR: test bazelrc.local being populated from restricted context

hborawski:
same as #97 to make sure builds from the main repo work as intended


hborawski:
closed in #97

PR #105, opened 2022-11-28
Check path plugin in rust wasm

cleverjam:


plugins/check-path/core-rs/src/paths.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+impl Paths {
+    pub fn new() -> Self {
+        Self {
+            store: Rc::new(RefCell::new(HashMap::new())),
+        }
+    }
+
+    pub fn parse(&self, data: JsValue) {
```
Brekcel, line 24: What's the point of this function? Why not inline traverse into it, or remove it and make traverse pub?

plugins/check-path/core-rs/src/paths.rs
```diff
@@ -0,0 +1,84 @@
...
+        value: &JsValue,
+        path: Vec<String>,
+    ) -> Option<Vec<(String, NodeValue, Vec<String>)>> {
+        if !value.is_object() {
+            return None;
+        }
+        let keys = Reflect::own_keys(&value).unwrap();
+        let len = keys.length();
+
+        let mut result = Vec::with_capacity(len as usize);
```
Brekcel, line 64: You can use `keys.iter()` to get a Rust iterator over the array, and then map the keys and return that iterator

plugins/check-path/core-rs/src/paths.rs
```diff
@@ -0,0 +1,84 @@
...
+            .unwrap()
+            .clone()
+    }
+
+    fn traverse(&self, root: JsValue) {
+        let path: Vec<String> = vec![];
+        let mut stack =
+            Paths::get_key_values(&root, path.clone()).expect("Couldn't read root node.");
+        while let Some((key, value, path)) = stack.pop() {
+            if let NodeValue::StringType(value) = value {
```
Brekcel, line 42: Use a match instead of multiple if lets

plugins/check-path/core-rs/src/paths.rs
```diff
@@ -0,0 +1,84 @@
...
+                object_path.push(key);
+                stack.append(Paths::get_key_values(&value, object_path).unwrap().as_mut());
+            }
+        }
+    }
+
+    fn get_key_values(
+        value: &JsValue,
+        path: Vec<String>,
+    ) -> Option<Vec<(String, NodeValue, Vec<String>)>> {
```
Brekcel, line 57:
```suggestion
    ) -> impl Iterator<Item=(String, NodeValue, Vec<String>)> {
```
Or something like this.

Then return the iterator instead of collecting it into a vec.

plugins/check-path/core-rs/src/paths.rs
```diff
@@ -0,0 +1,84 @@
...
+                let mut object_path = path.clone();
+                object_path.push(key);
+                stack.append(Paths::get_key_values(&value, object_path).unwrap().as_mut());
+            }
+        }
+    }
+
+    fn get_key_values(
+        value: &JsValue,
+        path: Vec<String>,
```
Brekcel, line 56:
```suggestion
        path: &[String],
```

plugins/check-path/core-rs/src/paths.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    pub fn parse(&self, data: JsValue) {
+        self.traverse(data);
+    }
+
+    pub fn get(&self, key: &str) -> Vec<String> {
+        self.store
+            .borrow()
+            .get(key)
+            .or(Some(&vec!["".to_string()]))
```
Brekcel, line 32:
```
            .cloned()
            .unwrap_or_else(|| vec![String::new()])
```
and remove the clone at the end.
Does this need to return a vec with an empty string in it? Or is just an empty vec enough?

plugins/check-path/core-rs/src/paths.rs
```diff
@@ -0,0 +1,84 @@
...
+            .unwrap()
+            .clone()
```
Brekcel, line 34: Also look into [Ref::map](https://doc.rust-lang.org/std/cell/struct.Ref.html#method.map) to see if you can return a borrowed value instead of cloning

Brekcel (changes requested):

WORKSPACE
```diff
@@ -181,3 +181,57 @@ maven_install(
...
+load("@rules_rust//crate_universe:defs.bzl", "crate", "crates_repository", "splicing_config")
+
+
+crates_repository(
+    name = "crate_index",
+    cargo_lockfile = "//plugins/check-path/core-rs:Cargo.lock",
+    # `generator` is not necessary in official releases.
+    # See load statement for `cargo_bazel_bootstrap`.
+#    generator = "@cargo_bazel_bootstrap//:cargo-bazel",
+    manifests = ["//plugins/check-path/core-rs:Cargo.toml"],
```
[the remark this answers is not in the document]
cleverjam: I was thinking since we are wrapping the WASM binary in an NPM package for publishing the published version would only need to be in `package.json` and not in `Cargo.toml`. do we want to keep the `Cargo.toml` version up to date though, even though the crate won't be published? 

One option to avoid confusing manifests mismatched versions is to remove the `Cargo.toml` altogether and use a rust rule to define all the dependencies directly [Like this](https://github.com/bazelbuild/rules_rust/blob/main/examples/crate_universe/WORKSPACE.bazel#L218-L222)

plugins/check-path/core-rs/src/paths.rs
cleverjam: empty string is expected in the consuming code - JS tool does that

plugins/check-path/core-rs/src/paths.rs
cleverjam: If I understand the docs, at a quick glance (so probably not really getting it), I would have to wrap up the vector in an RefCell in order to get a `Ref<Vec<String>>` from ` self.key_paths.borrow().get(key).borrow()` ??

plugins/check-path/core-rs/src/paths.rs
cleverjam: just got rid of this function `get_key_values`, iterating through all keys here just to add to stack to iterate again was not the best approach.

plugins/check-path/core-rs/src/paths.rs
cleverjam: When I change the return type to a borrow of the `Vec<String>` I get the whole borrow of value owned in this function, I think I need some lifetime annotation or something here I think... idk yet, still lots to figure out in this file, but definitely will dig more into this.

cleverjam: I think I improved this significantly by wrapping the store hash map in a reference counter pointer `Rc`

plugins/check-path/core-rs/package.json
```diff
@@ -0,0 +1,11 @@
...
+  "name": "@player-ui/check-path-plugin-rs",
+  "main": "index.js",
+  "module": "index.js",
+  "dependencies": {
+    "@player-ui/asset-transform-plugin": "^0.3.0",
+    "@player-ui/check-path-plugin": "^0.3.0",
+    "@player-ui/make-flow": "^0.3.0",
+    "@player-ui/player": "^0.3.0"
+  }
+}
```
KetanReddy, line 11: These should probably be linked to the local packages similar to how they are pulled in for other inter-project dependencies.

cleverjam: good catch, will move these over to the `BUILD` file.

PR #117, opened 2023-01-03
Expose utils for expression LSP

adierkens:
## Release Notes
- Expose `parseExpression` from the expression parser
- Adds a `strict` flag to `parseExpression` that will (when disabled) attach any parsing errors to the top-level node instead of throwing

### DIFF
--- a/core/player/src/expressions/parser.ts
+++ b/core/player/src/expressions/parser.ts
@@ -797,6 +797,10 @@ export function parseExpression(
       args.push(node);
     }
 
+    if (charIndex !== termination) {
+      throwError(`Expected ${String.fromCharCode(termination)}`, index);
+    }
+
     return args;
   }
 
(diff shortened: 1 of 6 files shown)


core/player/src/expressions/parser.ts
```diff
@@ -899,37 +911,51 @@ export default function parseExpression(expr: string): ExpressionNode {
...
 
-  while (index < length) {
-    const chIndex = exprICode(index);
+  try {
+    while (index < length) {
+      const chIndex = exprICode(index);
+
+      // Expressions can be separated by semicolons, commas, or just inferred without any
+      // separators
+      if (chIndex === SEMCOL_CODE || chIndex === COMMA_CODE) {
```
KetanReddy, line 920: Would there be any worth to having the kind of characters that are ignored be configurable?

adierkens: That may make it pretty difficult to define a proper bnf grammar for expressions. But that might be worth considering as a future change

PR #123, opened 2023-01-19
dont try to push pod if its a canary release

hborawski:
Added a flag to track if the canary hook is called, rather than spend time polling when we know there will be no tag/release

<details>
  <summary>📦 Published PR as canary version: <code>0.3.1--canary.123.4401</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.3.1--canary.123.4401
</details>

### DIFF
--- a/scripts/after-shipit-pod-push.js
+++ b/scripts/after-shipit-pod-push.js
@@ -24,10 +24,13 @@ class AfterShipItPodPush {
         while (!found && attempt < 10) {
           const { data } = await auto.git.github.repos.listReleases({owner: auto.config.owner, repo: auto.config.repo})
           const release = data.find(element => element.tag_name === newVersion)
-          const { data: releaseData } = await auto.git.github.repos.getRelease({owner: auto.config.owner, repo: auto.config.repo, release_id: release.id})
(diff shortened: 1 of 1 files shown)


scripts/after-shipit-pod-push.js
```diff
@@ -6,10 +6,17 @@ const { execSync } = require('child_process');
 class AfterShipItPodPush {
   name = 'after-shipit-pod-push';
 
+  wasCanary = false;
+
   apply(auto) {
+    auto.hooks.canary.tap(this.name, () => {
```
sugarmanz, line 12: This looks good to me. Only thought is if it would make sense just to early return if we `release` is `undefined` and log? Just trying to think if there are any other use cases where `afterShipIt` would be invoked but not have an actual release.

hborawski: AFAIK theres only 3 flows, release, prerelease and canary? and the first two we do have tags, so if we get the canary hook called, i can skip even hitting the release API since i know it wont have a tag

sugarmanz (approved):

PR #136, opened 2023-01-30
extend AnyType cases for other primitives

hborawski:
Extends `AnyType` for additional primitive data structures, numbers and booleans and arrays + dictionaries of numbers/bools


hborawski:
covering this in an upcoming PR

Issue #181, opened 2023-10-17
Core: New Reference Asset - Choice


hborawski:
To reduce the time it takes for new users to evaluate Player, we need to expand the reference asset set to incoporate more common user interaction patterns.

A `choice` asset represents a single selection choice, often displayed as radio buttons in a web context. This will allow users to test out more complex flows than just inputs + buttons.

## Requirements

`choice`

- A `choice` asset must have a `title` to give the user context on what choice they are making
  - The `title` should be an asset
- A `choice` asset can have an optional `note` to provide more detailed information about the selection
  - The `note` should be an asset
- A `choice` asset should update the value of a `binding` property supplied to it
- A `choice` asset should retrieve `validation` information for its binding, and expose that for all rendering platforms to display
- A `choice` asset will have an array of `choices`
  - Each entry in `choices` will not be an asset, as the `choice` assets per platform will need direct control over the layout
  - Each entry must have a `value` that is set to the `binding` when the option is selected
    - The `value` should be a primitive
  - Each entry must have a `label` that describes what value is being selected
    - The `label` should be an asset

## Execution

1. Create a new `choice` asset in the [core reference assets](https://github.com/player-ui/player/tree/main/plugins/reference-assets/core/src/assets) and add types for the authored content, and the transformed content.
2. Create a transform for this `choice` asset
   1. The transform should use the `binding` and provide functions on the transformed type to make it simple for the rendering platform to select an option, similar to `run` in the `action` asset
   2. The transform should populate a property on each choice to indicate its selection status
   3. `validation` to be populated from the binding/schema information
      1. `validation` is not a content authored field
3. Add mocks to the [mocks package](https://github.com/player-ui/player/tree/main/plugins/reference-assets/mocks) to exercise required and optional fields, to ease unit testing, and implementations for rendering
4. Add the [DSL Component](https://github.com/player-ui/player/blob/main/plugins/reference-assets/components/src/index.tsx) for `choice` to allow programmatic content authoring
   1. [DSL Docs](https://player-ui.github.io/latest/tools/dsl)
   2. Add [unit test](https://github.com/player-ui/player/blob/main/plugins/reference-assets/components/src/index.test.tsx) to verify JSON conversion of the DSL component
5. 
   Add unit tests for the transform
   1. Verify selecting options
   2. Verify validation information being populated

LeonDong02:
Hey, can I take on this task?

PR #185, opened 2023-10-17
Core: New Reference Asset - Choice

LeonDong02:
### What changed?

Added core support for the new reference asset - choice.

(From [issue](https://github.com/player-ui/player/issues/181)) "A choice asset represents a single selection choice, often displayed as radio buttons in a web context. This will allow users to test out more complex flows than just inputs + buttons."

### Todo

- Unit tests for transform (verify selecting options, verify validation information being populated)
- Confirm json result in DSL tests (auto generated id names)
- Fix DSL for entries

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`


plugins/reference-assets/core/src/assets/choice/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  choices: Array<ChoicesEntry>
+
+  /** Optional additional data */
+  metaData?: {
+    /** Additional data to beacon when this input changes */
+    beacon?: BeaconDataType;
+  };
+}
+
+type ChoicesEntry<AnyTextAsset extends Asset = Asset> = {
```
hborawski, line 29: you will need a transformed version of this as well, to populate a property indicating whether or not it is currently selected

plugins/reference-assets/core/src/assets/choice/types.ts
```diff
@@ -0,0 +1,49 @@
...
+
+type ValueType = string | undefined;
+
+/** A stateful instance of an action */
+export interface TransformedChoice extends ChoiceAsset {
+  /** A function to commit the new value to the data-model */
+  set: (choiceValue: ValueType) => void;
+
+  /** The current value of the input from the data-model */
+  value: ValueType;
```
hborawski, line 45: The choice asset itself does not need to have a `value` associated with it, this looks like carryover from the `input` transform?

LeonDong02: Yea 😅, forgot that there's no value to show in the choice asset itself.

plugins/reference-assets/core/src/assets/choice/transform.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Choice asset transform
+ */
+export const choiceTransform: TransformFunction<ChoiceAsset, TransformedChoice> = (
+  asset,
+  options
+) => {
+  return {
+    ...asset,
+    set(val) {
```
hborawski, line 14: we don't need a `set` function for the whole choice itself, if the binding was updating from another expression, this would not get called, and the selection state would not update.


each individual `ChoiceEntry` should have a `select` function that sets its value into the data model for the binding

```typescript
const currentValue = options.data.model.get(binding)
const transformedChoices = asset.choices.map(choiceEntry => {
  return {
    ...choiceEntry,
    selected: choiceEntry.value === currentValue,
    select: () => {
        options.data.model.set([asset.binding, choiceEntry.value])
    }
  }
})

return {
  ...asset,
  choices: transformedChoices,
  // other properties like validation
}
```

plugins/reference-assets/core/src/assets/choice/choices-entry/transform.ts
```diff
@@ -0,0 +1,17 @@
+import type { ChoicesEntry, TransformedChoicesEntry } from './types';
+
+/**
+ * Choice asset transform
+ */
+export const choicesEntryTransform: (entry: ChoicesEntry) => TransformedChoicesEntry = (
```
hborawski, line 6: I don't think this is needed, and can be handled in the main transform, to ensure proper scope to capture the `binding` for adding the `select` function, and `selected` property

plugins/reference-assets/core/src/assets/choice/transform.ts
```diff
@@ -0,0 +1,33 @@
...
+import { choicesEntryTransform } from './choices-entry'
+
+/**
+ * Choice asset transform
+ */
+export const choiceTransform: TransformFunction<ChoiceAsset, TransformedChoice> = (
+  asset,
+  options
+) => {
+  return {
```
hborawski, line 12: It seems like you're limiting your options by trying to keep this transform to a single `return` statement

plugins/reference-assets/core/src/assets/choice/transform.ts
```diff
@@ -0,0 +1,33 @@
...
+          entry.set(false);
+        }
+      });
+      return options.data.model.set([[asset.binding, val]]);
+    },
+    validation:
+      asset.binding === undefined
+        ? undefined
+        : options.validation?.get(asset.binding, { track: true }),
+    transformedChoices: asset.choices.map(choicesEntryTransform)
```
hborawski, line 31: we should be transforming the `choices`, not providing a new `transformedChoices`

hborawski (changes requested):

plugins/reference-assets/core/src/assets/choice/transform.ts
LeonDong02: Just curious how this would work since I would still need to set the previously selected value as not selected, or is there a hook that updates all the selected values to choiceEntry.value === currentValue when the binding is set to a new value?

hborawski: you are never setting the selected value as not selected in this scenario, it would always use the data model as the source of truth, setting the value of the binding to a new value will rerender and the transform will run again and `selected` will be populated based on the new value in the data model

Issue #163, opened 2023-09-29
Android: Add footer to info reference asset


hborawski:
Depends on #161 

To better model some real world scenarios, the `info` asset should be able to render content below the `actions`.

Render the `footer` for  the [Android reference asset](https://github.com/player-ui/player/blob/main/plugins/reference-assets/android/src/main/java/com/intuit/player/android/reference/assets/info/Info.kt)

sugarmanz:
Closing as was done in #398

PR #196, opened 2023-10-19
add footer to android info asset

mashrurc-alt:
This PR addresses https://github.com/player-ui/player/issues/163


plugins/reference-assets/android/src/main/res/layout/info.xml
```diff
@@ -33,4 +33,15 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@+id/info_primary_info" />
 
+    <LinearLayout
```
brocollie08, line 36: is there a reason for the choice of `LinearLayout` here? It looks like we only handle `footer` as a singular asset

brocollie08:
Could you please add to the tests for this asset

sugarmanz:
Closing, was done in #398

PR #197, opened 2023-10-20
Core Image Reference Asset 

### LINKED ISSUE #186: Core: New Reference Asset - Image
An image asset should be added as a reference asset that will allow users to upload relevant images to complement or enhance text. An image asset should be able to have an optional placeholder tag and an optional caption incase users want to provide text with that image.

## Requirements

- An image asset should have a metaData,  

> - Which should include a ref to the image . 

> - As well as an optional accessibility , which will be a string for accessibility support.

- An image asset should have an optional placeholder text.

> - The placeholder is a string

- An image asset should have an optional caption text, if a caption needs to be displayed along with the image. 

> - This caption should be an asset

 
At its simplest case, this JSON should render an image
```
    {
      "id": "image-1",
      "type": "image",
      "metaData": {
        "ref": "https://player-ui.github.io/latest/logo/logo-dark-large.png"
      }
    }
``` 

## Execution:
Create a new Image asset in the [core reference assets](https://github.com/player-ui/player/tree/main/plugins/reference-assets/core/src/assets) .
Add mocks to the [mocks](https://github.com/player-ui/player/tree/main/plugins/reference-assets/mocks) package to show the optional fields and show all implementations for rendering.

aarianasingh:
Fixes #186 
Created an Image reference asset (core and React).

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`


plugins/reference-assets/core/src/assets/image/transform.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Sets the Image's placeholder and accessibilty
+ */
+export const imageTransform: TransformFunction<ImageAsset, TransformedImage> = (
+  props
+) => {
+  const altText = getImageAlt(props);
+
+  let newImage = {
```
hborawski, line 26: this can be `const` since it does not get mutated

hborawski (changes requested):
Looks like the lint is failing, you should be able to run:
```sh
bazel test //plugins/reference-assets/core:@player-ui/reference-assets-plugin-lint
```
to reproduce

hborawski (approved):

PR #201, opened 2023-10-24
Add user info to assetViewModel (Breaking Change)

nancywu1:
Model can now access the userInfo which can contain information set by the user usually through the plugins (SwiftUIPendingTransactionPlugin, SwiftUIBeaconContextPlugin) which will come later in OSS

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


ios/packages/swiftui/Sources/types/assets/ControlledAsset.swift
```diff
@@ -15,16 +15,21 @@ open class AssetViewModel<T: AssetData>: ObservableObject {
...
     public var bag = Set<AnyCancellable>()
 
     /**
      Constructs an instance of this ViewModel with the given data
      - parameters:
         - data: The data to publish from this ViewModel
+        - userInfo:  Any contextual information set by the user 
      */
-    public required init(_ data: T) {
+    public required init(_ data: T, userInfo: [CodingUserInfoKey: Any]? = nil) {
```
hborawski, line 30: Should this be optional? or default to `[:]`?

ios/packages/swiftui/Sources/types/assets/ControlledAsset.swift
```diff
@@ -75,7 +80,7 @@ open class ControlledAsset<DataType: AssetData, ModelType>: SwiftUIAsset where M
             decoder.logger?.t("Updating model for \(data.id)")
             model.data = data
         } else {
-            self.model = ModelType(data)
+            self.model = ModelType(data, userInfo: decoder.userInfo)
```
hborawski, line 83: Do we need to update the `userInfo` when we get the model from the cache?

ios/packages/swiftui/Sources/types/assets/ControlledAsset.swift
```diff
@@ -15,16 +15,21 @@ open class AssetViewModel<T: AssetData>: ObservableObject {
     /// The decoded data
     @Published public var data: T
 
+    /// Any contextual information set by the user
+    public var userInfo: [CodingUserInfoKey: Any]?
```
hborawski, line 19: If we don't need to update this, we can make this `let` instead of `var`

hborawski (commented):
Need to update [documentation about this as well](https://github.com/player-ui/player/blob/main/docs/site/pages/assets/custom.mdx#L177)

ios/packages/swiftui/Sources/types/assets/ControlledAsset.swift
nancywu1: good call

nancywu1:
not sure why the following tests InfoAssetUITests.testInfoBasic(), InputAssetUITests.testBasicInput(), InputAssetUITests.testInputValidation() keep failing for eyes not finding elements, but passes locally when i run them

hborawski (approved):

PR #206, opened 2023-10-25
Mocks for plugins

mercillo:
### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`


plugins/pubsub/mocks/build.bzl
```diff
@@ -0,0 +1,27 @@
+def _generate_manifest(context):
```
hborawski, line 1: should we move this into [rules_player](https://github.com/player-ui/rules_player) so we dont duplicate this file in multiple packages?

jvm/testutils/deps.bzl
```diff
@@ -11,11 +11,11 @@ main_exports = [
...
     "//plugins/common-types/jvm:common-types",
-    "//plugins/reference-assets/jvm:reference-assets",
+    "//plugins/mocks:jar",
 ]
 
 main_deps = main_exports
 
 main_resources = [
-    "//plugins/reference-assets/mocks",
+    "//plugins/mocks:jar",
```
sugarmanz, line 20: I don't think this is necessary anymore if you're defining the JAR as a dep. It's more for when you have actual files to include within a JAR.

plugins/external-action/mocks/index.ts
```diff
@@ -0,0 +1,7 @@
+export interface Mock {
+  group: string;
+  name: string;
+  path: string;
+}
+
+export type Manifest = Array<Mock>;
```
sugarmanz, line 7: Is this file necessary?

sugarmanz: If so, could we put in a common place to reference from each individual mocks JS package?

plugins/reference-assets/android/build.bzl
```diff
@@ -13,11 +13,13 @@ def kt_asset_test(
...
         test_class = test_class,
         deps = deps + [
+            "//plugins/mocks:jar",
             "//plugins/reference-assets/android/src/androidTest/java/com/intuit/player/android/reference/assets/test",
             "//jvm/j2v8:j2v8-all",
+
         ],
         resources = [
-            "//plugins/reference-assets/mocks",
+            "//plugins/mocks:jar",
```
sugarmanz, line 22: Shouldn't be needed

plugins/reference-assets/android/deps.bzl
```diff
@@ -10,6 +10,7 @@ main_exports = [
 
 main_deps = main_exports + parse_coordinates(maven_main) + [
     "//jvm:kotlin_serialization",
+    "//plugins/mocks:jar",
```
sugarmanz, line 13: The main dependency shouldn't include the mocks

jvm/core/deps.bzl
```diff
@@ -15,5 +15,8 @@ main_deps = main_exports + [
 main_runtime_deps = []
 
 # Test dependencies
-test_deps = []
+test_deps = [
+     "//plugins/mocks:jar",
+     "//plugins/reference-assets/jvm:reference-assets",
```
sugarmanz, line 20: Will these flow automatically from the `testutils` dependency?

mercillo: let me try removing these to see if these the flows get loaded still!

plugins/external-action/mocks/index.ts
mercillo: I'm having trouble accessing this file from other places with a single source.

I tried moving this to the root of plugins/index.ts

and within the BUILD file, I made this a filegroup

```
filegroup(
    name = "mock_types",
    srcs = glob(["*.ts"]),
    visibility = ["//visibility:public"],
)
``` 

within one of the plugins/mocks folder (in this case i used `plugins/pubsub/mocks/BUILD` 
I have a rule that copies this over to the root of the index folder 

```
genrule(
    name = "copy",
    srcs = ["//plugins:mock_types"],
    outs = ["index.ts"],
    cmd = "cp -r $(SRCS) $(OUTS)"
)

javascript_pipeline(
    name = "@player-ui/pubsub-plugin-mocks",
    entry = "index.ts",
    dependencies = [
        "@npm//@player-tools/dsl",
        ":copy"
    ],
    other_srcs = [
       "index.ts",
        ":manifest",
    ] + glob(["**/*.tsx"]),
    out_dir = "",
)
```



with the error being: 

ERROR: /Users/mercillo/Documents/intuit/playerOS/player/plugins/pubsub/mocks/BUILD:23:20: output 'plugins/pubsub/mocks/package.json' was not created
ERROR: /Users/mercillo/Documents/intuit/playerOS/player/plugins/pubsub/mocks/BUILD:23:20: Action plugins/pubsub/mocks/package.json failed: not all outputs were created or valid

would this be the right approach haha

index.bzl
```diff
@@ -175,3 +175,32 @@ def javascript_pipeline(
...
+            ]
+        ]),
+    )
+    return [DefaultInfo(files = depset([manifest], transitive = [mock[DefaultInfo].files for mock in context.attr.mocks]))]
+
+generate_manifest = rule(
+    attrs = {
+        "mocks": attr.label_list(
+            mandatory = True,
+            allow_files = [".json"],
```
sugarmanz, line 202: Should this be enabled or not? I don't think we can use glob anymore, but filegroup would get us what we need. Would removing this line enforce that a filegroup is mandatory, rather than a glob?

sugarmanz (approved):

Issue #169, opened 2023-09-29
Core: Transform reference action asset role


hborawski:
The reference [react action asset]() currently renders a `<` icon when the value of the action indicates that the user would go backwards in the flow. Rather than have assets on each platform implement this check, we can move it into the [transform](https://player-ui.github.io/latest/assets/transforms) and expose a generic property in the [action metadata](https://github.com/player-ui/player/blob/main/plugins/reference-assets/core/src/assets/action/types.ts#L24) to make the coupling between icon and value looser.

- Move logic for determining `isBackAction` into the [action transform](https://github.com/player-ui/player/blob/main/plugins/reference-assets/core/src/assets/action/transform.ts) and populate a `role` property in the `metaData` when applicable.

- Update the [react asset](https://github.com/player-ui/player/blob/main/plugins/reference-assets/react/src/assets/action/Action.tsx#L22) to use `metaData.role` instead of checking the value directly.

abhisheksrivastava177:
Hi @hborawski 
I am interested in working this feature of player. Can you assign me this and it's references ?

manansharma18:
Hey @mercillo , I can have a crack at it. @abhisheksrivastava177 do you have a fix for this issue?

manansharma18:
Working on it now

manansharma18:
Hey @hborawski and @mercillo , a quick question. If we add a ref property in the asset  then going ahead all the FRF can specify that if they need an icon or not. For the already implemented back button, we need to make sure they keep showing the icon. Is that what the transform would be used for?  Is it to make sure the changes are backward compatible with older back buttons.

mercillo:
I think previously, we're only checking to see if the value contains `Prev` to see if its a back button. 
but if the value contains Prev we can add the metadata in the transform

metaData: { "role": "back}

hborawski:
Yes this is to consolidate existing logic, not to add arbitrary icons to the `action` asset

manansharma18:
https://github.com/player-ui/player/pull/211

PR #211, opened 2023-10-29
feat/decouple back button

manansharma18:
Screenshot 2023-11-05 at 11 58 40 AM

https://github.com/player-ui/player/issues/169
### Change Type (required)
Adds a role property inside the metaData on the action asset. This property is used to decouple the back Icon from the back button.

- [X] `patch`
- [ ] `minor`
- [ ] `major`

### DIFF
--- a/plugins/reference-assets/core/src/assets/action/types.ts
+++ b/plugins/reference-assets/core/src/assets/action/types.ts
@@ -29,7 +29,7 @@ export interface ActionAsset<AnyTextAsset extends Asset = Asset>
     skipValidation?: boolean;
 
     /** boolean value to decide for the left anchor sign */
-    role?: boolean;
+    role?: string;
   };
 }
 
--- a/plugins/reference-assets/core/src/assets/action/types.ts
+++ b/plugins/reference-assets/core/src/assets/action/types.ts
@@ -27,6 +27,9 @@ export interface ActionAsset<AnyTextAsset extends Asset = Asset>
 
     /** Force transition to the next view without checking for validation */
     skipValidation?: boolean;
+
+    /** boolean value to decide for the left anchor sign */
+    role?: boolean;
   };
 }
 
--- a/plugins/reference-assets/core/src/assets/action/types.ts
+++ b/plugins/reference-assets/core/src/assets/action/types.ts
@@ -28,7 +28,7 @@ export interface ActionAsset<AnyTextAsset extends Asset = Asset>
     /** Force transition to the next view without checking for validation */
     skipValidation?: boolean;
 
-    /** boolean value to decide for the left anchor sign */
+    /** string value to decide for the left anchor sign */
     role?: string;
   };
 }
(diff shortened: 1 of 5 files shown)


plugins/reference-assets/core/src/assets/action/types.ts
```diff
@@ -27,6 +27,9 @@ export interface ActionAsset<AnyTextAsset extends Asset = Asset>
 
     /** Force transition to the next view without checking for validation */
     skipValidation?: boolean;
+
+    /** boolean value to decide for the left anchor sign */
+    role?: boolean;
```
hborawski, line 32: I think we would want this to be `role?: string;` so we can potentially support other roles in the future, but the transform can populate it with a `back` value when appropriate

plugins/reference-assets/mocks/action/action-navigation.json
```diff
@@ -26,6 +26,23 @@
...
+            "value": "Prev",
+            "label": {
+              "asset": {
+                "id": "action-prev-id-without-icon",
+                "type": "text",
+                "value": "Go Back Without Icon"
+              }
+            },
+              "metaData": {
+                "role": false
```
hborawski, line 42: Is this to test that the transform will set it to `true`? 

If this is set explicitly, would we even want to change it?

hborawski (changes requested):

plugins/reference-assets/core/src/assets/action/types.ts
manansharma18: Hey @hborawski , so if I make the role as string then In the transform, I need to check if it is a backButton and then apply the role as 'back'(maybe). In the action component, then check for the role value. Is this understanding correct?

plugins/reference-assets/mocks/action/action-navigation.json
manansharma18: So, currently the user can specify the role property as true/false to show/hide the back icon. This part of story book shows a back button with a back icon

plugins/reference-assets/core/src/assets/action/types.ts
hborawski: yes this is correct, as `"role": true` does not really have any semantic intent, `"role": "back"` is more clear as to what is happening,

plugins/reference-assets/mocks/action/action-navigation.json
hborawski: if this is `false`, why would this show the icon?

manansharma18: Hey, this frf is to show, the useage of `'role':'back'`. The one above this shows how to get the button without the back icon.

plugins/reference-assets/core/src/assets/action/types.ts
```diff
@@ -27,6 +27,9 @@ export interface ActionAsset<AnyTextAsset extends Asset = Asset>
 
     /** Force transition to the next view without checking for validation */
     skipValidation?: boolean;
+
+    /** boolean value to decide for the left anchor sign */
```
hborawski, line 31: this docstring should be updated since it is a string now

hborawski (approved):

mercillo (approved):

PR #212, opened 2023-10-31
link mock flows from plugins to ios mockflows

nancywu1:
Branched off of Marky's https://github.com/player-ui/player/pull/206

Looks for mocks under each plugin and pulls the JSON files if they exist and append them onto the existing mocks flows in iOS 

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


nancywu1:
replace by https://github.com/player-ui/player/pull/213 because it doesn't need to be branched off of marky's PRs

PR #213, opened 2023-11-01
link shared plugin mocks for iOS

nancywu1:
Looks for mocks under each plugin and pulls the JSON files if they exist and append them onto the existing mocks flows in iOS

Will merge this before Marky's https://github.com/player-ui/player/pull/206

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


nancywu1:
going to link mocks from .json files

PR #215, opened 2023-11-02
177/ add plugin examples and managed player to demo app

nancywu1:
* add examples for pubsub plugin, beacons plugin, external action and pending transaction plugin
* add new tab for managed player flows
* add UI tests

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


ios/packages/test-utils/Sources/ui-test/AssetAndPluginCollection.swift
```diff
@@ -0,0 +1,274 @@
...
+                    Text(section.title)
+                }
+            }
+        }
+        .accessibility(identifier: "AssetAndPluginCollection")
+        .navigationBarTitle(Text("Flows"))
+    }
+
+    var playerListSection: some View {
+        List {
```
hborawski, line 89: These utilities are published for consuming users to use, this would mean that all users using the utility will now have a section for these hardcoded managedplayer flows if they were to use this.

This will need to be refactored so that we can show those flows in our demo app, using the same tools we publish for consuming users. But we do not want to include our plugin examples in other users demo/example apps.

In my mind, we would not have changed `AssetCollectionView`, we would provide an additional utility to setup more specific flows, and have it use `AssetCollectionView` to display the asset flows as it does today.

hborawski (changes requested):

ios/packages/test-utils/Sources/ui-test/MockFlowManager.swift
```diff
@@ -0,0 +1,39 @@
+//
+//  MockFlowManager.swift
+//  PlayerUI
+//
+//  Created by Zhao Xia Wu on 2023-10-26.
+//
+
+import Foundation
+
+public class MockFlowManager: FlowManager {
```
hborawski, line 10: This seems very similar to [ConstantFlowManager](https://github.com/player-ui/player/blob/main/ios/packages/swiftui/Sources/ManagedPlayer/ConstantFlowManager.swift) can they be converged?

ios/packages/test-utils/Sources/ui-test/SegmentControlView.swift
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    var playerListSection: some View {
+        List {
+            // Plugin flows
+            ForEach(pluginSections, id: \.title) { section in
+                Section {
+                    ForEach(section.flows, id: \.name) { flow in
+                        NavigationLink(flow.name) {
+                            PlayerLoader(flow: flow.flow)
```
hborawski, line 112: I think we can probably just use `AssetFlowView` here instead of introducing a new `PlayerLoader`, as `PlayerLoader` doesn't use the same plugins that are passed into this `SegmentControlView`

ios/packages/test-utils/Sources/ui-test/SegmentControlView.swift
```diff
@@ -0,0 +1,231 @@
...
+                        }
+                        .accessibility(identifier: "\(section.title) \(flow.name)")
+                    }
+                } header: {
+                    Text(section.title)
+                }
+            }
+
+            // Managed Player flows
+            Section {
```
hborawski, line 123: Since this is tightly coupled to  ManagedPlayer example flows, we should probably move this file to the demo app, so as not to publish it for users that consume the UI test utilities

ios/packages/test-utils/Sources/ui-test/String + Extensions.swift
```diff
@@ -0,0 +1,156 @@
...
+//  String + Extensions.swift
+//  PlayerUI-UI-XCUITests
+//
+//  Created by Zhao Xia Wu on 2023-11-01.
+//
+
+import Foundation
+
+// mocks for managed player flows
+extension String {
```
hborawski, line 11: Also move to the demo app

ios/packages/test-utils/Sources/ui-test/FlowManagerView.swift
```diff
@@ -0,0 +1,62 @@
...
+//  Created by Zhao Xia Wu on 2023-11-01.
+//
+
+import SwiftUI
+
+/**
+ SwiftUI View to wrap the `ManagedPlayer` and handle the result
+ for use in UI testing
+ */
+public struct FlowManagerView: View {
```
hborawski, line 14: This should probably go in the demo app as well, since its tightly coupled to `ReferenceAssetsPlugin`

ios/packages/test-utils/Sources/ui-test/AssetUITestCase.swift
```diff
@@ -51,7 +51,16 @@ open class AssetUITestCase: XCTestCase {
...
-        app.otherElements.buttons[mockName].firstMatch.tap()
+
+        let button = app.otherElements.buttons[mockName].firstMatch
+        app.scrollToElement(element: button)
+
+        waitFor(button)
+        button.tap()
+    }
+
+    open func switchToPlayerTab() {
```
hborawski, line 62: this looks unused

ios/packages/test-utils/Sources/ui-test/AssetUITestCase.swift
```diff
@@ -51,7 +51,16 @@ open class AssetUITestCase: XCTestCase {
      */
     open func openFlow(_ mockName: String) {
         navigateToAssetCollection()
-        app.otherElements.buttons[mockName].firstMatch.tap()
+
+        let button = app.otherElements.buttons[mockName].firstMatch
+        app.scrollToElement(element: button)
+
+        waitFor(button)
+        button.tap()
```
hborawski, line 59: this should be moved into our actual UItests by overriding `navigateToAssetCollection`, since the `SegmentControlView` should be in the demo app

hborawski (changes requested):

String + Extensions.swift
```diff
@@ -0,0 +1,156 @@
+//
```
hborawski, line 1: this looks like a duplicate file, and is at the repo root, should probably be removed

ios/packages/test-utils/Sources/ui-test/MockFlowManager.swift
hborawski, line 10: Came back in the rebase?

ios/packages/test-utils/Sources/ui-test/AssetUITestCase.swift
```diff
@@ -50,7 +50,7 @@ open class AssetUITestCase: XCTestCase {
         - mockName: The name of the mock to open
      */
     open func openFlow(_ mockName: String) {
-        navigateToAssetCollection()
+
```
hborawski, line 53: probably shouldnt be removed, we can just not use `super` in the override location

hborawski (changes requested):

hborawski (approved):

PR #237, opened 2023-11-15
adding swiftuipendingtransactionplugin

nancywu1:
ported over plugin from internal
added guide to docs
will make example for plugin in separate PR

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


docs/site/pages/plugins/swiftui-pending-transaction.mdx
```diff
@@ -0,0 +1,128 @@
...
+        set { self[TransactionContextAlternativeKey.self] = newValue }
+    }
+}
+```
+
+**Step 3:**
+
+To start registering transactions with the assets, access the transactionContext through the default environment variable above or the new one you created
+
+Register the transaction with the "input" phase (the view update) when a user starts editing the input,  (code snippet taken from [SwiftUIInputAsset](https://github.intuit.com/player/ios/blob/master/iOSPlayer/src/packages/reference-assets/src/SwiftUI/SwiftUIInputAsset.swift)
```
hborawski, line 80: this link needs to be updated to point to the open source reference asset

docs/site/pages/plugins/swiftui-pending-transaction.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+```swift
+onEditingChanged: {editing in
+    guard !editing else {
+        // register the transaction once editing begins
+        transactionContext?.register(.input) {
+            self.model.set()
+        }
+        return** context.editBegin()
```
hborawski, line 89:
```suggestion
        return context.editBegin()
```

docs/site/pages/plugins/swiftui-pending-transaction.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+```
+Note: we can register multiple transactions under a single phase and we commit based on phase, when commiting a phase if multiple transactions exist they commit in no particular order
+
+
+**Step 4:**
+
+Commit the transaction (view update) before the action to navigate takes place. The WrappedFunction provides access to the userInfo where the TransactionContext is stored and users can use this how they see fit.
+
+Now in the [SwiftUIActionAsset](https://github.intuit.com/player/ios/blob/master/iOSPlayer/src/packages/reference-assets/src/SwiftUI/SwiftUIActionAsset.swift) instead of calling WrappedFunction `run` directly, we can extend the WrappedFunction and create another function called commitCallbacksThenCall which does the same thing as the normal call function except it will check for any input callbacks and commit then if they exist.
```
hborawski, line 106: link to open source reference asset

hborawski (changes requested):

hborawski (approved):

PR #256, opened 2023-12-07
Fix `com.intuit.player:j2v8` transitive deps

sugarmanz:
This was initially supposed to be a quick PR to address an integration issue. Unfortunately, the integration issue was not a quick fix. We hit roadblocks on AAR publishing, viewbinding support, GPG signing, JDK versions, etc. During the journey, we:

- Modified J2V8 transitive dependency
  - Support for `android-debug` artifact to automatically pull in correct dependencies
- Swapped from [vaticle/bazel-distribution](https://github.com/vaticle/bazel-distribution) to [bazelbuild/rules_jvm_external](https://github.com/sugarmanz/rules_jvm_external/tree/maven-export-aar) for JVM publishing
  - [Enhanced for AAR support](https://github.com/bazelbuild/rules_jvm_external/compare/master...sugarmanz:rules_jvm_external:maven-export-aar)
  - Required JDK 11 upgrade (using embedded JDK)
  - Fix for uploading to Artifactory ([relevant discussion](https://github.com/vaticle/bazel-distribution/pull/318))
  - Update [rules_player#39](https://github.com/player-ui/rules_player/pull/39)
- Enhanced [grab_bazel_common](https://github.com/sugarmanz/grab-bazel-common/tree/pre-1.8) databinding_aar to support publishing through `rules_jvm_external`

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

<details>
  <summary>📦 Published PR as canary version: <code>0.5.2--canary.256.8714</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.5.2--canary.256.8714
</details>


KetanReddy (approved):

jvm/j2v8/BUILD
```diff
@@ -17,6 +17,9 @@ j2v8_platform("macos")
 
 j2v8_platform("linux")
 
+# TODO: These should probably be AARs?
```
brocollie08, line 20: there's no android specific resource here, right? I think i stripped all that out in order to fit the debugger in the jvm repo?

patches/rules_jvm_external.default_public_visibility.patch
```diff
@@ -0,0 +1,13 @@
+diff --git coursier.bzl coursier.bzl
```
brocollie08, line 1: this looks like a git diff, do we need this? I see it used[ here ](https://github.com/player-ui/player/pull/256/files#diff-5493ff8e9397811510e780de47c57abb70137f1afe85d1519130dc3679d60ce5R20) but why

build_constants.bzl
```diff
@@ -0,0 +1,36 @@
...
+    name = 'files',
+    srcs = glob(['**']),
+    visibility = ['//visibility:public']
+)
+"""
+    repository_ctx.file("BUILD", _BUILD_FILE, False)
+
+    version = repository_ctx.read(repository_ctx.attr.version_file)
+
+    _CONSTANTS_FILE = """
```
brocollie08, line 14: is this building version constants for the whole repo? i'm curious to see what this looks like

WORKSPACE
```diff
@@ -137,12 +147,6 @@ overridden_targets = {
 
 load("@bazel_tools//tools/build_defs/repo:maven_rules.bzl", "maven_aar")
 
-# Because J2V8 is published as type `aar.asc`
-maven_aar(
```
brocollie08, line 141: how does this part work now? i noticed you removed the dep for `android_binary`, I'm not following how _this_ source is fixed

jvm/j2v8/BUILD
sugarmanz: The main problem here is they rely on transitive AARs, so if you try to pull the JAR into a Gradle project, it'll technically work, but will blow up when it tries to process the transitive deps. So, the idea is to make this an Android only artifact up front to make it clearer where this is meant to be used.

That being said, I'm not sure that's a great way to do it. From looking at a lot of the Android libs, a lot of them are JARs if they don't have any bundled Android resources.

patches/rules_jvm_external.default_public_visibility.patch
sugarmanz: `rules_jvm_external` made the change to not make the downloaded artifacts public:
https://github.com/bazelbuild/rules_jvm_external/pull/649

We only rely on the artifacts directly for `kotlinx_coroutines_core_fixed`, which I hope will go away at some point. So, rather than push a commit to change the behavior in `rules_jvm_external`, we apply a patch on demand to fix the issue in our workspace.

build_constants.bzl
sugarmanz: Yep! This is what usage looks like:
https://github.com/player-ui/player/blob/78345ac70c5bb34fcd58147ab49e38f22d3524fd/android/player/BUILD#L2

This is required because the new `maven_publish` rule requires the version to be defined within the `maven_coordinates` tag.

sugarmanz: Inspiration in thread:
https://github.com/bazelbuild/rules_jvm_external/issues/618

WORKSPACE
sugarmanz: Somewhat hilariously, I think this is because it's a transitive dependency of `com.github.AlexTrotsenko:j2v8-debugger`, so it gets pulled into the `@maven` workspace.

I could be wrong, and maybe it just works now, but that's the main thing that changed.

build_constants.bzl
brocollie08: now that i'm reading this again, it's not dynamic
does this have the potential to be dynamic for constants that are shared across the platforms or will this always be hard coded

or I guess i'm asking what's the diff between this and creating that output ourselves

WORKSPACE
brocollie08: 😅

patches/rules_jvm_external.default_public_visibility.patch
brocollie08: oh wow so these changes are actually taking effect when you use the repo at specified branch? That's pretty cool but at the same time it's kinda nuts this capability exists

sugarmanz: Yeah, this patch would be applied to the repo after it's cloned. Definitely a pretty "dangerous" feature. It exists for other types of repo declarations as well, like `http_archive`:
https://bazel.build/rules/lib/repo/http#http_archive-patches

build_constants.bzl
sugarmanz: Hmm, I'm not sure I understand what you mean. It's dynamic in the sense that it makes the version defined in our `VERSION` file accessible at build configuration time. We don't have to change this script every time the version changes. Typically, you're not able to read files during build configuration, only during execution.

The only thing that's hardcoded is the `build_constants` repository rule itself. If we want to add other things, like the `APPLITOOLS_API_KEY`, we could by reading the environment and stuffing into this file. Then you could have access to the key at build configuration time, and maybe even not generate certain test targets if that key doesn't exist.

brocollie08 (approved):

PR #260, opened 2023-12-08
[BAZEL-6] Add coverage support for JS files

adierkens:
## Release Notes

- Adds support for coverage info to vitest tests
- Updates to build with node@18.x


KetanReddy:
The snapshot updates make sense but any idea why the binding test returns a different value?

adierkens:
I'm guessing it has to do with the update to node@18 -- changes to how the Error.message is serialized

PR #263, opened 2024-01-03
Latest sync including AsyncNodePlugin

brocollie08:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


core/player/src/view/parser/index.ts
```diff
@@ -117,13 +140,15 @@ export class Parser {
       }
     }
 
+    /**
+     *
+     */
```
KetanReddy, line 145: We should probably have a comment here or remove the comment strings.

hborawski (approved):

PR #270, opened 2024-01-18
Playa 8756 - iOS add callTryCatchWrapper function on JSValue

nancywu1:
for functions not called as a part of a player process like "transition", errors are ignored

`tryCatch` calls the function inside a try catch to catch the error and return it to allow the callee to determine what they want to do with the error 

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

  # Release Notes
  iOSPlayer to PlayerUI Migration Guide

This guide contains the breaking changes between iOSPlayer and PlayerUI

### FlowController Transition

FlowControllers `transition` function can throw errors in Core Player, however it's not called as a part of a player process so these errors were ignored on iOS. 
The change makes the `transition` function a throwing function. Any existing calls to this now require the `try` keyword

#### Before

```swift
public func transition(with action: String) {
         value.invokeMethod("transition", withArguments: [action])
```

#### After
```swift
public func transition(with action: String) throws {
         try self.value.objectForKeyedSubscript("transition").tryCatch(args: [action])
```

Error is returned from the `tryCatch` function if it exists

#### Usage After

```swift
(player.state as? InProgressState)?.controllers?.flow.transition(with: "NEXT")
```

#### Usage Before

```swift
do {
  try (player.state as? InProgressState)?.controllers?.flow.transition(with: "NEXT")
} catch let error {
   // error handling
   // For example: call fail on InProgressState to get ErrorState for flow to end
}
```


ios/packages/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -0,0 +1,78 @@
...
+     - parameters:
+        - args: List of arguments taken by the function
+     */
+    public func callTryCatchWrapper(args: Any...) throws {
+        let result = self.tryCatchWrapper?.call(withArguments: [self, args])
+
+        let isError = self.errorCheckWrapper?.call(withArguments: [result as Any])
+
+        if isError?.toBool() == true {
+            throw JSValueError.thrownFromJS
```
hborawski, line 50: Should we include error information about the underlying JS Error?

ios/packages/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -0,0 +1,78 @@
...
+import Foundation
+import JavaScriptCore
+
+extension JSValue {
+    var tryCatchWrapper: JSValue? {
+        self.context.evaluateScript(
+        """
+           (fn, args) => {
+               try {
+                   console.log(args)
```
hborawski, line 17:
```suggestion
```

ios/packages/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -0,0 +1,78 @@
...
+                   return fn(...args)
+               } catch(e) {
+                   console.log(e)
```
hborawski, line 20:
```suggestion
```

ios/packages/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -0,0 +1,78 @@
...
+                   if (e instanceof Error) {
```
hborawski, line 21: this should always be true, so its probably not needed

ios/packages/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+
+    /**
+     A way to catch errors for void functions not called inside a player process. Called on void functions.
+     - parameters:
+        - args: List of arguments taken by the function
+     */
+    public func callTryCatchWrapper(args: Any...) throws {
+        let result = self.tryCatchWrapper?.call(withArguments: [self, args])
```
hborawski, line 45: Since `tryCatchWrapper` is a computed property, that means it evals that snippet every time we call this

does JavascriptCore clean it up automatically correctly? or should will we need to handle something?

ios/packages/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -0,0 +1,78 @@
...
+        """)
+    }
+
+
+    /**
+     A way to catch errors for void functions not called inside a player process. Called on void functions.
+     - parameters:
+        - args: List of arguments taken by the function
+     */
+    public func callTryCatchWrapper(args: Any...) throws {
```
hborawski, line 44: i think this name is a bit long, can we just call it `tryCatch` maybe ?

ios/packages/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+
+    /**
+     A way to catch errors for functions not called inside a player process. Called on functions with a return value.
+     - parameters:
+        - args: List of arguments taken by the function
+     */
+    public func callTryCatchWrapperWithReturnValue(args: Any...) throws -> JSValue? {
```
hborawski, line 60: you could potentially just use this one and mark it `@discardableResult` since a `Void` function in JS would still return `undefined` technically

hborawski (changes requested):

hborawski (approved):
🚀

hborawski (approved):

PR #281, opened 2024-01-29
plugins core and react doc addition

lexfm:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


docs/site/pages/writing-plugins.mdx
```diff
@@ -13,6 +13,22 @@ platform: core,react
 
 Core plugins are the easiest way to extend Player functionality regardless of what platform you are using Player on. To make writing core plugins easy `@player-ui/player` exposes an interface `PlayerPlugin` that denotes everything needed. The two mandatory features are a `name` property which is lets Player know how to refer to the plugin and an implemented `apply` function that takes a `player` object. Optionally a `symbol` property can be used to provide a unique identifier that can be used to retrieve the plugin from Player. 
 
+The first step for creating a plugin is making our plugin class, making sure it implements the `PlayerPlugin` interface from `@player/core`. By convention, a name attribute with the dash-cased name of your plugin should be defined.
```
KetanReddy, line 16: This doesn't look like the right package

docs/site/pages/writing-plugins.mdx
```diff
@@ -13,6 +13,22 @@ platform: core,react
...
+
+```javascript
```
KetanReddy, line 18: Should use `typescript` here instead of `javascript`

docs/site/pages/writing-plugins.mdx
```diff
@@ -41,7 +57,7 @@ _Note: For the React Player you can import and load the plugin the same way you
   </core>
   <react>
 
-React Player Plugins are very similar to core plugins in both their composition and use. The `@player-ui/react` package exposes an interface `ReactPlayerPlugin` that, much like the `PlayerPlugin` interface provides the necessary attributes that are required for a React Player plugin. Again a `name` attribute is required and a function `applyReact` is required that takes a `ReactPlayer` instance. Similarly to core plugins in the `applyReact` function you have access to the React Player object and access to the three exposed hooks:
+React Player Plugins are very similar to core plugins in both their composition and use. The `@player-ui/react` package exposes an interface `ReactPlayerPlugin` that, much like the `PlayerPlugin` interface provides the necessary attributes that are required for a React Player plugin. Again a `name` attribute in dash-case is required and a function `applyReact` is required that takes a `ReactPlayer` instance. Similarly to core plugins in the `applyReact` function you have access to the React Player object and access to the three exposed hooks:
```
KetanReddy, line 60: Previously we say that the dash case naming is "by convention" and here we are saying its required. We should keep that consistent

KetanReddy:
I think one thing that is missing is that "next step" of documentatio. Demonstrating how to create a plugin is good but covering what you can actually do would be better. Doing it inline might get a bit long so maybe that becomes a different section? One thing we do in some other parts of the docs is take something already in the codebase and explain how/why its built the way it is.

lexfm:
Closing due to forked branch not completing checks, moved changes to [this PR](https://github.com/player-ui/player/pull/294) instead.

PR #284, opened 2024-01-30
iOS: Update plugin documentation

hborawski:
- Adds Cocoapods instructions for each plugin
- adds a usage example for each plugin

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

### DIFF
--- a/docs/site/pages/plugins/beacon.mdx
+++ b/docs/site/pages/plugins/beacon.mdx
@@ -32,6 +32,8 @@ interface DefaultBeacon {
 }
 ```
 
+## Usage
+
 <PlatformTabs>
   <core>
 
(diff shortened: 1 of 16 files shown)


docs/site/pages/plugins/beacon.mdx
```diff
@@ -32,6 +32,8 @@ interface DefaultBeacon {
 }
 ```
 
+## Usage
```
brocollie08, line 35: 👍 this should be the standard nomenclature for all plugin docs, I think Expression is missing this

hborawski: https://github.com/player-ui/player/blob/0ad9807460e6cc5dd533264ef40e4d09cc959ed9/docs/site/pages/plugins/expression.mdx#L10

should be there for expression

PR #288, opened 2024-01-31
Refactor existing DSL docs.

KetanReddy:
Expand existing DSL docs

### Change Type

- [x] `patch`
- [ ] `minor`
- [ ] `major`

# Release Notes
Update DSL docs


docs/site/pages/dsl/index.mdx
```diff
@@ -0,0 +1,152 @@
+---
+title: DSL Content Overview
+---
+
+# TSX/JSX Content Authoring (Player DSL)
+
+While Player content _can_ be written directly in JSON, it's definitely not the preferable authoring format. To take advantage of existing developer tool-chains, Player provides a mechanism for authoring content in (J/T)SX as React components and simple TypeScript objects. The Player CLI can then be used to transpile the React tree into a JSON content.
```
spentacular, line 7: I think the intro sentence could go into selling the DSL a little more, such as linting, validation, brevity, etc. Maybe even showing a basic comparison of JSON versus DSL, showing the auto id, auto collection wrapping, auto text wrapping?

docs/site/pages/dsl/index.mdx
```diff
@@ -0,0 +1,152 @@
...
+"player": {
+    "dsl": {
+      "src": "./src/main/tsx",
+      "outDir": "./out"
+    },
+    "json": {
+      "src": "./out/*.json"
+    },
+  }
+```
```
spentacular, line 41: Does this block blow up in the preview since it's not valid JSON due to the missing surrounding `{}`?

docs/site/pages/dsl/index.mdx
```diff
@@ -0,0 +1,152 @@
...
+    },
+    END_Done: {
+      state_type: 'END',
+      outcome: 'done',
+    },
+  },
+};
+```
+
+One convenience feature is the auto injection of the the `ref` proeprty for a `VIEW` type state if the corresponding view is a React tree.
```
spentacular, line 90:
```suggestion
One convenience feature is the auto injection of the the `ref` property for a `VIEW` type state if the corresponding view is a React tree. 
```

docs/site/pages/dsl/index.mdx
```diff
@@ -0,0 +1,152 @@
...
+      outcome: 'done',
+    },
+  },
+};
+```
+### Bindings and Expressions
+
+Both `binding` and `expression` in the JSX authoring leverages a tagged template, typically abbreviated as `b` and `e` respectively. In a similar fashion to using `css` or `graphql` in a JS file, this enables syntax-highlighting and validation of bindings and expressions within a JS file.
+
+To create a binding, or expression:
```
spentacular, line 126: I think this line can go and just leave the example.

docs/site/pages/dsl/schema.mdx
```diff
@@ -0,0 +1,226 @@
...
+const mycustombooleantype = {
+  type: "BooleanType",
+  validation: [
+    {
+      type: 'oneOf',
+      message: 'Value has to be true or false',
+      options: [true, false],
+    },
+  ],
+} satisfies Schema.DataType
```
spentacular, line 151: OOOO nice `satisfies`

docs/site/pages/dsl/views.mdx
```diff
@@ -0,0 +1,128 @@
...
+            "value": "Some label"
+          }
+        }
+      }
+    }
+  ]
+}
+```
+
+Not only is the source DSL content a fraction of the output object's size (making it easier to read and maintain) as the base components use the same TypeScript types as the assets themselves, you will recieve in editor suggestions and type checks as you author your content.
```
spentacular, line 52:
```suggestion
Not only is the source DSL content a fraction of the output object's size (making it easier to read and maintain) as the base components use the same TypeScript types as the assets themselves, you will receive in editor suggestions and type checks as you author your content. 
```

docs/site/pages/dsl/views.mdx
```diff
@@ -0,0 +1,128 @@
...
+      </Switch.Case>
+      <Switch.Case>
+        <Text>Text 1</Text>
+      </Switch.Case>
+    </Switch>
+  </Collection.Label>
+</Collection>
+```
+
+# DSL Benefits in Views
```
spentacular, line 106: Is it worth doing a jump link from the DSL intro to this section? Conceptually I feel like these are both benefits of the DSL, but also the mindset to approach writing your own components.

docs/site/pages/assets/dsl.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+```tsx
+import { Asset, AssetPropsWithChildren } from '@player-tools/dsl';
+
+export const Action = (props: AssetPropsWithChildren<ActionAsset>) => {
+  return <Asset type="action" {...props} />;
+}
+```
+
+This would allow users to import the `Action` component, and _render_ it to JSON:
```
nancywu1, line 44: should we show the DSL and JSON content side by side somewhere, maybe for the Action example below with Label?

docs/site/pages/assets/dsl.mdx
```diff
@@ -0,0 +1,126 @@
...
+const myView = (
+  <Action value="next">
+    <Action.Label>
+      <Text value="Continue" />
+    </Action.Label>
+  </Action>
+);
+```
+
+It also accepts components for automatically creating `text` and `collection` assets, allowing users to shortcut the `Text` component:
```
nancywu1, line 80: should we add example for automatically creating the `collection` asset as well?

docs/site/pages/assets/dsl.mdx
nancywu1: oh i didnt read the Overview section which has the JSON, so maybe can just link to that section?

docs/site/pages/assets/dsl.mdx
```diff
@@ -0,0 +1,126 @@
...
+```
+
+The `AssetPropsWithChildren` type is a utility type to help convert the `Asset` type (which has a required `id` and `type` properties) to a type more suited for components. It changes the `id` to be optional, and adds a `applicability` property automatically.
+
+## Slots
+
+Continuing the example fo the `ActionAsset`, we need a way for users to users to specify the nested `label` property, which itself is another asset. This can be accomplished using the `createSlot` utility function.
+
+```tsx
+import { Asset, AssetPropsWithChildren } from '@player-tools/dsl';
```
adierkens, line 57:
```suggestion
import { Asset, AssetPropsWithChildren, createSlot } from '@player-tools/dsl';
```

docs/site/pages/content/assets-views.mdx
```diff
@@ -69,3 +65,284 @@ Example:
...
+    {
+      "case": "{{name.first}} == 'adam'",
+      "asset": {
+        "id": "name",
+        "type": "text",
+        "value": "Yay"
+      }
+    },
+    {
+      "case": "{{name.first}} == 'margie'",
```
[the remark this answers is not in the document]
adierkens: Names of pets is also a good one to use that's a bit more generic

docs/site/pages/assets/dsl.mdx
```diff
@@ -0,0 +1,126 @@
...
+---
+
+# Creating TSX Components
+
+In order to take advantage of the auto-completion and validation of TypeScript types, asset libraries can export a component library for content authoring. Creating components isn't much different than writing a React component for the web. The primative elements uses the [react-json-reconciler](https://github.com/intuit/react-json-reconciler) to create the JSON content tree, with utilities to make it quick and painless to create new asset-components.
+
+
+## Creating a Basic Component
+
+The `Asset` component from the `@player-tools/dsl` package is the quickest way to create a new component. The `Asset` component will take all the Asset's properties and convert them to their equivalent JSON representation when serialized.
```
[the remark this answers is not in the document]
KetanReddy: Good call. It is a bit confusing, almost intentionally, since we want `Asset` to always mean the base that we provide. In the example below we pull in `Asset` from player types for the base TS definition of the Asset's type *and* we pull in `Asset` from the DSL package for the base of the Asset's DSL component. I can add a note explaining that.

docs/site/pages/assets/dsl.mdx
KetanReddy: I think we should definitely inline it. I can add that.

docs/site/pages/assets/dsl.mdx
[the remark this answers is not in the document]
KetanReddy: Meaning if there was no shortcut for the Text/Collection creation? It would compile the string as the value of the Label property which would cause the content to fail at validation.

KetanReddy: Good call! I can add that

docs/site/pages/dsl/index.mdx
KetanReddy: Yeah I can write some copy for that

docs/site/pages/dsl/index.mdx
KetanReddy: Oh maybe, let me fix that

docs/site/pages/content/assets-views.mdx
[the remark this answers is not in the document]
KetanReddy: I was going to go with "Not Adam"

docs/site/pages/dsl/views.mdx
KetanReddy: Actually I think you may be right. I might write a higher level benefits section in the DSL info page but link out to specific sections in the views/schema pages for more detailed explanations of features?

PR #290, opened 2024-01-31
Plugins android clean up

brocollie08:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


docs/site/pages/plugins/async-node.mdx
```diff
@@ -0,0 +1,122 @@
...
+const player = new Player({
+  plugins: [
+    asyncNodePlugin
+  ]
+})
+```
+  </core>
+  <react>
+
+The `react` version of the AsyncNodePlugin is identical to using the core plugin. Refer to core usage for handler configuration:
```
sugarmanz, line 71: If both `core` and `react` are the same, can we consolidate the tabs? Like just have `react` or just have `core`? This kinda goes back to the difference between headless and UI/platform plugins. Like the `AsyncNodePlugin` doesn't have a `react` specific version in source either.

docs/site/pages/plugins/async-node.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+  </react>
+  <ios>
+
+  </ios>
+  <android>
+
+In build.gradle
+```kotlin
+implementation "com.intuit.player.plugins:async-node:$JVM_PLAYER_VERSION"
```
sugarmanz, line 94: just `PLAYER_VERSION` since divergent JVM player version was an internal concept

docs/site/pages/plugins/beacon.mdx
```diff
@@ -186,24 +186,37 @@ For convenience, there are several extension methods for utilizing a pre-install
...
+  data: Any? = null
+): Unit
+
+// extension on RenderableAsset
+fun RenderableAsset.beacon(
+    action: BeaconAction,
+    element: BeaconElement,
+    asset: Asset = this.asset,
+    data: Any? = null
+): Unit
```
sugarmanz, line 219: Should we just link to the source here to ensure it stays up-to-date?

docs/site/pages/plugins/coroutines.mdx
```diff
@@ -0,0 +1,112 @@
...
+Kotlin coroutines is a popular Kotlin library to handle asynchronous programming and concurrency in the JVM. More information on Coroutines can be found [here](https://kotlinlang.org/docs/coroutines-overview.html)
+This module contains several JVM exclusive plugins to aid various tooling needs.
+
+## UpdatesPlugin
+
+Plugin using a `ReceiveChannel` to provide an updated asset whenever the Player View is updated.
+
+### Usage
+
+<PlatformTabs>
```
sugarmanz, line 17: I'm wondering if `PlatformTabs` is necessary since it's the only platform -- but it does offer a sense of consistency.

docs/site/pages/plugins/coroutines.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+// Experimental
+
+/** This cleans out the [updates] channel to prepare it to receive new updates */
+public fun flush(): Unit
+
+// Player extensions
+public val Player.updatesPlugin: UpdatesPlugin?
+public fun Player.waitForUpdates(timeout: Long = 500): Asset?
+```
```
sugarmanz, line 59: Same comment around just linking to source - if we want separate API docs, we should just generate them

plugins/beacon/jvm/README.md
```diff
@@ -2,4 +2,4 @@
 
 ## Gradle Dependency
 
-`implementation("com.intuit.player.jvm.plugins:beacon:$VERSION")`
+`implementation("com.intuit.player.plugins:beacon:$VERSION")`
```
sugarmanz, line 5: We should either link to the docs for this plugin, or maybe even figure out how we could put the info in this file and load it in the doc site one?

docs/site/pages/plugins/async-node.mdx
```diff
@@ -0,0 +1,122 @@
...
+    // e.g.
+    // listOf(
+    //  mapOf(
+    //    "asset" to mapOf(
+    //      "id" to "asset-1",
+    //      "type" to "text",
+    //      "value" to "new asset!"
+    //    )
+    //  )
+    // )
```
sugarmanz, line 116: I wouldn't put this in terms of nodes -- that means little to consumers. Really, what they're interested in is UI, so asset representation. This goes for the above docs too. I'd just put it in terms of content, which is really what you have here, but it's worth calling out.

Maybe there can be a sentence or two calling out that this'd work (?) for replacing non-asset nodes (like metadata or something), if that's applicable. Not sure there is a great use case for that though.

docs/site/pages/plugins/coroutines.mdx
brocollie08: I like having this here, I imagine people go into every plugin doc page expecting to see docs specific to the platform they're working in, having the platforms here alongside the icon imo highlights it's only available for jvm consumers without forcing people to read what coroutines are

android/demo/README.md
```diff
@@ -25,7 +25,7 @@ bazel run //android:demo
 ```
 
 ```
-bazel mobile-install //android/demo
+bazel run //android/demo:install
```
sugarmanz, line 28: 🥇

sugarmanz (approved):

PR #298, opened 2024-02-14
fix: help to improve docs links with relative path

rafbcampos:
### Description

Use relative paths for the `Help to improve this page` doc's links, otherwise we get the absolute path pointing to the Bazel output.

### Change Type

- [x] `patch`
- [ ] `minor`
- [ ] `major`


KetanReddy (approved):

hborawski (approved):
i saw the bazel links when running locally, didnt realize it was doing it deployed too

PR #302, opened 2024-02-16
Common Types Plugin restoring old dataRefs

lexfm:
### Change Type (required)
- Reverted common-types plugin `dataRefs`. Those can now be dynamically created when generating the dynamic types for the `DSLSchema`.
- Generating and exporting `dataRefs`, `validationRefs` and `DSLSchema` Type from `@player-ui/reference-assets-components`. 

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


KetanReddy:
Can we also export the validations from the `common-type` plugin and export a `DSLSchema` type to tie it all together

KetanReddy (approved):

PR #307, opened 2024-02-26
Android MVVM Module

brocollie08:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`


docs/site/pages/assets/reference.mdx
```diff
@@ -44,12 +44,12 @@ struct MyApp: View {
...
 ```
 
-## Android 
+## Android
 
 Alongside distributing the Android Player framework, there is a reference assets library used as an example of how Player APIs work for building assets. These are intended to serve as a reference, but can be used early on in development as a means of understanding how to work with Player. This asset set uses the shared TypeScript transform functions that the Web and Android players use for their reference assets, ensuring consistent behavior across platforms. These assets can be loaded into Player like so:
 
 ```kotlin
-import com.intuit.player.android.reference.assets.ReferenceAssetsPlugin
+import com.intuit.player`.android.reference.assets.ReferenceAssetsPlugin
```
hborawski, line 52:
```suggestion
import com.intuit.player.android.reference.assets.ReferenceAssetsPlugin
```

android/player/BUILD
```diff
@@ -26,13 +26,20 @@ distribution(
...
     name = "player-tests",
-    srcs = glob(["src/test/java/**"]),
-    associates = [":player-kotlin"],
-    kotlinc_opts = "//jvm:test_options",
-    test_package = "com.intuit.playerui.android",
-    deps = [":player"] + test_deps,
+    srcs = [
+        "src/test/java/com/intuit/playerui/android/AndroidPlayerTest.kt",
+        "src/test/java/com/intuit/playerui/android/AssetContextTest.kt",
+        "src/test/java/com/intuit/playerui/android/BrokenAssetTest.kt"
```
brocollie08, line 34: the first two are not great because the test target should really be in the test package, but it's moved here because of the associates line below, allowing the tests to access internal player API.

the `BrokenAssetTest` is _really_ bad since i had to move it out of the original package, otherwise i couldn't reference it as a source. It's really the same problem trying to access internal API

the main problem here is `:player-kotlin` target is created as an intermediate target by the db android macro, and the visibility isn't marked as public so it can only be used from this package. The end result target does not meet the requirements to be used as an associate

I could either modify the grab rules or we need to write some hack, maybe wrapper library to expose `:player-kotlin`? Is there a better way

android/player/BUILD
```diff
@@ -26,13 +26,26 @@ distribution(
...
+    associates = [":player-kotlin"]
+)
+
+filegroup(
+    name = "manifest",
+    srcs = ["src/main/AndroidManifest.xml"],
+    visibility = ["//visibility:public"]
+)
+
+kt_jvm_library(
```
brocollie08, line 45: this is a terrible workaround for internals in `Constants.kt`. Since the vals are internal but they're used for the test utils in `utils/BUILD`, we make this target to use as a publicly visible associate

brocollie08: in the latest commit this actually encapsulates everything, so then i can move `BrokenAssetTest.kt` file back to where it belongs and get rid of this test target above and place it into the test package. I just don't know if this is the way we want to go

PR #310, opened 2024-03-04
iOS: prefix resource bundles to prevent naming collisions

hborawski:
Prefixed all the resource bundles with `PlayerUI_` since resource bundles for an app have to have globally unique names. This is to reduce the chance another integration in an app would have the same named bundle

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`


PlayerUI.podspec
```diff
@@ -184,7 +184,7 @@ and display it as a SwiftUI view comprised of registered assets.
 
     utils.source_files = 'ios/packages/test-utils-core/Sources/**/*'
```
nancywu1, line 185: i think the ReferenceAssets resource bundle on this file was missed?

nancywu1 (approved):

PR #336, opened 2024-04-11
iOS: expose data controller onUpdate hook

nancywu1:
expose on update hook so Swift plugins can recieve data changes

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/ios/packages/core/Sources/Types/Hooks/DataControllerHooks.swift
+++ b/ios/packages/core/Sources/Types/Hooks/DataControllerHooks.swift
@@ -1,0 +1,29 @@
+//
+//  DataControllerHooks.swift
+//  PlayerUI
+//
+//  Created by Zhao Xia Wu on 2024-04-08.
+//
+
+import Foundation
+import JavaScriptCore
+
+/**
+Hooks that can be tapped into for the DataController
+This lets users tap into events in the JS environment
+*/
+public struct DataControllerHooks {
+    /// Fired when the data changes
+    public var onUpdate: HookDecode<[Update]>
+}
+
+public struct Update: Decodable {
+    /** The updated binding */
+    var binding: BindingInstance
+    /** The old value */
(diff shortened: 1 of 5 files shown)


ios/packages/core/Sources/Types/Hooks/Hook.swift
```diff
@@ -83,6 +83,59 @@ public class Hook2<T, U>: BaseJSHook where T: CreatedFromJSValue, U: CreatedFrom
...
+     - parameters:
+        - hook: A function to run when the JS hook is fired
+     */
+    public func tap(_ hook: @escaping (T, U) -> Void) {
+        let tapMethod: @convention(block) (JSValue?, JSValue?) -> Void = { value, value2 in
+            guard
+                let val = value,
+                let val2 = value2,
+                let hookValue = try? JSONDecoder().decode(T.self, from: val),
+                let hookValue2 = try? JSONDecoder().decode(U.self, from: val2)
```
hborawski, line 130: probably trivial perf, but we can just make one decoder and reuse it

ios/packages/core/Sources/Types/Hooks/DataControllerHooks.swift
```diff
@@ -0,0 +1,29 @@
...
+This lets users tap into events in the JS environment
+*/
+public struct DataControllerHooks {
+    /// Fired when the data changes
+    public var onUpdate: HookDecode<[Update]>
+}
+
+public struct Update: Decodable {
+    /** The updated binding */
+    var binding: BindingInstance
```
hborawski, line 22: can you add a regression test also to make sure that `binding` is decoded properly with a value?

hborawski (approved):

PR #343, opened 2024-04-23
[Android] Reorganize `PlayerFragment` state updates

sugarmanz:
Currently,`playerViewModel.state` collection in `PlayerFragment` is a mix between UI updates and callback handling. This separates each to ensure the latest states take precedence for UI updates and such updates don't block the state callbacks.

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.7.3--canary.343.11262</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.7.3--canary.343.11262
</details>

### DIFF
--- a/android/player/src/main/java/com/intuit/playerui/android/ui/PlayerFragment.kt
+++ b/android/player/src/main/java/com/intuit/playerui/android/ui/PlayerFragment.kt
@@ -8,8 +8,11 @@ import android.view.ViewGroup
 import android.widget.ProgressBar
 import androidx.core.view.doOnLayout
 import androidx.fragment.app.Fragment
+import androidx.lifecycle.Lifecycle.State
 import androidx.lifecycle.LifecycleCoroutineScope
 import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
+import androidx.lifecycle.whenStarted
 import androidx.transition.Transition
 import com.intuit.playerui.android.AndroidPlayer
 import com.intuit.playerui.android.asset.RenderableAsset
@@ -28,6 +31,7 @@ import com.intuit.playerui.android.lifecycle.PlayerViewModel
 import com.intuit.playerui.android.lifecycle.fail
 import com.intuit.playerui.core.experimental.ExperimentalPlayerApi
 import com.intuit.playerui.core.managed.AsyncFlowIterator
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.collectLatest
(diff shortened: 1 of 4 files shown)


scripts/mvn-install.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -u -e -o pipefail
+
+# Find all the maven packages in the repo
+readonly DEPLOY_LABELS=$(bazel query --output=label 'kind("maven_publish rule", //...) - attr("tags", "\[.*do-not-publish.*\]", //...)')
+for pkg in $DEPLOY_LABELS ; do
+  bazel run "$pkg" --define=maven_repo="file://$HOME/.m2/repository"
+done
```
A1shK, line 9: Does this script help us use these packages locally?

android/player/src/main/java/com/intuit/playerui/android/ui/PlayerFragment.kt
```diff
@@ -175,11 +178,14 @@ public abstract class PlayerFragment : Fragment(), ManagedPlayerState.Listener {
...
-                exception.printStackTrace()
-                playerViewModel.fail("Error rendering asset", exception)
+            whenStarted { // TODO: This'll go away when we can call a suspend version of this
+                try {
+                    renderIntoPlayerCanvas(asset, animateTransition)
+                } catch (exception: Exception) {
+                    if (exception is CancellationException) throw exception
+                    exception.printStackTrace()
+                    playerViewModel.fail("Error rendering asset", exception)
+                }
```
A1shK, line 188: why is a CancellationException treated differently?

scripts/mvn-install.sh
sugarmanz: Precisely! Same as doing `gradle publishToMavenLocal` or `mvn install`

android/player/src/main/java/com/intuit/playerui/android/ui/PlayerFragment.kt
sugarmanz: Good question, `CancellationException` occurs if the coroutine is cancelled - so, in the case that the `PlayerFragment` is destroyed, it'll cancel the `PlayerViewModel.state` collection and any jobs spawned from it. If we're in the middle of rendering an asset, it'll throw inside `handleAssetUpdate` and be caught in this `catch`. To ensure we don't hinder cancellation and treat this case as a `PlayerViewModel.fail`, we need to re-throw.

scripts/mvn-install.sh
A1shK: nice!! we needed this. 

Is it possible to add some docs for this in either this PR or another

A1shK (approved):

PR #347, opened 2024-04-24
Parse multi node switch fixv1

mercillo:
When multinodes contained switches; the switch would not evaluate correctly.

```
          id: 'foo',
          fields: [
            staticSwitch: [
              {
                case: '{{foo.baz}}',
                asset: {
                  id: 'input-1',
                  type: 'input',
                },
              },
              {
                case: '{{foo.bar}}',
                asset: {
                  id: 'input-2',
                  type: 'input',
                },
              },
            ],
          ],
``` 

when staticSwitches get evaluated to false; the asset should not live in the AST, however there was a bug that caused staticSwitches not to get read as a switch when they were inside of multinodes.

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+registry=https://registry.npmjs.com
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-registry=https://registry.npmjs.com
--- a/core/player/src/view/parser/index.ts
+++ b/core/player/src/view/parser/index.ts
@@ -128,6 +128,10 @@ export class Parser {
   ): Node.Node | null {
     const nodeType = this.hooks.determineNodeType.call(obj);
 
+    console.log('@@ PARSEOBJECT[obj]: ',obj)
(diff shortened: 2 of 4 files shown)


.npmrc
```diff
@@ -1 +0,0 @@
-registry=https://registry.npmjs.com
```
hborawski, line 1: is this intentional?

core/player/src/view/parser/index.ts
```diff
@@ -121,6 +121,10 @@ export class Parser {
     );
   }
 
+  private hasSwitchKey(localKey: string) {
```
KetanReddy, line 124: Can we move this logic into the switch plugin?

hborawski (approved):

.npmrc
mercillo: not intentional! just added this back

KetanReddy (approved):

PR #350, opened 2024-04-26
Specify some RC flags only for mac

hborawski:
adds a `--config=ci-mac` option to set the CPU/RAM limits for the macOS executor, as it has less memory than the docker executor

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -46,9 +46,8 @@ build:ci --build_metadata=ROLE=CI --workspace_status_command=./scripts/workspace
 build:ci --local_cpu_resources=8
 build:ci --local_ram_resources=15000
 build:ci --config="release"
-# Circle macos.m1.medium.gen1 4 cores 6gb ram
-build:ci-mac --config=ci --local_cpu_resources=4 --local_ram_resources=5000
-coverage:ci-mac --config=ci --local_cpu_resources=4 --local_ram_resources=5000
+# Circle macos.m1.large.gen1 8 cores 12gb ram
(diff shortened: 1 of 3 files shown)


KetanReddy (approved):

mercillo (approved):

.bazelrc
```diff
@@ -46,6 +46,8 @@ build:ci --build_metadata=ROLE=CI --workspace_status_command=./scripts/workspace
 build:ci --local_cpu_resources=8
 build:ci --local_ram_resources=15000
 build:ci --config="release"
+# Circle macos.m1.large.gen1 8 cores 12gb ram
+common:ci-mac --config=ci --local_cpu_resources=8 --local_ram_resources=11000
```
mercillo, line 50: do we have any guidelines/best practices on what we can use so we dont go over budget? Or is this just a by case basis. Knowing me I would just try to use a machine with maxed out ram and cores so builds are fast; but cant imagine how much that would cost

hborawski: Builds use up "credits", and the macOS executors use more credits per minute, so its more case by case, and more important for iOS related things, docker credits are cheap

You can see some credit usage here: https://app.circleci.com/insights/github/player-ui

hborawski: the bazel remote cache should also aid in build times without needing to increase CI machine resources

sugarmanz (approved):

nancywu1 (approved):

PR #356, opened 2024-05-09
Move collection of state onto Dispatchers.Default in PlayerFragment.kt

osquitoOfTheNorth:
### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

#Release Notes
Move collection of player view model state onto Dispatchers.Default in PlayerFragment.kt

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

Background:

Prior to [this]( https://github.com/player-ui/player/pull/343/files  ) change, `handleAssetUpdate` would be called once and only once for each navigation / change in the asset view tree.  After the aforementioned change, there was a certain scenario (navigation to another view in the same flow) where `handleAssetUpdate` would be called twice once for the new view and another time for the previous view.  This resulted in odd scenarios on the UI where it would appear that a user briefly navigated to a new screen before returning to the original screen.  

I noticed that collecting the view state on anything other than Default.Dispatchers caused this change in behaviour, so this PR simply reverts that change from the linked PR + brings back the `intoOnMain`.

### DIFF
--- a/android/player/src/main/java/com/intuit/playerui/android/ui/PlayerFragment.kt
+++ b/android/player/src/main/java/com/intuit/playerui/android/ui/PlayerFragment.kt
@@ -20,6 +20,7 @@ import com.intuit.playerui.android.asset.SuspendableAsset
 import com.intuit.playerui.android.databinding.FallbackViewBinding
 import com.intuit.playerui.android.databinding.PlayerFragmentBinding
 import com.intuit.playerui.android.extensions.into
+import com.intuit.playerui.android.extensions.intoOnMain
 import com.intuit.playerui.android.extensions.transitionInto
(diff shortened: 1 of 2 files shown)


android/player/src/main/java/com/intuit/playerui/android/ui/PlayerFragment.kt
```diff
@@ -91,7 +92,7 @@ public abstract class PlayerFragment : Fragment(), ManagedPlayerState.Listener {
     public abstract val playerViewModel: PlayerViewModel
 
     init {
-        lifecycleScope.launch {
+        lifecycleScope.launch(Dispatchers.Default) {
```
osquitoOfTheNorth, line 95: Why is this preventing the two states from being emitted ? @sugarmanz

PR #366, opened 2024-05-23
add unit tests for reference input transform

hborawski:
- Added basic transform tests for `input` transform

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


plugins/reference-assets/core/src/assets/input/__tests__/transform.test.ts
```diff
@@ -0,0 +1,112 @@
...
+            ],
+          },
+        },
+      },
+    };
+    const { current } = runTransform("input", inputTransform, testFlow, [
+      new CommonTypesPlugin(),
+    ]);
+
+    expect(current?.validation).not.toBeUndefined();
```
mrigankmg, line 104: Do we need this if we are already testing for the object `toHaveProperty("severity", "error");` on line 106?

mrigankmg:
Should we also test for existence and functionality of the `format` and `set` methods?

plugins/reference-assets/core/src/assets/input/__tests__/transform.test.ts
hborawski: I think it just makes the test more easily identifiable if it fails, this will fail for the object not being defined, instead of saying it cant find a property

PR #367, opened 2024-05-24
Choice Asset

mrigankmg:
### Changes Made

- Added core support for the new reference asset - choice. (From https://github.com/player-ui/player/issues/181) "A choice asset represents a single selection choice, often displayed as radio buttons in a web context. This will allow users to test out more complex flows than just inputs + buttons."

#### Additional Changes:

- Extended `runTransform` to accept an `additionalPlugins` parameter to simplify testing for users of Player.

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/plugins/reference-assets/components/src/index.tsx
+++ b/plugins/reference-assets/components/src/index.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import type {
+  WithChildren,
   AssetPropsWithChildren,
   BindingTemplateInstance,
   ExpressionTemplateInstance,
@@ -7,12 +8,16 @@ import type {
   DataTypeReference,
   DataTypeRefs,
   ValidatorFunctionRefs,
+  WithPlayerTypes,
 } from "@player-tools/dsl";
 import {
   createSlot,
   Asset,
   View,
   getObjectReferences,
+  toJsonProperties,
+  ObjectWithIndexTracking,
+  GeneratedIDProperty,
 } from "@player-tools/dsl";
 import type { Asset as AssetType } from "@player-ui/player";
 import type {
@@ -22,6 +27,8 @@ import type {
   InfoAsset,
   InputAsset,
   ImageAsset,
+  ChoiceAsset,
+  ChoiceItem as ChoiceItemType,
 } from "@player-ui/reference-assets-plugin";
 import { dataTypes, validators } from "@player-ui/common-types-plugin";
 
@@ -64,21 +71,23 @@ const CollectionComp = (props: AssetPropsWithChildren<AssetType>) => {
 };
 
 /** A utility for quickly creating named slots using the text and collection factories */
-const slotFactory = (name: string, isArray = false) =>
+const slotFactory = (name: string, isArray = false, wrapInAsset = true) =>
   createSlot({
     name,
     TextComp: Text,
     CollectionComp,
     isArray,
-    wrapInAsset: true,
+    wrapInAsset,
   });
 
 export const LabelSlot = slotFactory("label");
 export const ValueSlot = slotFactory("value");
 export const TitleSlot = slotFactory("title");
+export const NoteSlot = slotFactory("note");
 export const SubtitleSlot = slotFactory("subtitle");
 export const ActionsSlot = slotFactory("actions", true);
 export const PrimaryInfoSlot = slotFactory("primaryInfo");
+export const ItemsSlot = slotFactory("items", true, false);
 
 Collection.Values = createSlot({
   name: "values",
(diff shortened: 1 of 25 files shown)


plugins/reference-assets/components/src/index.tsx
```diff
@@ -22,9 +29,15 @@ import type {
   InfoAsset,
   InputAsset,
   ImageAsset,
+  ChoiceAsset,
+  ChoiceItem as ChoiceItemType,
 } from "@player-ui/reference-assets-plugin";
 import { dataTypes, validators } from "@player-ui/common-types-plugin";
 
+type WithApplicability<T = Record<string, unknown>> = T & PlayerApplicability;
```
KetanReddy, line 37: Given how often we will need to reuse this, this should probably be a utility that comes from the DSL package. We should also expose a singular type that provides it with the `WithTemplateTypes` type to make it easy to use (`WithPlayerTypes` maybe?)

plugins/reference-assets/components/src/index.tsx
```diff
@@ -39,6 +52,17 @@ export type DSLSchema = PlayerDSLSchema<
   DataTypeReference<typeof dataTypes, ValidatorFunctionRefs<typeof validators>>
 >;
 
+/** An object wrapper that tracks it's index from a parent */
+const ObjectWithIndexTracking = (props: WithChildren) => {
```
KetanReddy, line 56: Same comment as `WithApplicability` this should also come from the DSL package since we provide the already `OptionalIDSuffixProvider`

plugins/reference-assets/components/src/index.tsx
```diff
@@ -140,3 +166,49 @@ Info.Subtitle = SubtitleSlot;
...
+Choice.Title = TitleSlot;
+Choice.Note = NoteSlot;
+Choice.Items = ItemsSlot;
+
+/**
+ * Comp for a choice item
+ */
+const ChoiceItem = (
+  props: WithChildren<
+    WithApplicability<WithTemplateTypes<Optional<ChoiceItemType, "id">>>
```
KetanReddy, line 194: Instead of defining and using the `Optional` type, I would use the same pattern we use with `Omit` to pull out the `binding` property on the Choice asset to just make `id` conditional

plugins/reference-assets/core/src/assets/choice/types.ts
```diff
@@ -0,0 +1,73 @@
...
+  items?: Array<ChoiceItem>;
+
+  /** Additional optional data to assist when a choice item is selected */
+  metaData?: {
+    /** Additional data to beacon */
+    beacon?: BeaconDataType;
+  };
+}
+
+type ValueType = string | number | boolean | undefined | null;
```
KetanReddy, line 35: 1. There might be portability issues if this type isn't exported
2. `undefined` can probably be excluded from the union if  `value` is made optional on both authored and transformed types

plugins/reference-assets/components/src/index.tsx
hborawski: Tools are in https://github.com/player-ui/tools

DSL specifically is in https://github.com/player-ui/tools/tree/main/language/dsl

plugins/reference-assets/core/src/assets/choice/types.ts
```diff
@@ -0,0 +1,73 @@
...
+  /** The location in the data-model to store the data */
+  binding?: Binding;
+
+  items?: Array<ChoiceItem>;
+
+  /** Additional optional data to assist when a choice item is selected */
+  metaData?: {
+    /** Additional data to beacon */
+    beacon?: BeaconDataType;
+  };
```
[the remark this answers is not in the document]
mrigankmg: Unsure what that would look like

plugins/reference-assets/components/src/index.tsx
mrigankmg: > We should also expose a singular type that provides it with the WithTemplateTypes type to make it easy to use

Could I get some more clarification on this?

plugins/reference-assets/core/src/assets/choice/types.ts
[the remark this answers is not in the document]
hborawski: Well given that `BeaconMetaData` defines the exact same `metaData` object, i would imagine you can:

```suggestion
  metaData?: BeaconMetaData;
```

or have `ChoiceAsset` implement `AssetBeaconInfo`?

hborawski (approved):

PR #371, opened 2024-05-31
Player 8901 - Fix typings for onAsyncNode hook

sakuntala-motukuri:
### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`

Test case passes locally:
image

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/plugins/async-node/core/src/index.test.ts
+++ b/plugins/async-node/core/src/index.test.ts
@@ -382,7 +382,7 @@ test('should call onAsyncNode hook when async node is encountered', async () => 
 
   // Call the deferredResolve with type 'Async'
   if (deferredResolve) {
-    //we are passing the objects variable to the deferredResolve function and then expecting the type of the object to be 'Async'
+    // we are passing the objects variable to the deferredResolve function and then expecting the type of the object to be 'Async'
     const objects = [
       {
         asset: {
@@ -393,12 +393,13 @@ test('should call onAsyncNode hook when async node is encountered', async () => 
       },
       {
         id: 'another-async',
-        type: 'Async',
+        type: NodeType.Async,
         async: true,
       },
     ];
     deferredResolve(objects);
-    const types = objects.map(obj => obj.asset ? obj.asset.type : obj.type);
-    expect(types).toContain('Async');
+    const types = objects.map((obj) => (obj.asset ? obj.asset.type : obj.type));
+    // eslint-disable-next-line jest/no-conditional-expect
+    expect(types).toContain(NodeType.Async);
   }
 });
--- a/plugins/async-node/core/src/index.test.ts
+++ b/plugins/async-node/core/src/index.test.ts
@@ -1,9 +1,7 @@
 import type { Node, InProgressState } from '@player-ui/player';
-import { NodeType, Player } from '@player-ui/player';
+import { Player } from '@player-ui/player';
 import { waitFor } from '@testing-library/react';
 import { AsyncNodePlugin } from './index';
-import { any } from 'parsimmon';
-import { type } from 'node:os';
 
 jest.useFakeTimers();
 
@@ -45,7 +43,7 @@ test('replaces async nodes with provided node', async () => {
 
   let deferredResolve: ((value: any) => void) | undefined;
 
-  plugin.hooks.onAsyncNode.tap('test', async (node: Node.Async) => {
+  plugin.hooks.onAsyncNode.tap('test', async (node: Node.Node) => {
     return new Promise((resolve) => {
       deferredResolve = resolve;
     });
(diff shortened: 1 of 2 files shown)


plugins/async-node/core/src/index.test.ts
```diff
@@ -340,3 +342,64 @@ test('replaces async nodes with chained multiNodes singular', async () => {
...
+
+test('should call onAsyncNode hook when async node is encountered', async () => {
+  const plugin = new AsyncNodePlugin();
+
+  let deferredResolve: ((value: any) => void) | undefined;
+
+  // Call the tap method of the onAsyncNode hook
+ plugin.hooks.onAsyncNode.tap('test', async (node: Node.Async) => {
+    return new Promise((resolve) => {
+      deferredResolve = resolve;
```
brocollie08, line 354: I don't think we'll really need to do anything with this `deferredResolve`, since we don't really need to resolve with any content later as we're just checking the node typing when the hook is first called.

plugins/async-node/core/src/index.test.ts
```diff
@@ -340,3 +342,64 @@ test('replaces async nodes with chained multiNodes singular', async () => {
...
+  player.start(basicFRFWithActions as any);
+
+  let view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view).toBeDefined();
+  expect(view?.actions[1]).toBeUndefined();
+
+  await waitFor(() => {
+    expect(updateNumber).toBe(1);
```
brocollie08, line 379: you _might_ be able to get away with skipping this as well, which would get rid of the `onUpdate` tap

plugins/async-node/core/src/index.test.ts
```diff
@@ -340,3 +342,64 @@ test('replaces async nodes with chained multiNodes singular', async () => {
...
+    expect(deferredResolve).toBeDefined();
+  });
+
+  // Call the deferredResolve with type 'Async'
+  if (deferredResolve) {
```
brocollie08, line 384: this is likely not necessary for the reason stated above, we don't need to resolve the async node with anything for this test, we just need to check the tap matches what's [here](https://github.com/player-ui/player/blob/main/plugins/async-node/core/src/index.ts#L56-L63) along with the properties

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: redone the entire test case

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: redone the entire test case

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: redone the entire test case

plugins/async-node/core/src/index.test.ts
```diff
@@ -340,3 +342,29 @@ test('replaces async nodes with chained multiNodes singular', async () => {
...
 });
+
+test('should call onAsyncNode hook when async node is encountered', async () => {
+  const plugin = new AsyncNodePlugin();
+  // Mock function to check the node passed to the hook
+  const mockFn = jest.fn();
+
+  // Call the tap method of the onAsyncNode hook
+  plugin.hooks.onAsyncNode.tap('test', async (node: Node.Async) => {
+    mockFn(node);
```
brocollie08, line 353: is this needed? are we not able to grab the arg here and verify that it matches up to [this](https://github.com/player-ui/player/pull/371/files#diff-fbc47a3e301bdf595a6d686ded538fe6767e534af8821c421f97dc56264d1332R23-R25)

sakuntala-motukuri: Yes this is needed because Inside the function, the first line is a call to mockFn with the node as an argument. mockFn is a mock function, typically used to track calls to the function and the arguments it was called with. Im using this mock function in line no 364 and 365 to verify that the actual response received as in when the async node has been identified if it matches with the mock response or not. This how we can actually verify the response received

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,6 +1,8 @@
 import type { Node, InProgressState } from '@player-ui/player';
-import { Player } from '@player-ui/player';
+import { NodeType, Player } from '@player-ui/player';
 import { waitFor } from '@testing-library/react';
+import { any } from 'parsimmon';
+import { type } from 'node:os';
```
hborawski, line 5:
```suggestion
```

these imports seem like incorrect additions from an IDE shortcut as they both mirror TS default types/keywords

brocollie08 (approved):

PR #382, opened 2024-06-11
Fix binding tracking for validations within nested multi-nodes

tmarmer:
### What is being fixed
Because the `afterNodeUpdate` tap from the resolver doesn't always pass the original parent of the node, the `addToTree` logic would fail to recreate the original tree structure of the nodes that were nested in multiple multi-nodes. This caused some bindings not to get tracked properly and allow the navigation controller to proceed even when there were blocking errors present. To fix this the following changes were made:

- During the `afterNodeUpdate` tap in `binding-tracker`, the `currentBindingTree` now uses the resolved nodes as the keys, and uses the `children` and `values` properties to get the child nodes instead of tracking it using the old `addToTree` function.
- In order to be compatible with the other taps and maps that still use the original node instead of the resolved one, the `resolvedNodeMap` was added to map the resolved nodes back to their originals. The `lastComputedBindingTree` is then setup by mapping the keys from `currentBindingTree` back to their original nodes. 

### Change Type (required)

- [X] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [X] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/core/player/src/controllers/validation/binding-tracker.ts
+++ b/core/player/src/controllers/validation/binding-tracker.ts
@@ -1,5 +1,6 @@
 import type { Validation } from '@player-ui/types';
 import type { ViewPlugin, Resolver, Node, ViewInstance } from '../../view';
+import { NodeType } from '../../view';
 import type {
   BindingInstance,
   BindingLike,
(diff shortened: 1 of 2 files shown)


core/player/src/controllers/validation/binding-tracker.ts
```diff
@@ -214,46 +205,66 @@ export class ValidationBindingTrackerViewPlugin
       };
     });
 
-    resolver.hooks.afterNodeUpdate.tap(CONTEXT, (node, parent, update) => {
-      if (parent) {
-        addToTree(node, parent);
-      }
+    resolver.hooks.afterNodeUpdate.tap(
+      CONTEXT,
```
KetanReddy, line 209: Minor but we usually reserve all capital variable names for constants

hborawski: it is a constant, already defined further up -> https://github.com/player-ui/player/blob/main/core/player/src/controllers/validation/binding-tracker.ts#L11

KetanReddy: Ah, my bad. I forgot the first argument of a tap was the tap id. In my opinion we still might want to change that to be something more reflective of it being an ID and not context that comes in with the hook but thats definitely out of scope of this.

KetanReddy (approved):

PR #383, opened 2024-06-11
Logs cleanup of async node plugin and its resolver and changed test data

sakuntala-motukuri:
Logs cleanup of async node plugin and its resolver and changed test data

### Change Type (required)

Modified the comments and logs the following files:

core/player/src/view/resolver/index.ts
plugins/async-node/core/src/index.ts
plugins/async-node/core/src/index.test.ts

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/core/player/src/view/resolver/index.ts
+++ b/core/player/src/view/resolver/index.ts
@@ -24,11 +24,14 @@ export * from './types';
 export * from './utils';
 
 interface NodeUpdate extends Resolve.ResolvedNode {
-  /** A flag to track if a node has changed since the last resolution */
+  /** A flag to track if a node has changed since the last resolution and to resolve the updated view*/
   updated: boolean;
 }
 
-/** Add model context to the data model */
+/** Adding model context to the data model allows for more flexibility and extensibility,
+    * as it enables the data model to be aware of its environment and its state.
+    * This can be useful in various scenarios,
+    * such as when dealing with complex data structures or when working with multiple data sources. */
 const withContext = (model: DataModelWithParser): DataModelWithParser => {
   return {
     get: (binding: BindingLike, options?: DataModelOptions): any => {
@@ -185,8 +188,6 @@ export class Resolver {
 
     if (id) {
       if (this.idCache.has(id)) {
-        // Only log this conflict once to cut down on noise
-        // May want to swap this to logging when we first see the id -- which may not be the first render
         if (isFirstUpdate) {
           if (node.type === NodeType.Asset || node.type === NodeType.View) {
             this.logger?.error(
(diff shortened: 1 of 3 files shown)


core/player/src/view/resolver/index.ts
```diff
@@ -24,11 +24,16 @@ export * from './types';
 export * from './utils';
 
 interface NodeUpdate extends Resolve.ResolvedNode {
-  /** A flag to track if a node has changed since the last resolution */
+  /** A flag to track if a node has changed since the last resolution and to resolve the updated view */
```
brocollie08, line 27: where is this from?

plugins/async-node/core/src/index.ts
```diff
@@ -12,8 +12,8 @@ import { omit } from 'timm';
 export * from './types';
 
 /**
- * Async node plugin used to resolve async nodes in the content
- * If an async node is present, allow users to provide a replacement node to be rendered when ready
+ * An async player node plugin used to resolve async nodes in the content
+ * It automatically keeps track of the async node (if present), and provides a replacement node to be rendered when ready.
```
brocollie08, line 16: this doesn't actually provide anything, it allows the consumers to provide a replacement

core/player/src/view/resolver/index.ts
```diff
@@ -185,21 +190,17 @@ export class Resolver {
...
-        // Only log this conflict once to cut down on noise
-        // May want to swap this to logging when we first see the id -- which may not be the first render
         if (isFirstUpdate) {
           if (node.type === NodeType.Asset || node.type === NodeType.View) {
             this.logger?.error(
               `Cache conflict: Found Asset/View nodes that have conflicting ids: ${id}, may cause cache issues.`
             );
           } else if (node.type === NodeType.Value) {
             this.logger?.info(
-              `Cache conflict: Found Value nodes that have conflicting ids: ${id}, may cause cache issues. To improve performance make value node IDs globally unique.`
```
brocollie08, line 197: i think the cache conflict here is important, it's the duplicate id that would be causing potential issues

core/player/src/view/resolver/index.ts
```diff
@@ -185,21 +190,17 @@ export class Resolver {
 
     if (id) {
       if (this.idCache.has(id)) {
-        // Only log this conflict once to cut down on noise
```
brocollie08, line 188: i'd keep these comments and the ones at the bottom

brocollie08:
i realized the extra logs are actually in the bazel 6 branch, inside the async node plugin, and not in main, I can just clean those up

core/player/src/view/resolver/index.ts
sakuntala-motukuri: So i have removed the cache conflict as this is providing an info and not the error message. In the error we are already giving the same info ... Just a thought that how would both the error and the info be the same. Nevertheless, I can revert it.

PR #389, opened 2024-06-14
This PR fixes the Ability to resolve with nothing issue

sakuntala-motukuri:
This PR fixes the Ability to resolve with nothing issue

image

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/plugins/async-node/core/src/index.test.ts
+++ b/plugins/async-node/core/src/index.test.ts
@@ -91,7 +91,6 @@ const asyncNodeTest = async (resolvedValue: any, expectedActionType: string) => 
 
   expect(view?.actions[0].asset.type).toBe('action');
   expect(view?.actions.length).toBe(1);
-  expect(updateNumber).toBe(2); // Replace with the actual expected number of updates
 
   viewInstance.update();
 
@@ -104,7 +103,6 @@ const asyncNodeTest = async (resolvedValue: any, expectedActionType: string) => 
 
   expect(view?.actions[0].asset.type).toBe("action");
   expect(view?.actions.length).toBe(1);
-  expect(updateNumber).toBe(3);
 };
 
 test('should return current node view when the resolved node is null', async () => {
--- a/plugins/async-node/core/src/index.test.ts
+++ b/plugins/async-node/core/src/index.test.ts
@@ -95,7 +95,6 @@ const asyncNodeTest = async (resolvedValue: any, expectedActionType: string) => 
 
   viewInstance.update();
 
-  // Updated this to 2 as the resolver is getting updated only twice and the onAsyncNode is not getting called as assumed and the test is failing when we expect it to be 3
   await waitFor(() => {
     expect(updateNumber).toBe(3);
   });
--- a/plugins/async-node/core/src/index.test.ts
+++ b/plugins/async-node/core/src/index.test.ts
@@ -1,50 +1,50 @@
 import { expect, test } from "vitest";
-import { Node, InProgressState, Resolver, ViewInstance } from '@player-ui/player';
-import { Player } from '@player-ui/player';
-import { waitFor } from '@testing-library/react';
-import { AsyncNodePlugin, AsyncNodePluginPlugin } from './index';
+import { Node, InProgressState, ViewInstance } from "@player-ui/player";
+import { Player } from "@player-ui/player";
+import { waitFor } from "@testing-library/react";
+import { AsyncNodePlugin, AsyncNodePluginPlugin } from "./index";
 
 const basicFRFWithActions = {
-  id: 'test-flow',
+  id: "test-flow",
   views: [
     {
-      id: 'my-view',
+      id: "my-view",
       actions: [
         {
           asset: {
-            id: 'action-0',
-            type: 'action',
-            value: '{{foo.bar}}'
-          }
+            id: "action-0",
+            type: "action",
+            value: "{{foo.bar}}",
+          },
         },
         {
-          id: 'uhh',
-          async: 'true'
-        }
-      ]
-    }
+          id: "nodeId",
+          async: "true",
+        },
+      ],
+    },
   ],
   navigation: {
-    BEGIN: 'FLOW_1',
+    BEGIN: "FLOW_1",
     FLOW_1: {
-      startState: 'VIEW_1',
+      startState: "VIEW_1",
       VIEW_1: {
-        state_type: 'VIEW',
-        ref: 'my-view',
-        transitions: {}
-      }
-    }
-  }
+        state_type: "VIEW",
+        ref: "my-view",
+        transitions: {},
+      },
+    },
+  },
 };
 
-const asyncNodeTest = async (resolvedValue: any, expectedActionType: string) => {
+const asyncNodeTest = async (resolvedValue: any) => {
   const plugin = new AsyncNodePlugin({
-    plugins: [new AsyncNodePluginPlugin()]
+    plugins: [new AsyncNodePluginPlugin()],
   });
 
   let deferredResolve: ((value: any) => void) | undefined;
 
-  plugin.hooks.onAsyncNode.tap('test', async (node: Node.Node) => {
+  plugin.hooks.onAsyncNode.tap("test", async (node: Node.Node) => {
     return new Promise((resolve) => {
       deferredResolve = resolve; // Promise would be resolved only once
     });
@@ -54,11 +54,12 @@ const asyncNodeTest = async (resolvedValue: any, expectedActionType: string) => 
 
   const player = new Player({ plugins: [plugin] });
 
-  let viewInstance : ViewInstance;
-  player.hooks.viewController.tap('async-node-test', (vc) => {
-    vc.hooks.view.tap('async-node-test', (view) => {
+  let viewInstance: ViewInstance | undefined;
+
+  player.hooks.viewController.tap("async-node-test", (vc) => {
+    vc.hooks.view.tap("async-node-test", (view) => {
       viewInstance = view;
-      view.hooks.onUpdate.tap('async-node-test', () => {
+      view.hooks.onUpdate.tap("async-node-test", () => {
         updateNumber++;
       });
     });
@@ -70,7 +71,7 @@ const asyncNodeTest = async (resolvedValue: any, expectedActionType: string) => 
     ?.lastUpdate;
 
   expect(view).toBeDefined();
-  expect(view?.actions[0].asset.type).toBe('action');
+  expect(view?.actions[0].asset.type).toBe("action");
   expect(view?.actions[1]).toBeUndefined();
 
   await waitFor(() => {
@@ -89,7 +90,7 @@ const asyncNodeTest = async (resolvedValue: any, expectedActionType: string) => 
   view = (player.getState() as InProgressState).controllers.view.currentView
     ?.lastUpdate;
 
-  expect(view?.actions[0].asset.type).toBe('action');
+  expect(view?.actions[0].asset.type).toBe("action");
   expect(view?.actions.length).toBe(1);
 
   viewInstance.update();
@@ -105,12 +106,12 @@ const asyncNodeTest = async (resolvedValue: any, expectedActionType: string) => 
   expect(view?.actions.length).toBe(1);
 };
 
-test('should return current node view when the resolved node is null', async () => {
-  await asyncNodeTest(null, undefined);
+test("should return current node view when the resolved node is null", async () => {
+  await asyncNodeTest(null);
 });
 
-test('should return current node view when the resolved node is undefined', async () => {
-  await asyncNodeTest(undefined, undefined);
+test("should return current node view when the resolved node is undefined", async () => {
+  await asyncNodeTest(undefined);
 });
 
 test("replaces async nodes with provided node", async () => {
--- a/plugins/async-node/core/src/index.test.ts
+++ b/plugins/async-node/core/src/index.test.ts
@@ -37,6 +37,33 @@ const basicFRFWithActions = {
   },
 };
 
+  test('should return null when onAsyncNode hook does not return a result', async () => {
+    const plugin = new AsyncNodePlugin({
+      plugins: [new AsyncNodePluginPlugin()],
+    });
+
+    plugin.hooks.onAsyncNode.tap("test", async (node: Node.Node) => {
+      return new Promise((resolve) => {
+        resolve(undefined);
+      });
+    });
+
+    const player = new Player({ plugins: [plugin] });
+
+    player.start(basicFRFWithActions as any);
+
+    let view = (player.getState() as InProgressState).controllers.view.currentView
+      ?.lastUpdate;
+
+    expect(view).toBeDefined();
+    expect(view?.actions[1]).toBeUndefined();
+
+    await waitFor(() => {
+      // Check that the original node is still present
+      expect(view?.actions[1]).toBeUndefined();
+    });
+  });
+
 test("replaces async nodes with provided node", async () => {
   const plugin = new AsyncNodePlugin({
     plugins: [new AsyncNodePluginPlugin()],
--- a/plugins/async-node/core/src/index.test.ts
+++ b/plugins/async-node/core/src/index.test.ts
@@ -113,14 +113,14 @@ const asyncNodeTest = async (resolvedValue: any, expectedActionType: string) => 
     expect(beforeResolveCalled).toBe(true);
   });
 
-  // Now we will force update the view
+  // Now we will force update the view using the resolver instance
   const thirdForcedUpdate = resolverInstance.update();
 
   if (deferredResolve) {
     deferredResolve(thirdForcedUpdate);
   }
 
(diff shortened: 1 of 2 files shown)


plugins/async-node/core/src/index.ts
```diff
@@ -166,11 +166,13 @@ export class AsyncNodePluginPlugin implements AsyncNodeViewPlugin {
               options.parseNode && result
                 ? options.parseNode(result)
                 : undefined;
-
             if (parsedNode) {
               this.resolvedMapping.set(node.id, parsedNode);
-              view.updateAsync();
+            } else {
+              // If parsedNode is undefined, update the resolvedMapping with the existing node (supporting null | undefined) as of now
+              this.resolvedMapping.set(node.id, node);
```
brocollie08, line 173: the existing node is an async node, correct? Does this not trigger additional rounds of the hook calls?

sakuntala-motukuri: Good catch ! How about this .. 
              if (!this.resolvedMapping.has(node.id)) {
                this.resolvedMapping.set(node.id, node);
              } // This is to ensure that the node is not resolved again

plugins/async-node/core/src/index.ts
```diff
@@ -166,11 +166,16 @@ export class AsyncNodePluginPlugin implements AsyncNodeViewPlugin {
...
                 ? options.parseNode(result)
                 : undefined;
-
             if (parsedNode) {
               this.resolvedMapping.set(node.id, parsedNode);
-              view.updateAsync();
+            } else {
+              // If parsedNode is undefined, update the resolvedMapping with the existing node (supporting null | undefined) as of now
+              // This is to ensure that the node is not resolved again
+              if (!this.resolvedMapping.has(node.id)) {
```
brocollie08, line 174: This plugin, if registered, will live as long as the player instance it's registered to is still active.
That means as we repeated update UI, this check will only be true once.

I'm not 100% sure your original solution _won't_ work, i'd write some tests to make sure the solution is okay through multiple updates

sakuntala-motukuri: Changed the business logic to such the node is resolved irrespective of the type which ensures that the consumer can expect the node to be resolved irrespective of the input they provide us with. Also updated respective test case such that we ensure nothing is resolved and also that the resolution happens only once and after the promise has been resolved.

plugins/async-node/core/src/index.test.ts
```diff
@@ -37,6 +37,68 @@ const basicFRFWithActions = {
...
+  await waitFor(() => {
+    expect(deferredResolve).toBeDefined();
+  });
+
+  if (deferredResolve) {
+    deferredResolve(resolvedValue);
+  }
+
+  await waitFor(() => {
+    expect(updateNumber).toBe(1);
```
brocollie08, line 83: i think this number will be 1 at this point without waiting.

https://github.com/player-ui/player/pull/389/files#diff-fbc47a3e301bdf595a6d686ded538fe6767e534af8821c421f97dc56264d1332R60
this will change the 0 to a 1 on the first update with the async node unresolved, so i think you'd want to wait until this is 2, that means the async node has been resolved and the tree has been updated again

plugins/async-node/core/src/index.ts
brocollie08: I think we should check that returning the same async node doesn't impact future updates, I think you could do so by manually calling the `update` function in the resolver, see what happens if you try to update again after the resolve with `null`/`undefined` has been processed

sakuntala-motukuri: I think you're not looking at the updated code .. This is the updated code 
image

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: ok I reverted the test case as it is

plugins/async-node/core/src/index.test.ts
```diff
@@ -37,6 +38,90 @@ const basicFRFWithActions = {
...
+      view.hooks.onUpdate.tap('async-node-test', () => {
+        updateNumber++;
+      });
+    });
+  });
+
+  player.hooks.viewController.tap('async-node-test', (vc) => {
+    vc.hooks.view.tap('async-node-test', (view) => {
+      view.hooks.resolver.tap('async-node-test', () => {
+        updateNumber++;
```
brocollie08, line 69: what is the `updateNumber`? Originally I had intended for `updateNumber` to represent how many times the tree is resolved and an update is provided for the UI, but if we're updating `updateNumber` here, then it's not longer following the same semantic intent.

plugins/async-node/core/src/index.test.ts
```diff
@@ -37,6 +38,90 @@ const basicFRFWithActions = {
...
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe('action');
+  expect(view?.actions[1]?.asset.type).toBe(expectedActionType);
+  expect(view?.actions[2]?.asset.type).toBe(expectedActionType);
+
+  await waitFor(() => {
+    expect(updateNumber).toBe(3);
```
brocollie08, line 104: going by the original intent of this variable, this update number should be 2 at this point, since it's the first replacement of an async node (triggering a second update)

plugins/async-node/core/src/index.test.ts
brocollie08: following our slack conversation, this resolver is supposed to be used to force a third update through the `update` function, forcing the tree to be resolved again to make sure the replacement of an async node with another one is "safe"

In that process, I believe `onAsyncNode` should be called again, but it won't go to the users

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,147 @@
...
+  await waitFor(() => {
+    expect(updateNumber).toBe(2);
+  });
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe("action");
+  // updateOnAsyncCounter is also getting updated only once , I think there is no need of this variable
+  expect(updateOnAsyncCounter).toBe(1);
```
brocollie08, line 133: you're right with this comment

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,147 @@
...
+
+  // Updated this to 2 as the resolver is getting updated only twice and the onAsyncNode is not getting called as assumed and the test is failing when we expect it to be 3
+  await waitFor(() => {
+    expect(updateNumber).toBe(2);
+  });
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe("action");
```
brocollie08, line 131: at this point, we should also check that there is only 1 action in the array

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,147 @@
...
+  await waitFor(() => {
+    expect(resolverInstance).toBeDefined();
+    expect(beforeResolveCalled).toBe(true);
+  });
+
+  // Now we will force update the view using the resolver instance
+  const thirdForcedUpdate = resolverInstance.update();
+
+  if (deferredResolve) {
+    deferredResolve(thirdForcedUpdate);
```
brocollie08, line 120: I'm not sure this will actually update anything, nothing is waiting for this to resolve, right?

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,147 @@
...
+  }
+
+  // Updated this to 2 as the resolver is getting updated only twice and the onAsyncNode is not getting called as assumed and the test is failing when we expect it to be 3
+  await waitFor(() => {
+    expect(updateNumber).toBe(2);
```
brocollie08, line 125: this is likely incorrect and _should_ be 3

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,147 @@
...
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe('action');
+  expect(updateNumber).toBe(2); // Replace with the actual expected number of updates
+
+  await waitFor(() => {
+    expect(resolverInstance).toBeDefined();
+    expect(beforeResolveCalled).toBe(true);
```
brocollie08, line 113: this is an unnecessary check, `beforeResolve` should be called on the path to every update

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,147 @@
...
+  }
+
+  await waitFor(() => {
+    expect(updateNumber).toBe(2);
+  });
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe('action');
```
brocollie08, line 108: would also check for other actions here

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,134 @@
...
+  // Updated this to 2 as the resolver is getting updated only twice and the onAsyncNode is not getting called as assumed and the test is failing when we expect it to be 3
+  await waitFor(() => {
+    expect(updateNumber).toBe(2);
+  });
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe("action");
+  expect(view?.actions[1]?.asset.type).toBe(expectedActionType);
```
brocollie08, line 120: this passes but not for the reason intended. This `actions` array as we have checked the other day is only of length 1, because the replacement node never makes it to the view level.

the ideal check here would be the above, rather than checking the asset type of the 2nd item in that array

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,134 @@
...
+
+  await waitFor(() => {
+    expect(updateNumber).toBe(2);
+  });
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe('action');
+  expect(view?.actions[1]?.asset.type).toBe(expectedActionType);
```
brocollie08, line 101: refer to comment below re. actions array

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,134 @@
...
+  player.hooks.viewController.tap('async-node-test', (vc) => {
+    vc.hooks.view.tap('async-node-test', (view) => {
+      view.hooks.onUpdate.tap('async-node-test', () => {
+        updateNumber++;
+      });
+    });
+  });
+
+  player.hooks.viewController.tap("async-node-test", (vc) => {
+    vc.hooks.view.tap("async-node-test", (view) => {
```
brocollie08, line 68: 67-68 is the same as 59-60. You don't need to tap it twice, you can accomplish all of this within one nested tap.

brocollie08: the resolverInstance here as we've talked about doesn't accomplish what we want it to, which is to call the `onUpdate` hook, the `onUpdate` hook is exclusively called within the bigger scope of the view update process as the last step. 

this means if we want to make sure that the view is updated again through a 3rd manual update through `onUpdate` hook's counter, we need to update it using a `ViewInstance` that can be grabbed [here](https://github.com/player-ui/player/pull/389/files#diff-fbc47a3e301bdf595a6d686ded538fe6767e534af8821c421f97dc56264d1332R60) and calling update on that instance

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,134 @@
...
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe('action');
+  expect(view?.actions[1]?.asset.type).toBe(expectedActionType);
+  expect(updateNumber).toBe(2); // Replace with the actual expected number of updates
+
+  await waitFor(() => {
+    expect(resolverInstance).toBeDefined();
```
brocollie08, line 105: this isn't a meaningful check, [this](https://github.com/player-ui/player/pull/389/files#diff-fbc47a3e301bdf595a6d686ded538fe6767e534af8821c421f97dc56264d1332R102) makes more sense to wait for here, 1 it verifies that the update count has increased, 2 it's not an immediate check pass as `expect(resolverInstance).toBeDefined();`

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,134 @@
...
+  });
+
+  // Now we will force update the view using the resolver instance
+  resolverInstance.update();
```
brocollie08, line 109: this is where you would update using the view instead of the resolver, since if you search for `onUpdate.call` in the code, you'll only find it in the view update scope, not the resolver update scope

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: removed

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: removed

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: removed

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: removed

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,120 @@
...
+  await waitFor(() => {
+    expect(updateNumber).toBe(3);
+  });
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe("action");
+  expect(view?.actions.length).toBe(1);
+  expect(updateNumber).toBe(3);
```
brocollie08, line 107: you can remove this because it's implied [here](https://github.com/player-ui/player/pull/389/files#diff-fbc47a3e301bdf595a6d686ded538fe6767e534af8821c421f97dc56264d1332R99)

plugins/async-node/core/src/index.test.ts
```diff
@@ -1,42 +1,120 @@
...
+  await waitFor(() => {
+    expect(updateNumber).toBe(2);
+  });
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe('action');
+  expect(view?.actions.length).toBe(1);
+  expect(updateNumber).toBe(2); // Replace with the actual expected number of updates
```
brocollie08, line 94: can remove

plugins/async-node/core/src/index.ts
```diff
@@ -169,8 +169,10 @@ export class AsyncNodePluginPlugin implements AsyncNodeViewPlugin {
 
             if (parsedNode) {
               this.resolvedMapping.set(node.id, parsedNode);
-              view.updateAsync();
+            } else {
+              this.resolvedMapping.set(node.id, node);
```
brocollie08, line 173: would this work instead of if/else
```this.resolvedMapping.set(node.id, parsedNode? parsedNode : node)```

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.ts
sakuntala-motukuri: done

brocollie08 (approved):

PR #390, opened 2024-06-17
iOS: use updated ios pipeline macro with uitest target

nancywu1:
* delete existing use of `ios_ui_test` rule for uitest since its covered in macro now

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


tools/ios/util.bzl
```diff
@@ -48,6 +48,7 @@ def swiftui_plugin(name, resources = [], deps = [], test_deps = []):
       hasUnitTests = False, 
       hasViewInspectorTests = True,
       test_host = test_host,
+      hasUITests = True,
```
hborawski, line 51: i think this should be a parameter for `swiftui_plugin` probably, since currently only one usage of `swiftui_plugin` has UI tests

hborawski (approved):

PR #296, opened 2024-02-07
[Android] `AsyncHydrationTrackerPlugin`

sugarmanz:
### Change Type (required)
Indicate the type of change your pull request is:

- [x] `patch`
- [ ] `minor`
- [ ] `major`

# Release Notes

Introduction of `AsyncHydrationTrackerPlugin` to provide a mechanism for reacting when `SuspendableAsset` hydration is completely finished.

```kotlin
androidPlayer.asyncHydrationTrackerPlugin!!.hooks.onHydrationComplete.tap(this::class.java.name) {
    // process effects after hydration is complete
}
```

PR #396, opened 2024-06-24
Documentation update for Ability to resolve with "nothing"

sakuntala-motukuri:
Documentation update for Ability to resolve with "nothing"

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/docs/site/pages/plugins/async-node.mdx
+++ b/docs/site/pages/plugins/async-node.mdx
@@ -36,7 +36,7 @@ User should tap into the `onAsyncNode` hook to examine the node's metadata befor
 
 ### Edge cases
 
-If for some reason, the process for resolving some `AsyncNode` fails, or is no longer relevant, you can resolve the promise for that `AsyncNode` with `undefined` or `null` to replace that node with "nothing" in the UI.
+If for some reason, the process for resolving some `AsyncNode` fails, or is no longer relevant, you can resolve the promise for that `AsyncNode` with `undefined` or `null` to replace that node with "existing node".
 
 ### Continuous Streaming
 
--- a/docs/site/pages/plugins/async-node.mdx
+++ b/docs/site/pages/plugins/async-node.mdx
@@ -34,11 +34,9 @@ The `AsyncNodePlugin` exposes an `onAsyncNode` hook on all platforms. The `onAsy
 
 User should tap into the `onAsyncNode` hook to examine the node's metadata before making a decision on what to replace the async node with. The return could be a single asset node or an array of asset nodes.
 
-### Improvements
+### Edge cases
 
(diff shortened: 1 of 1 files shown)


docs/site/pages/plugins/async-node.mdx
```diff
@@ -34,6 +34,11 @@ The `AsyncNodePlugin` exposes an `onAsyncNode` hook on all platforms. The `onAsy
 
 User should tap into the `onAsyncNode` hook to examine the node's metadata before making a decision on what to replace the async node with. The return could be a single asset node or an array of asset nodes.
 
+### Improvements
+
+The AsyncNodePluginPlugin class is now capable of handling the below :
+- The resolvedMapping map now sets the node.id as the key and either parsedNode or node as the value, depending on whether parsedNode is truthy.
+- This means that the plugin is now capable of resolving the async node even when the consumer returns a null/undefined value.
```
sugarmanz, line 41: Rather than document the implementation details, can we document what it enables? Maybe something like:

```suggestion
### Edge cases

If for some reason, the process for resolving some `AsyncNode` fails, or is no longer relevant, you can resolve the promise for that `AsyncNode` with `undefined` or `null` to replace that node with "nothing"...
```

sugarmanz: If you want to document the implementation changes, you should put those in the release notes for this PR

sakuntala-motukuri: but we are actually returning the existing node itself instead of nothing

sugarmanz: Example for adding release notes:
https://github.com/player-ui/player/pull/296#issue-2124105491

brocollie08: should we be more intentional about the use case, rather than stating this to be a fallback option of sort

sakuntala-motukuri: is this a question to me or @sugarmanz ?

brocollie08: both, right now it's framed as a fallback option, I'm thinking we make this more intentional because i'm thinking about the future node removal work

sugarmanz: I think that's covered with:
> or is no longer relevant

But we could have an explicit use case around using async node to power toasts, or something like that.

sakuntala-motukuri:
Pushing this to backlog as a collective decision has been made that this will be done after all the related iterations are in , so closing this as of now

PR #404, opened 2024-06-28
Ability to remove a resolved async node

sakuntala-motukuri:
Ability to remove a resolved async node

### Change Type (required)
Attaching logs FYR to see how the fucntionality works : 

**Key Changes**

This version of the plugin allows for the flexibility needed to handle complex async updates, including conditionally removing resolved async nodes from rendering.

Test results:
https://app.buildbuddy.io/invocation/fc720b64-f848-4123-9e95-22216f410acc

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


plugins/async-node/core/src/index.test.ts
```diff
@@ -119,6 +114,86 @@ test("should return current node view when the resolved node is undefined", asyn
...
+        asset: {
+          id: "value-1",
+          type: "text",
+          value: "1st value in the multinode",
+        },
+      },
+      {
+        id: "another-async",
+        async: true,
+      },
```
sugarmanz, line 166: For testing this use case, if you resolve with another async node, the `deferredResolve` value will be updated to be the next async node. We'll want to explicitly test this without triggering another async node resolution to ensure we use the same promise.

plugins/async-node/core/src/index.test.ts
```diff
@@ -119,6 +114,86 @@ test("should return current node view when the resolved node is undefined", asyn
...
+
+  await waitFor(() => {
+    expect(updateNumber).toBe(3);
+  });
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe("action");
+  expect(view?.actions.length).toBe(2);
```
sugarmanz, line 194: Then, down here, I would expect that re-resolving with `null` would mean there would only be a single action value again, like the original assertion before you resolved the first time:
https://github.com/player-ui/player/pull/404/commits/bed39af8c55558871de753c8e8d075afec9f9702#diff-fbc47a3e301bdf595a6d686ded538fe6767e534af8821c421f97dc56264d1332R149-R152

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: Yes the reason I have explicitly resolved with another async node is because the same promise is not getting explicitly resolved . I wanted to show this behaviour to you so that we understand that for the given business logic this test actually fails . If you'd like to see that behaviour I can commit that code but it will not pass . I think we would need to revisit the logic in that case

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: As mentioned in above comment , I have updated this test case to make the build pass . Based on your above review , I will change this in next iteration.

plugins/async-node/core/src/index.test.ts
sugarmanz: Okay, gotcha - yeah, let's get a test case in explicitly for what we're trying to solve for. If it fails, that's fine, we can revisit the implementation logic to get it passing 👍

plugins/async-node/core/src/index.test.ts
```diff
@@ -114,6 +114,80 @@ test("should return current node view when the resolved node is undefined", asyn
...
+  await waitFor(() => {
+    expect(updateNumber).toBe(2);
+  });
+
+  view = (player.getState() as InProgressState).controllers.view.currentView
+    ?.lastUpdate;
+
+  expect(view?.actions[0].asset.type).toBe("action");
+  expect(view?.actions[1].asset.type).toBe("action");
+  expect(view?.actions[2]).toBeUndefined();
```
brocollie08, line 173: is this not checking for something out of bounds?

sakuntala-motukuri:
Updated business logic

plugins/async-node/core/src/index.ts
```diff
@@ -151,8 +153,12 @@ export class AsyncNodePluginPlugin implements AsyncNodeViewPlugin {
         let resolvedNode;
         if (this.isAsync(node)) {
           const mappedValue = this.resolvedMapping.get(node.id);
+          if (this.pendingUpdates.has(node.id)) {
+            return; // Skip processing if already scheduled
+          }
           if (mappedValue) {
             resolvedNode = mappedValue;
+            this.pendingUpdates.add(node.id);
```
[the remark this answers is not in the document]
sakuntala-motukuri: **Intent of pendingUpdates**

Prevent Duplicate Processing: When an asynchronous node is being processed (e.g., fetching data or resolving content), it's marked in pendingUpdates. This prevents the same node from being processed multiple times if it's encountered again before the current processing completes. This is particularly useful in scenarios where nodes might be rapidly updated or re-rendered.
Optimize Performance: By avoiding duplicate processing of nodes, the application can save on unnecessary computations or network requests, leading to better performance.
Synchronization: It helps in synchronizing the state of the node between the time it is first processed and when it is updated. This ensures that updates to nodes are handled in an orderly manner, respecting the sequence of updates.

**Separation from resolvedMapping**

Different Purposes: resolvedMapping is used to store the final resolved state of an asynchronous node. In contrast, pendingUpdates is used to track nodes that are currently being processed or have updates pending. Once the processing of a node is complete, it's moved from pendingUpdates to resolvedMapping.
Lifecycle Management: pendingUpdates is more about managing the lifecycle and processing state of nodes, while resolvedMapping is about storing the resolved data. This separation allows for more granular control over node processing and rendering.
Concurrency Control: The separation allows the system to handle concurrent updates to nodes more effectively. While resolvedMapping holds the current resolved state, pendingUpdates ensures that concurrent processes do not interfere with each other.

In summary, pendingUpdates is a mechanism to track and manage the processing state of asynchronous nodes to prevent duplicate processing and optimize performance, while resolvedMapping stores the resolved state of these nodes for rendering. This separation is crucial for efficient and accurate state management of asynchronous operations within the application.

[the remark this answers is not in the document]
brocollie08: I get where you're coming from, if we want to avoid all unnecessary processing, why is this added to the set only after there is already a mapped value? The existence of a mapped value implies the node has already been processed and the final result is ready to be used, so at this point, [this](https://github.com/player-ui/player/pull/404/files#diff-9b846b61d9cdd8ebb4e14c47525e1fbc6d69e6088fd7d938d62577a72feb3327R156) is no longer a useful check.

I think you want to track this, you'd want to track it right away, instead of at the end

plugins/async-node/core/src/index.ts
```diff
@@ -127,7 +149,12 @@ export class AsyncNodePluginPlugin implements AsyncNodeViewPlugin {
...
       if (!resolvedNode && node?.type === NodeType.Async) {
         queueMicrotask(async () => {
-          const result = await this.basePlugin?.hooks.onAsyncNode.call(node);
+          const result = await this.basePlugin?.hooks.onAsyncNode.call(
+            node,
+            (result) => {
+              this.updateAsyncFunc(node, result, options, this.currentView);
+            },
+          );
           const parsedNode =
```
sugarmanz, line 158: Can we also update this part to use the `updateAsyncFunc` since it performs the same logic?

plugins/async-node/core/src/index.ts
```diff
@@ -70,6 +77,21 @@ export class AsyncNodePluginPlugin implements AsyncNodeViewPlugin {
 
   private currentView: ViewInstance | undefined;
 
+  private updateAsyncFunc(
```
sugarmanz, line 80: Nit: Naming for this - I wouldn't usually expect a member function to have `Func` in the name since it's redundant. `handleAsyncUpdate` might fit better, but open to other ideas!

plugins/async-node/core/src/index.test.ts
```diff
@@ -114,6 +127,90 @@ test("should return current node view when the resolved node is undefined", asyn
   await asyncNodeTest(undefined);
 });
 
+test("should handle the promise using .then", async () => {
```
sugarmanz, line 130: Can you update the test name to something a bit more accurate to the new approach? Maybe `can handle multiple updates through callback mechanism`?

plugins/async-node/core/src/index.test.ts
```diff
@@ -424,30 +521,3 @@ test("replaces async nodes with chained multiNodes singular", async () => {
   expect(view?.actions[1].asset.type).toBe("text");
   expect(view?.actions[2].asset.type).toBe("text");
 });
-
-test("should call onAsyncNode hook when async node is encountered", async () => {
```
sugarmanz, line 428: I don't think this test should be deleted, right?

plugins/async-node/core/src/index.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
```diff
@@ -114,6 +127,140 @@ test("should return current node view when the resolved node is undefined", asyn
...
+    },
+  };
+
+  player.start(basicFRFWithAsyncNode as any);
+
+  // Wait for any asynchronous operations to potentially complete
+  await new Promise((resolve) => setTimeout(resolve, 100));
+
+  // Verify that the updateNumber is still 0, indicating that the async node was not resolved
+  expect(updateNumber).toBe(0);
```
sugarmanz, line 261: Hmm, I'm a bit skeptical with this test -- it doesn't look like you have the setup to update `updateNumber` on view updates. And even then, I would expect it to update at least once for when the initial view resolution occurs when you start the Player

plugins/async-node/core/src/index.test.ts
```diff
@@ -346,7 +493,7 @@ test("replaces async nodes with chained multiNodes singular", async () => {
 
   let deferredResolve: ((value: any) => void) | undefined;
```
sugarmanz, line 495: I don't think this change should be made (probably a miss on some rebase)

plugins/async-node/core/src/index.test.ts
sakuntala-motukuri: yes I removed it on purpose , as I wanted to check if the view would still be resolved even if the baseplugin instance doesn't exist , do we still want the view to update ?

plugins/async-node/core/src/index.test.ts
```diff
@@ -114,6 +127,146 @@ test("should return current node view when the resolved node is undefined", asyn
...
+          transitions: {},
+        },
+      },
+    },
+  };
+
+  player.start(basicFRFWithAsyncNode as any);
+
+  // Wait for any asynchronous operations to potentially complete
+  await new Promise((resolve) => setTimeout(resolve, 100));
```
sugarmanz, line 264: Instead of this, can you just wrap the assertions in a `waitFor`? This delay is a little nebulous, and depends on test timer configuration

sakuntala-motukuri: do you mean something like this await waitFor(() => {
  expect(updateNumber).toBe(1);
  expect(deferredResolve).toBeDefined();
});

sakuntala-motukuri: done

plugins/async-node/core/src/index.test.ts
brocollie08: since `view` is just a record, and the value of `actions` can be of type `any`, could you get `actions[2]` being undefined but `actions[3]` being well defined? If so, does it make more sense to check size rather than `actions[2]` specifically

brocollie08 (commented):
We have duplicate code in `applyResolverHooks`, we can probably clean that up to be called in the `apply` method. We should probably follow the convention of `ApplicabilityPlugin` and rename it to `applyResolver`

docs/site/pages/plugins/async-node.mdx
```diff
@@ -72,7 +86,7 @@ const player = new Player({
   </core>
   <react>
 
-The `react` version of the AsyncNodePlugin is identical to using the core plugin. Refer to core usage for handler configuration:
+The `react` version of the AsyncNodePluginPlugin is identical to using the core plugin. Refer to core usage for handler configuration:
```
sugarmanz, line 89: I think this should stay `AsyncNodePlugin`

docs/site/pages/plugins/async-node.mdx
```diff
@@ -42,7 +43,8 @@ This means there must be a constant renewal of new `AsyncNode`s after the previo
 
 ## Usage
 
-The AsyncNodePlugin itself accepts a list of plugins and should take the AsyncNodePluginPlugin to be passed in. The AsyncNodePluginPlugin also comes from the `'@player-ui/async-node-plugin'` and contains the resolver and parser functionality
+The `AsyncNodePlugin` itself accepts an options object with a `plugins` array, enabling the integration of multiple view plugins for extended functionality.
+The `AsyncNodePluginPlugin` is provided as a default way of handling asset-async nodes, it is just one handler for one possible way of using async nodes. If the default behaviour does not align with the desired usage, users are able to provide their own implementation of the handler in the form of a plugin to be passed to the base `AsyncNodePlugin`.The `AsyncNodePluginPlugin` also comes from the `'@player-ui/async-node-plugin'` and contains the resolver and parser functionality
```
sugarmanz, line 47:
```suggestion
The `AsyncNodePluginPlugin` is provided as a default way of handling asset-async nodes, it is just one handler for one possible way of using async nodes. If the default behavior does not align with the desired usage, users are able to provide their own implementation of the handler in the form of a plugin to be passed to the base `AsyncNodePlugin`. The `AsyncNodePluginPlugin` also comes from the `'@player-ui/async-node-plugin'` and contains the resolver and parser functionality.
```

sakuntala-motukuri: done

docs/site/pages/plugins/async-node.mdx
sakuntala-motukuri: done

docs/site/pages/plugins/async-node.mdx
```diff
@@ -32,8 +32,9 @@ In the example below, node with the id "some-async-node" will not be rendered on
...
 
-User should tap into the `onAsyncNode` hook to examine the node's metadata before making a decision on what to replace the async node with. The return could be a single asset node or an array of asset nodes.
+User should tap into the `onAsyncNode` hook to examine the node's metadata before making a decision on what to replace the async node with. The return could be a single asset node or an array of asset nodes, or the return could be even a null|undefined if the async node is no longer relevant.
 
+Returning a value in the above context enables uses cases where the async node only needs to be resolved once. For use cases where the async node needs to be updated multiple times, the onAsyncNode hook provides a second callback argument that can be used to update the value multiple times. For example, if the async node is used to represent some placeholder for toasts, or notifications, the async node handler could initially resolve with some content, and then update with null after some time to remove those views.
```
brocollie08, line 37: 👍

plugins/async-node/core/src/index.test.ts
brocollie08: I mean I think we shouldn't check for `view?.actions[2]` at all  because there's no significance to it

sakuntala-motukuri: Anyhow view?.actions[2] would be undefined , but yes as you said it makes no sense checking for this at this point of time as view?.actions[2] is not expected at all. Removed it

plugins/async-node/core/src/index.ts
```diff
@@ -113,69 +181,13 @@ export class AsyncNodePluginPlugin implements AsyncNodeViewPlugin {
...
-            this.resolvedMapping.set(node.id, parsedNode ? parsedNode : node);
-            view.updateAsync();
-          });
-
-          return node;
-        }
-
-        return newNode;
-      });
+      this.handleAsyncApi(resolver, view);
```
brocollie08, line 190:
```suggestion
     this.applyResolver.bind(this)
```
to keep convention consistent with other `ViewPlugin`s

brocollie08 (approved):

PR #395, opened 2024-06-24
[Hermes] JSI JNI

sugarmanz:
Initial groundwork for using the Hermes runtime on the JVM. There are a few parts to this PR:
1. Configuring Hermes & FBJNI as part of our build
2. Wrap JSI (JavaScript Interface) constructs through JNI holders
3. Create Hermes runtime through JNI constructs

What we're aiming for (some slight inaccuracies):
https://gist.github.com/sugarmanz/d07d443f24f6eee1db9ac1c7ba1681c1

Some helpful links:
- [Hermes Repo](https://github.com/facebook/hermes)
  - [JSI Definition](https://github.com/facebook/hermes/blob/main/API/jsi/jsi/jsi.h)
  - [HermesRuntime Definition](https://github.com/facebook/hermes/blob/main/API/hermes/hermes.h)
- [FBJNI Repo](https://github.com/facebookincubator/fbjni)

There are several docs around intro to JSI, but they mainly are in the context of React Native. Our paradigm is different, as we aim to extend the JSI to the JVM to enable "direct" (at least through JNI) integration with the native JavaScript Runtime implementations. This way, we can connect it to our existing JS abstraction (`Node` & `Runtime`) for use within Player without having to change any of the Player code. This is a realization of the goals for that abstraction piece, and really, I see that abstraction piece as a parallel to the native JSI definition. We'll likely want to converge the two abstractions and maybe even position this as bringing JSI to the JVM.

For diving in, start with the JVM JSI definitions and see how they're used in the tests (relies on `HermesRuntime`):
- [Value.kt](https://github.com/player-ui/player/blob/hermes/jsi-jni/jvm/hermes/src/main/kotlin/com/intuit/playerui/jsi/Value.kt): JVM FBJNI wrappers for JSI constructs
  - [JJSIValue.h](https://github.com/player-ui/player/blob/hermes/jsi-jni/jvm/hermes/src/main/jni/JJSIValue.h): C++ FBJNI wrappers for JSI constructs that map directly to the `Value.kt` definitions.
- [JSITests.kt](https://github.com/player-ui/player/blob/hermes/jsi-jni/jvm/hermes/src/test/kotlin/com/intuit/playerui/jsi/JSITests.kt)
- [HermesRuntime.kt](https://github.com/player-ui/player/blob/hermes/jsi-jni/jvm/hermes/src/main/kotlin/com/intuit/playerui/hermes/bridge/runtime/HermesRuntime.kt): JVM FBJNI wrappers for the Hermes runtime
[...]

PR #399, opened 2024-06-25
[Hermes] Node, Runtime, Format Implementations

sugarmanz:
### Change Type (required)
Indicate the type of change your pull request is:

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

### Does your PR have any documentation updates?
- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

PR #410, opened 2024-07-02
[Hermes] Android integration

sugarmanz:
## Rationale

By default, the Android Player consumes `com.intuit.playerui:j2v8-android` to provide the JS runtime integration since it is reasonably fast and small compared to some of the alternatives (6-7 MB per architecture). However, as the goal with the runtime abstraction was to enable any JS runtime to be swapped out at will, we want to officially support runtimes that bring value to the platform. The [Hermes](https://github.com/facebook/hermes) JS runtime created for React Native showed promising size improvements along with a potential runtime performance optimization for pre-compiled JS. Since we also leverage Bazel as our build tool, this empowers us to more easily bring in complex build dependencies as a patternized part of our build. Integrating with this runtime has been on our backlog for several years, and it's great to finally be able to properly consider it.

## Results

Shown below is a screenshot comparing the APK size of the `//android/demo` application, with some notes beneath:

image

- The overall size of the app doesn’t represent just Player, as it includes all the components required for our demo app as well.
- The breakdown includes raw size vs download size, as well as multiple CPU architectures. We really only care about the download size for a single architecture (as long a the app is published as an AAB — if it’s not, it’ll include all CPU architectures, but that would be a further optimization to be done on the app side).
- The largest part of Player is the native engine that runs the core Player code. The existing engine, J2V8 being 6-7 MB per architecture. Hermes, plus the native components we need to utilize it, is reported as 1.4-1.8 MB! With arm64 being the most popular architectures, this is a drop from 7 -> 1.6 MB. Again, these numbers might fluctuate with changes to the Hermes integration, but I don’t expect them to change significantly.
- The other parts of Player is the actual code, both core and JVM classes. Not accounting for plugins (as that varies by the needs of the integration), they come out to about 675 KB, resulting in a total size for Android Player of about 2.3 MB.

This PR doesn't serve to publish Hermes compatible runtime integrations for non-Android environments, however, it does build Hermes for the host platform to run our JVM `HeadlessPlayer` tests against. J2V8 seemed to have a leg up with regard to desktop performance (without bytecode optimization), but fortunately, Hermes built for Android is different in a few ways and ends up proving to be relatively flat runtime performance in comparison, if not slightly better. More performance testing to come with bytecode optimization.

Other improvements include:
- Better ES6 compatibility
- Better cross-JNI/JSI exception handling (thanks [FBJNI](https://github.com/facebookincubator/fbjni) and source map support)

## Next Steps

The initial integration proved to provide a substantial size decrease, while offering a relatively flat, if not slightly better, runtime performance (on Android). This is a great start, but there are more improvements we can make in future iterations, in order of complexity:

- Dev docs
  - Potentially help contribute towards Hermes dev docs (as somewhat requested in https://github.com/facebook/hermes/issues/958)
- Assume as default runtime for Android Player
- Hermes Bytecode compilation
- Publish Hermes integration for desktop JVM environments
- Consolidate JSI definition with Player runtime abstraction
  - Enables much more common parts of the runtime serialization layer 🎉 along with solving a very longstanding [TODO](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/Promise.kt#L49-L64) 
- [Static Hermes](https://github.com/facebook/hermes/discussions/1137)

## How to review this PR 

This is a relatively large PR that includes multiple languages and complex build setup. There will be more comprehensive development docs to come, but here are the key parts to look at:

### Bazel setup
  - [Hermes external dependency](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/WORKSPACE.bzlmod#L31-L37)
    - [For `host`](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/third_party/hermes/BUILD#L9-L39)
      - [Hello world test harness](https://github.com/player-ui/player/tree/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/third_party/hermes/hello)
    - [For Android](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/WORKSPACE.bzlmod#L49-L54)
  - [FBJNI external dependency](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/WORKSPACE.bzlmod#L39-L45)
    - [For `host`](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/third_party/fbjni/BUILD#L35-L58)
    - [For Android](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/third_party/fbjni/artifact.BUILD)
  - [React Native external dependency](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/WORKSPACE.bzlmod#L56-L63) for RN JSI override
  - [`libjsi.so` dependency](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/libs/BUILD#L6-L17)
    - The only native binary checked in b/c it was difficult to build or find prebuilt dynamically
  - [Hermes integration conditional JNI](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/src/main/jni/BUILD#L37-L50) (JVM vs Android)
    - [Packaging Hermes for Android](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/BUILD#L35-L48)
    - [Building `hermes-host`](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/src/main/jni/BUILD#L54-L93) for testing

Our Bazel setup uses different strategies for integrating Hermes for Android and host machines. For host, it's simple enough to compile Hermes directly and consume for running headless tests. For Android, we have initially used a prebuilt Hermes from React Native - not ideal, but helps us get integrated a bit quicker. We'll want to plan on configuring proper x-compilation for Android so we can have finer grain control of the Hermes build configuration and optimize for our use case. You can see which artifacts are used for which platforms below Android [here](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/BUILD#L39-L43) and host [here](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/src/main/jni/BUILD#L60-L93).

### JSI JNI (JavaScript Interface Java Native Interface)

Hermes is built to be a JSI compliant runtime, meaning we can build out our integration with respect to the JSI definition.  Some more on this in #395 PR description. More that's not covered in that description includes:

- [`SoLoader`](https://github.com/facebook/SoLoader) dependency for [loading native libraries](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/src/main/kotlin/com/intuit/playerui/hermes/bridge/runtime/HermesRuntime.kt#L46-L55)
  - Required for FBJNI, so we lean into this for loading all native libs
  - [`ResourceLoaderDelegate`](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/src/main/kotlin/com/intuit/playerui/bridge/loader/ResourceLoaderDelegate.kt) for host and fallback mechanism

### JSI Serialization

The Player runtime abstraction is built on the [`kotlinx.serialization`](https://github.com/Kotlin/kotlinx.serialization), and thus we need an encoder and decoder to support the new JSI types from above. Mostly similar to the other runtimes, leading towards a greater idea of how we can lean into JSI and consolidate the implementations a bit more. Some additional things to note below:

- [Serialization constructs](https://github.com/player-ui/player/tree/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/src/main/kotlin/com/intuit/playerui/jsi/serialization)
- [`RuntimeThreadContext`](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/src/main/kotlin/com/intuit/playerui/hermes/extensions/Lock.kt#L12-L34) to explicitly require the correct thread context for accessing thread-unsafe APIs
  - Uses [context receivers](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/hermes/src/main/kotlin/com/intuit/playerui/jsi/Value.kt#L37) enabled with [`rules_kotlin` patch](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/patches/rules_kotlin.context_receivers.patch) 

### `//jvm/core` Code Changes

Most of the changes in here are small tweaks to fix edge cases I encountered while testing, but there might be some new enhancements to better serve runtime integrations, like [`Runtime.executeRaw`](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/runtime/Runtime.kt#L26-L28) and [`ScriptContext.sourceMap`](https://github.com/player-ui/player/blob/becf2144b0dfbba0ab7f67b92633fe3aa7dea819/jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/runtime/Runtime.kt#L67).

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

# Release Notes

Initial integration with the [Hermes](https://github.com/facebook/hermes) JavaScript runtime. This shows a tremendous size improvement over the existing [J2V8](https://github.com/eclipsesource/J2V8) integration of ~70% (7.6 MB -> 2.3 MB, architecture dependent). 

### Opt-in

For now, the default runtime integration provided by the Android Player will still be `com.intuit.playerui:j2v8-android`, but Hermes can be opted in manually by excluding the J2V8 transitive dependency and including the Hermes artifact:

```kotlin
dependencies {
    // Android Player dependency
    implementation("com.intuit.playerui", "android", PLAYER_VERSION) {
        // J2V8 included for release versions
        exclude(group = "com.intuit.playerui", module = "j2v8-android")
        // Debuggable J2V8 included for canary versions
        exclude(group = "com.intuit.playerui", module = "j2v8-android-debug")
    }
    // Override with Hermes runtime
    implementation("com.intuit.playerui", "hermes-android", PLAYER_VERSION)
}
```

> [!TIP]
> If your application includes dependencies that may transitively depend on `com.intuit.playerui:android`, you would likely need to ensure the default runtime is transitively excluded from those as well, either manually or as a global strategy.
> 
> The `AndroidPlayer` will pick the first runtime it finds on the classpath - you can at least verify which runtime was used for the `Player` with a new log: `Player created using $runtime`. But that won't tell you for certain if the other runtimes were successfully excluded. You'll need to examine your APK, or your apps dependency tree, to tell for sure that redundant runtimes aren't unintentionally included.

Most of the setup for this integration is done simply by including the right dependency (and excluding the wrong one), however, the `hermes-android` integration also relies on the [SoLoader](https://github.com/facebook/SoLoader) for loading the native libraries. All that's needed is to initialize the `SoLoader` (should be on your classpath with the `hermes-android` dependency) with an Android `Context` somewhere before you use the `AndroidPlayer`, potentially in your activities `onCreate`:

```kotlin
override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

    SoLoader.init(this, false)
    // ...
}
```

> [!TIP]
> 

<details>
  <summary>📦 Published PR as canary version: <code>0.8.0--canary.410.16288</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.8.0--canary.410.16288
</details>


sugarmanz:
Contains the work from #395 & #399.

.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ orbs:
 executors:
   base:
     docker:
-      - image: docker.io/playerui/bazel-docker
+      - image: docker.io/playerui/bazel-docker:11
```
KetanReddy, line 19: Should we merge the docker changes in before merging this so we're not targeting a specific image?

sugarmanz: We can! I was avoiding merging that PR until the release was done to make sure it wouldn't accidentally break CI (does make a strong case for tagging our images per release and always targeting a tag).

.circleci/config.yml
```diff
@@ -237,6 +237,7 @@ jobs:
 
       - run: |
           bazel test --config=ci -- $(bazel query 'kind(".*_test", //...) except filter("ios|swiftui", //...)') -//android/demo:android_instrumentation_test
+          bazel test --config=ci --//android/player:runtime=hermes -- $(bazel query 'kind(".*_test", //...) except filter("ios|swiftui", //...)') -//android/demo:android_instrumentation_test
```
sugarmanz, line 240: This, plus the one below ensure we re-run all relevant tests (that depend on `//android/player:runtime`) against the Hermes runtime. It'll return the cached result for all tests that don't use that config flag, directly or transitively.

brocollie08 (approved):

KetanReddy (approved):

KetanReddy:
@sugarmanz with this merged can we close #410 and #399?

PR #413, opened 2024-07-05
View Parser Plugin Decomposition

### LINKED ISSUE #373: [Core] View Parser Plugin Decomposition
**Is your feature request related to a problem? Please describe.**

The [view parser](https://github.com/player-ui/player/blob/bazel-6/core/player/src/view/parser/index.ts) currently sits in a middleground between plugin architecture, and coupled functionality. 

**Describe the solution you'd like**

Refactor the view parser so that [parseLocalObject](https://github.com/player-ui/player/blob/bazel-6/core/player/src/view/parser/index.ts#L146) does not contain code to parse specific nodes, but rather delegates that functionality to plugins. Some functionality is already in [plugins](https://github.com/player-ui/player/tree/bazel-6/core/player/src/view/plugins).

**Additional context**

It should be possible to add a new node type through a plugin, in case there was some need for a more specific type of applicability or switch, etc, for a single integration.

### LINKED ISSUE #412: Error thrown from applicability when a dataset has null item
**Describe the bug**

When a null item is provided in an Array of a dataset for an asset it produces an error that comes from the applicability check:

TypeError: Cannot convert undefined or null to object
    at hasOwnProperty (<anonymous>)
    at hasApplicability (vendors~main.25b9d636.iframe.bundle.js:2:4575177) 

**To Reproduce**
Simply by testing any content with an array value, and adding a `null` item in the array.

**Expected behavior**

**Screenshots**

**Environment:**

**Additional context**

mrigankmg:
This PR refactors the view parser so that [parseLocalObject](https://github.com/player-ui/player/blob/bazel-6/core/player/src/view/parser/index.ts#L146) does not contain code to parse specific nodes, but rather delegates that functionality to plugins.

resolves #373 
resolves #412 

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


core/player/src/view/plugins/switch.ts
```diff
@@ -33,75 +47,92 @@ export default class SwitchPlugin implements ViewPlugin {
...
               dynamic,
               cases,
             },
-            obj,
+            objToParse,
           );
 
-          if (switchAST?.type === NodeType.Switch) {
+          if (!switchAST || switchAST.type === NodeType.Empty) {
+            return childOptions ? [] : null;
```
brocollie08, line 109: distinction between this `null` and [this](https://github.com/player-ui/player/pull/413/files#diff-f177d6cbe60df2262c39019c6c0a9323e88dd6ddfe776691c54cefa008bee645R70) `undefined`?

plugins/async-node/core/src/index.test.ts
```diff
@@ -547,3 +547,43 @@ test("should call onAsyncNode hook when async node is encountered", async () =>
     expect(localNode.type).toStrictEqual("async");
   });
 });
+
+test("missing node-id", async () => {
```
brocollie08, line 551: not sure i'm following this test, the node in here doesn't actually get parsed to be an async node, does it?

core/player/src/view/parser/utils.ts
```diff
@@ -1,8 +1,28 @@
...
+  return (
+    Object.hasOwnProperty.call(obj, "template") &&
+    Array.isArray(obj?.template) &&
+    obj.template.length &&
+    obj.template.find((tmpl: any) => tmpl.output === localKey)
+  );
+}
+
+/** Check to see if the string is a valid switch key */
+export function hasSwitchKey(localKey: string) {
```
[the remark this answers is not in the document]
mrigankmg: This is not currently used outside of `switch.ts`. The only reason I have it here in `utils.ts` is to keep it consistent with `hasTemplateKey`. I could move it to `switch.ts` though if you don't think there's value in keeping it here.

plugins/async-node/core/src/index.test.ts
mrigankmg: Here I'm testing for an invalid async node (one that has the `async` property but not the `id` property) that is a child of another property within the view. Here I'm specifically testing that the async node plugin returns a `null` AST by hitting L168 in `plugins/async-node/core/src/index.ts`.

core/player/src/view/__tests__/view.immutable.test.ts
```diff
@@ -34,6 +39,7 @@ test("uses the exact same object if nothing changes", () => {
   );
 
   new StringResolverPlugin().apply(view);
+  new AssetPlugin().apply(view);
```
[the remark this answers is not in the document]
mrigankmg: Do we need to create a separate test util plugin? Could we not use the `DefaultViewPlugin ` itself?

core/player/src/view/plugins/__tests__/applicability.test.ts
```diff
@@ -245,21 +248,26 @@ describe("applicability", () => {
     });
   });
 
-  it("determines if nodeType is applicability", () => {
```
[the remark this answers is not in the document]
mrigankmg: How would we achieve testing of a private function?

core/player/src/view/parser/utils.ts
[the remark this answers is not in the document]
brocollie08: I don't have strong opinions on this, I pointed it out because I thought this was similar to the functions in each plugin that checks the type (replacing `determineNodeType`) as they're only relevant to said plugin. You can decide what to do here

core/player/src/view/__tests__/view.immutable.test.ts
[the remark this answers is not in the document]
brocollie08: ya that's possibly fine, it's probably registering more plugins than you need for a lot of tests that are very targeted.
Maybe @KetanReddy can chime in here

plugins/async-node/core/src/index.test.ts
brocollie08: I see. The thing I have to point out here is that even if it's a valid async node, the _view_ would still be what you are checking because unresolved async nodes don't actually add anything to the view tree.

I'd check to see if the hook ever gets called for processing async nodes, or something along those lines

mrigankmg: Simply returning `null` on L168 in `plugins/async-node/core/src/index.ts` instead of doing the `childOptions` check, results in a view tree that looks like this `{ id: 'my-view', fields: { async: 'true' } }`. So there is a faulty async node still added to the tree, unless I'm missing something here.

core/player/src/view/plugins/__tests__/applicability.test.ts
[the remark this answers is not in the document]
brocollie08: you'd call the hook that calls the private function, then verify the side effects of the private function

plugins/async-node/core/src/index.test.ts
brocollie08: try adding an id to that nested object, I think you'll likely still get the test passing which would mean id or no id, the view is the same

mrigankmg: Ahhh ok! I see what you mean.

core/player/src/view/plugins/__tests__/applicability.test.ts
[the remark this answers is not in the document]
brocollie08: as discussed in slack, these tests might not be relevant anymore since the result being returned -> isApplicability returned true, as long as we have a test for negative case as well, this is fine

core/player/src/view/parser/utils.ts
```diff
@@ -1,8 +1,28 @@
...
+  return localKey === ("staticSwitch" || "dynamicSwitch");
```
KetanReddy, line 20: Huh, I didn't know this syntax was possible to check if a string was either value.

KetanReddy: i would have assumed `("staticSwitch" || "dynamicSwitch")` would have evaluated to `true` so it would have just checked if `localKey` was `true`

mrigankmg: So seems like this syntax does not work. This was pre-existing code that I moved over to `utils.ts` from parser `index.ts`. I'll leave the fix for this out of this PR and defer it to another ticket for now!

KetanReddy: Since we're making changes in the area I would opt to just make the changes now. Its also concerning that we don't have a test that would have caught this?

hborawski: I made #453 to track this, to make sure we're capturing additional work added to this long running task

hborawski (approved):
🚀 This is a huge refactor, glad to see it fitting properly into the plugin architecture!

brocollie08 (approved):

PR #417, opened 2024-07-09
bazel6-conflicts

mercillo:
### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs


core/player/src/__tests__/validation.test.ts
```diff
@@ -1038,7 +1038,7 @@ describe("validation", () => {
       // Try to transition again
       state.controllers.flow.transition("foo");
 
-      // Should work now that there's no error
+      // Should work now that there"s no error
```
sugarmanz, line 1041: comments in this file probably shouldn't need to replace `'` with `"`

plugins/async-node/core/src/index.ts
```diff
@@ -44,7 +44,7 @@ export class AsyncNodePlugin implements PlayerPlugin {
   }
 
   public readonly hooks = {
-    onAsyncNode: new AsyncParallelBailHook<[Node.Async], any>(),
```
sugarmanz: Hmm - the incoming change from `main` should be accepted here, as a part of:
#371

PR #419, opened 2024-07-10
Bazel 6 android add dsl component to async node

sakuntala-motukuri:
Bazel 6 android add dsl component to async node

Test results :
 https://app.buildbuddy.io/invocation/37db2f3b-067f-4144-b33f-a8abd66360fe

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


plugins/async-node/components/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@player-ui/async-node-plugin-components",
+  "version": "0.0.0-PLACEHOLDER",
+  "main": "src/index.tsx",
```
sugarmanz, line 4: I think the build is failing b/c this doesn't actually point to anything

KetanReddy: Yeah you shouldn't need to manually define this, it gets automatically added when the `package.json` is put together in the Bazel rules

plugins/async-node/components/src/dsl/async.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import type { ObjectNode } from "react-json-reconciler";
+
+export type AsyncNodeProps = {
+  id?: string;
```
KetanReddy, line 5: We'd probably want to have ID be mandatory if we aren't going to have a way to generate them. I would also push for them to have to be defined since content authors need to know the node id that they have in content to resolve in their plugin usage.

plugins/async-node/components/src/dsl/async.tsx
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export const AsyncNode = React.forwardRef<ObjectNode, AsyncNodeProps>(
+  (props) => {
+    const { id } = props;
+    const Wrapper = React.Fragment;
+
+    return (
+      <Wrapper>
```
KetanReddy, line 14: I would just use `React.Fragment` directly here instead of defining it in a `const`

KetanReddy:
One thing that we should do as part of this is add XLR exports to the core async node plugin (like we do for the [reference assets](https://github.com/player-ui/player/blob/bazel-6/plugins/reference-assets/core/src/plugin.ts#L23)) to allow it to be loaded into the validator via a cli preset during content compilation.

plugins/async-node/components/package.json
KetanReddy: There would need to be a root `index.tsx` file that exports the `async.tsx` file.

plugins/async-node/components/src/dsl/async.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import type { ObjectNode } from "react-json-reconciler";
+
+export type AsyncNodeProps = {
+  id?: string;
+};
+
+export const AsyncNode = React.forwardRef<ObjectNode, AsyncNodeProps>(
```
KetanReddy, line 8: Does this need to be a `forwardRef`? For our other DSL components they are just plain React Components.

plugins/async-node/components/package.json
sakuntala-motukuri: well I have taken reference-assets as plugin and it doesn't have a separate index file as mentioned its just that instead of index.tsx , I have named it as async.tsx 
image
Please can you suggest , ty

plugins/async-node/components/src/dsl/async.tsx
sakuntala-motukuri: Well , I have taken the reference of Asset component as below :
image

But I agree to the fact that this could be acheived using simple react components but I see the specific usage of. ref comp as  React.forwardRef is a technique used within functional components to forward refs to child components or DOM elements, enabling direct access to the DOM from parent components

plugins/async-node/components/src/dsl/async.tsx
sakuntala-motukuri: done

plugins/async-node/components/src/dsl/async.tsx
sakuntala-motukuri: good catch ! done

sakuntala-motukuri:
One thing that we should do as part of this is add XLR exports to the core async node plugin (like we do for the [reference assets](https://github.com/player-ui/player/blob/bazel-6/plugins/reference-assets/core/src/plugin.ts#L23)) to allow it to be loaded into the validator via a cli preset during content compilation. 

Reply: Not doing above as discussed with @KetanReddy , as its difficult to decide into what type the async node resolves to as this dsl configuration doesn't actually use the asset component but rather stands as an independent async component.

KetanReddy (approved):

Issue #376, opened 2024-06-05
React Asset: Improved Error messaging


hborawski:
**Is your feature request related to a problem? Please describe.**

The [ReactAsset](https://github.com/player-ui/player/blob/main/react/player/src/asset/index.tsx) component handles fetching the implementation from the asset registry, and throwing errors if an asset is not found. These errors indicate the direct cause, but could do a better job guiding the user to the root cause.

We should enhance this messaging to help the user self resolve the issue when possible.

**Describe the solution you'd like**

- Add additional metadata to the existing error messages to include what assets and registrations the current configuration supports.
- Add some [react-like](https://iws.io/2022/invalid-hook-multiple-react-instances) errors that give suggestions when there are multiple potential causes
  - Mismatched React Context could show no assets registered, but could be caused by no plugins being added to the Player instance.

**Additional context**

PR #421, opened 2024-07-11
ReactAsset improve error messages

cehan-Chloe:
Issue: https://github.com/player-ui/player/issues/376 

Test result: https://app.buildbuddy.io/invocation/8eb077c0-1586-4699-b732-c92c332e7697

What's changed: improve error msg thrown by ReactAsset. more info and react-like errors
Before: if no matched asset found, throw error: `No implementation found for id: bar-id type: bar`
After: 
1. add check for no assets in registry
image

2. modify existing error msg for no match asset found
image

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs


react/player/src/asset/index.tsx
```diff
@@ -49,11 +49,30 @@ export const ReactAsset = (
...
+      See https://player-ui.github.io/latest/tools/cli#player-dependency-versions-check for tips about how to debug and fix this problem`);
+  }
+
   const Impl = registry?.get(unwrapped);
 
+  const matchList: object[] = [];
+
+  registry.forEach((asset) => {
+    matchList.push(asset.key);
+  });
```
KetanReddy, line 65: Can we compute this only in the error case so we don't do this work if there is no error?

core/partial-match-registry/src/index.ts
```diff
@@ -75,4 +75,9 @@ export class Registry<V> {
   clear() {
     this.store = createSortedArray<V>();
   }
+
+  /** Check if the registry is empty*/
+  isRegistryEmpty() {
```
KetanReddy, line 80: @sugarmanz @hborawski Do we expose the partial match registry on every platform(It doesn't look like we do but I just wanted to check)? If so we should also add this to every platforms API.

react/player/src/asset/index.tsx
```diff
@@ -49,11 +49,30 @@ export const ReactAsset = (
...
+
   if (!Impl) {
     throw Error(
-      `No implementation found for id: ${unwrapped.id} type: ${unwrapped.type}`,
+      `No implementation found for id: ${unwrapped.id} type: ${
+        unwrapped.type
+      }. This could happen for one of the following reasons: \n
+      1. You might not have the asset id: ${unwrapped.id} type: ${unwrapped.type} registered. \n
+      2. You might have multiple assets libraries in the same app. \n
+      See match list below for more information: \n
```
KetanReddy, line 74:
```suggestion
      Registered Asset matching functions are listed below: \n
```

react/player/src/asset/index.tsx
```diff
@@ -49,11 +49,30 @@ export const ReactAsset = (
...
+  registry.forEach((asset) => {
+    matchList.push(asset.key);
+  });
+
   if (!Impl) {
     throw Error(
-      `No implementation found for id: ${unwrapped.id} type: ${unwrapped.type}`,
+      `No implementation found for id: ${unwrapped.id} type: ${
+        unwrapped.type
+      }. This could happen for one of the following reasons: \n
```
adierkens, line 71: Could we check for similarly named types and suggest alternatives? 

```
No implementation found for type: "txt" did you mean "text"?
```

core/partial-match-registry/src/index.ts
hborawski: https://github.com/player-ui/player/blob/bazel-6/ios/core/Sources/CorePlugins/PartialMatchFingerprintPlugin.swift#L75

We do have it in iOS, but it's not exposed to users + no Swift API currently

PR #443, opened 2024-07-22
Fix J2V8 Debugger

brocollie08:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.7.42--canary.443.15975</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.7.42--canary.443.15975
</details>


.circleci/config.yml
```diff
@@ -101,7 +111,7 @@ jobs:
       - attach_workspace:
           at: ~/player
       # need to filter because //android/demo rules dont resolve
-      - run: bazel build --config=ci -- $(bazel query 'filter("native_bundle$", //...)')
+      - run: bazel build --config=ci --define build_type=$BUILD_TYPE -- $(bazel query 'filter("native_bundle$", //...)')
```
sugarmanz, line 114: I don't think you need the `build_type` here since we aren't compiling anything that uses this

.circleci/config.yml
```diff
@@ -222,7 +232,7 @@ jobs:
           at: ~/player
 
       - run: |
-          bazel test --config=ci -- $(bazel query 'kind(".*_test", //...) except filter("ios|swiftui", //...)') -//android/demo:android_instrumentation_test
+          bazel test --config=ci --define build_type=$BUILD_TYPE -- $(bazel query 'kind(".*_test", //...) except filter("ios|swiftui", //...)') -//android/demo:android_instrumentation_test
```
sugarmanz, line 235: Would you also need the `build_type` defined for the release script?

.circleci/config.yml
```diff
@@ -76,6 +76,16 @@ jobs:
...
+      - run:
+          name: Set build type
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "main" ]; then
+              echo "Running on main"
+              echo 'export BUILD_TYPE=release' >> $BASH_ENV
+            else
+              echo "Running on PR"
+              echo 'export BUILD_TYPE=debug' >> $BASH_ENV
+            fi
```
sugarmanz, line 88: I think this strategy could be simplified by adding the `--define build_type=` to the bazel rc for CI below. This way, the `build_type` would be applied to all bazel invocations automatically, instead of requiring you to manually define for each Bazel command.

android/demo/deps.bzl
```diff
@@ -6,7 +6,7 @@ maven_main = [
     "@maven//:com_afollestad_material_dialogs_core",
     "@maven//:com_google_android_material_material",
     #"@maven//:com_squareup_leakcanary_leakcanary_android",
-    "@maven//:com_github_AlexTrotsenko_j2v8_debugger",
```
sugarmanz, line 9: We should be able to remove this altogether now, right? If the correct dependency comes with the android player dep

android/demo/src/main/java/com/intuit/playerui/android/reference/demo/lifecycle/DemoPlayerViewModel.kt
```diff
@@ -24,8 +24,10 @@ class DemoPlayerViewModel(iterator: AsyncFlowIterator) : PlayerViewModel(iterato
         AsyncHydrationTrackerPlugin(),
     )
 
+    public val isDebug = false
```
sugarmanz, line 27: Nit: another source of truth for `config.debuggable`

brocollie08: it is, it's purely for demo because of [this](https://github.com/player-ui/player/pull/443/files#diff-9ee98332d0ca1039bddbbec92d72d2f5cf5b0c4feba501c358b9dca1f23d7c75R33) check, otherwise i think it complains about classes not being available on classpath

.circleci/config.yml
brocollie08: does that not apply to all CI builds regardless of main/PR

brocollie08: nvm I get what you mean now

VERSION
```diff
@@ -1 +1 @@
-0.7.4
\ No newline at end of file
+0.7.42-SNAPSHOT
```
sugarmanz, line 1: revert

sugarmanz (approved):

PR #444, opened 2024-07-22
ReactAsset improve error messages

cehan-Chloe:
This pr obsolete https://github.com/player-ui/player/pull/421 which is pointing to bazel-6 with unverified commits. 

Issue: https://github.com/player-ui/player/issues/376 

What's changed: improve error msg thrown by ReactAsset. more info and react-like errors
Before: if no matched asset found, throw error: `No implementation found for id: bar-id type: bar`

After: 
1. add check for no assets in registry
2. modify existing error msg for no match asset found and give possible alternative with fuse.js
https://www.fusejs.io/demo.html

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs


KetanReddy:
Are there any alternatives to `string-similarity` we can use? It looks like that package isn't being actively maintained so I would be hesitant to add a dependency on it.

cehan-Chloe:
@KetanReddy  replaced `string-similarity` with `fuse.js`. thanks for pointing out. Now ready for review

KetanReddy:
Looks good! Once CI passes you should be good to merge.

KetanReddy (approved):

PR #468, opened 2024-08-01
Add recipes for just

adierkens:
[just](https://github.com/casey/just) has come about as one of the more popular ways of organizing project specific scripts. Given the multi-language and hard-to-find nature of bazel targets, this gives us a clean way of saving and reusing scripts for users. 

```
> just -l                                                                                                                                                                

Available recipes:
    build-all       # Build all targets in the project
    build-core      # Build all core JS/TS files
    build-js        # Build all JS/TS files
    dev-ios         # Generate and open the xcodeproj for Player
    lint-js         # Lint all JS/TS files
    mvn-install     # Install all generated artifacts into the system .m2 repository
    maven-install   # alias for `mvn-install`
    start-docs      # Run a dev server of the main docs page
    start-ios-demo  # Build and run the iOS demo app in a simulator
    start-storybook # Run a dev server of storybook
    test-all        # Test targets in the project
    test-core       # Test all core JS/TS files
    test-js         # Test all JS/TS files
```


justfile
```diff
@@ -0,0 +1,57 @@
+[doc('Build all targets in the project')]
+build-all:
+  bazel build //...
```
sugarmanz, line 3: My 2c, we shouldn't have any `build //*/...` commands in here because:
1. They'll build a lot of intermediate targets that may not be necessary for checking if the code works
2. Not all targets are buildable by themselves. Some of them require the context of what they're being built for, i.e. `cc_library` compilation changes depending on whether it's for JVM or Android. If there is no default set up, likely because is doesn't make sense to build that target by itself, then this would fail.

These scripts should be intent based, "what are you building for" -- testing, publishing, executing.

justfile
```diff
@@ -0,0 +1,57 @@
...
+[doc('Test all JS/TS files')]
+test-js:
+  bazel test -- $(bazel query "kind(js_test, //...)" --output label 2>/dev/null | tr '\n' ' ')
+
+[doc('Test all core JS/TS files')]
+test-core:
+    bazel test //core/...
+
+[doc('Lint all JS/TS files')]
+lint-js:
```
sugarmanz, line 26: Do we need to have language discrimination for common language things? Practically for performance, maybe, but if I want to lint the project, ideally I'd just lint the whole project.

KetanReddy: I don't see any downside to having per language targets in addition to global targets.

CONTRIBUTING.md
```diff
@@ -23,6 +23,9 @@ If the changes are larger (API design, architecture, etc), [opening an issue](ht
 * [Signed Commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification). For convenience it is recommended to set git to sign all commits by default as mentioned [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key)
 
 ## Building and Testing Locally (All platforms)
+
+> This project also contains [just](https://github.com/casey/just) recipes for many common commands. They can be listed using `just -l`
```
KetanReddy, line 27: Should installing `just` get added to the setup/contributing guides?

justfile
adierkens: I feel like people are more likely to want to lint smaller subsets of the project based on what they're updating rather than the full project all the time

CONTRIBUTING.md
adierkens: I figured we would first see how well we like using `just` before fully converting the usage guides over. Then we can update the examples commands and stuff to reference it.

justfile
adierkens: I agree with the general sentiment, but I'm not sure I'd go so far to say we shouldn't have _any_ `//...` commands here. I don't expect them to be used very often, but for things like testing, we may want to test the full project. 

> Not all targets are buildable by themselves

I'll admit I'm not as familiar with the standard patterns for `cc_library` -- but if that truly is the case, then I'd expect we should exclude those targets from the `build-all` target. Maybe it's more beneficial to convert this into a macro for building all of the sub-systems: 

```
build-core
build-plugins
build-android
build-react
build-ios
```

which I assume is similar to what we'd want to do for `lint-all`, `test-all`, etc

sugarmanz: > I'm not sure I'd go so far to say we shouldn't have any //... commands here. I don't expect them to be used very often, but for things like testing

Using `...` expansion for testing is fine, but that's because tests explicitly declare how the dependencies should be built. But I'm sticking to avoiding it for `build` commands -- you probably don't need to build the targets used for publishing locally, at least at the `build-all` level. It's somewhat misleading to expect all the tooling targets to also be build with a `build-all` command.

sugarmanz: This also isn't limited to just the `cc_library` targets - any Bazel target may be transitioned upon to change the underlying toolchain. It's called out in the iOS docs for the Swift-based targets:
https://github.com/player-ui/player/tree/main/ios#bazel

PR #471, opened 2024-08-02
On update race condition

brocollie08:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.8.1--canary.471.16682</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.8.1--canary.471.16682
</details>


core/player/src/view/view.ts
```diff
@@ -94,21 +96,29 @@ export class ViewInstance implements ValidationProvider {
...
     this.resolverOptions = resolverOptions;
     this.hooks.onTemplatePluginCreated.tap("view", (templatePlugin) => {
       this.templatePlugin = templatePlugin;
     });
+    resolverOptions.flowController.current?.hooks.transition.tap("view", (state1, state2) => {
+      this.transitioning = true;
+    })
+    resolverOptions.flowController.current?.hooks.afterTransition.tap("view", (flowInstance) => {
+      this.transitioning = false;
+    })
```
intuit-svc, line 110: You'll want to tap the `flow` hook on the `flowController` instead of using `current`

adierkens, line 110: You'll want to use the `.flow` hook on the flow-controller instead of using `current` directly.

brocollie08: while accessible. the call of that hook takes place earlier than the tap due to flow/view scope diff, so I moved the logic out to the view controller

sugarmanz (approved):

PR #479, opened 2024-08-05
Update link handling in docs for external

adierkens:


KetanReddy:
Do we still need this with the doc site refresh?

PR #483, opened 2024-08-07
externalize @player-ui/player from native bundles

sugarmanz:
Currently, the plugin native bundles contain an entire copy of Player - this PR externalizes the Player from the native bundles, reducing size burden of the native bundles.

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/core/player/BUILD
+++ b/core/player/BUILD
@@ -2,16 +2,22 @@ load("@rules_player//javascript:defs.bzl", "js_pipeline")
 load("@npm//:defs.bzl", "npm_link_all_packages")
 load("//tools:defs.bzl", "NATIVE_BUILD_DEPS", "tsup_config", "vitest_config")
 
+# TODO: Would be nice to macro all the native things to avoid missing this setup (consequence is bundling Player in a plugin bundle)
+NATIVE_BUNDLE = "Player"
+
 npm_link_all_packages(name = "node_modules")
 
-tsup_config(name = "tsup_config")
+tsup_config(
+    name = "tsup_config",
+    native_bundle = NATIVE_BUNDLE,
+)
 
 vitest_config(name = "vitest_config")
 
 js_pipeline(
     package_name = "@player-ui/player",
     build_deps = NATIVE_BUILD_DEPS,
-    native_bundle = "Player",
+    native_bundle = NATIVE_BUNDLE,
     test_deps = [
         ":node_modules/@player-ui/make-flow",
         "//:vitest_config",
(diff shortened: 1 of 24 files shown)


KetanReddy:
Can we do this as a part of the rule to minimize the config needed on the usage side?

sugarmanz:
> Can we do this as a part of the rule to minimize the config needed on the usage side?

It would likely require pulling more than what I touch out of this repo, since we define the template in this repo too:
https://github.com/player-ui/player/pull/483/files#diff-d20cc7f8a4827655e2b6e1c235a7ed8e4f4fa9446fb0c84e3f71da57cfef2f0eR19

KetanReddy:
> > Can we do this as a part of the rule to minimize the config needed on the usage side?
> 
> It would likely require pulling more than what I touch out of this repo, since we define the template in this repo too: https://github.com/player-ui/player/pull/483/files#diff-d20cc7f8a4827655e2b6e1c235a7ed8e4f4fa9446fb0c84e3f71da57cfef2f0eR19

Ah, sorry. I thought we had some base tsup config that was expanded in our rules.

core/player/BUILD
```diff
@@ -2,16 +2,22 @@ load("@rules_player//javascript:defs.bzl", "js_pipeline")
 load("@npm//:defs.bzl", "npm_link_all_packages")
 load("//tools:defs.bzl", "NATIVE_BUILD_DEPS", "tsup_config", "vitest_config")
 
+# TODO: Would be nice to macro all the native things to avoid missing this setup (consequence is bundling Player in a plugin bundle)
```
KetanReddy, line 5: Do you have any estimate for how much effort this would require? If its not huge would it be worth it to just do it now?

tools/tsup.config.native.ts.tmpl
```diff
@@ -0,0 +1,12 @@
+import { createConfig } from "%PREFIX%tsup.config";
+import ESBuildPluginExternalGlobal from "esbuild-plugin-external-global";
+
+const overrides: any = {};
+if (process.env.PLAYER_NATIVE_BUNDLE) {
+    overrides.esbuildPlugins = [ESBuildPluginExternalGlobal.externalGlobalPlugin({
+       // TODO: Would be cool to automatically populate this given all the native bundles we create
+      "@player-ui/player": "globalThis.Player",
```
sugarmanz, line 8: @brocollie08

PR #488, opened 2024-08-12
jvm and ios-async-node-ability-to-remove-resolved-async-node

sakuntala-motukuri:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


plugins/async-node/ios/Tests/AsynNodePluginTests.swift
```diff
@@ -292,6 +284,166 @@ class AsyncNodePluginTests: XCTestCase {
...
+
+    func testAsyncNodeWithNullNode() {
+        let handlerExpectation = XCTestExpectation(description: "null node handled")
+
+        let context = JSContext()
+
+        let resolveHandler: AsyncHookHandler = { node in
+            handlerExpectation.fulfill()
+            XCTAssertNil(node)
+            return .singleNode(.concrete(JSValue()))
```
sugarmanz, line 296: This is testing the `.singleNode` case, right? But with an empty `JSValue`? We'd want to test the new enum case here, such that the addition would suffice towards resolving with "empty" (`nil` or `null` or `undefined` - these are really all different representations of empty, which we concretely tie back to a specific enum case).

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
```diff
@@ -292,6 +284,166 @@ class AsyncNodePluginTests: XCTestCase {
...
+
+    func testAsyncNodeWithUndefinedNode() {
+        let handlerExpectation = XCTestExpectation(description: "undefined node handled")
+
+        let context = JSContext()
+
+        let resolveHandler: AsyncHookHandler = { node in
+            handlerExpectation.fulfill()
+            XCTAssertNil(node)
+            return .singleNode(.concrete(JSValue()))
```
sugarmanz, line 321: Same as `null` test above

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
```diff
@@ -292,6 +284,166 @@ class AsyncNodePluginTests: XCTestCase {
...
+
+        plugin.context = context
+
+        XCTAssertNotNil(asyncNodePluginPlugin.context)
+
+        let player = HeadlessPlayerImpl(plugins: [ReferenceAssetsPlugin(), plugin], context: context ?? JSContext())
+
+        player.start(flow: .asyncNodeJson, completion: {_ in})
+
+        wait(for: [handlerExpectation], timeout: 5)
```
sugarmanz, line 310: Where is the replacement value tested? We'd want to have similar assertions to the other platform tests, where we test the resolution did/or did not trigger another view update and what the view tree looks like after the update happens.

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
```diff
@@ -292,6 +284,166 @@ class AsyncNodePluginTests: XCTestCase {
...
+                        ReplacementNode.encodable(AssetPlaceholderNode(asset: PlaceholderNode(id: "text-2", type: "text", value: "2nd value in the multinode"))),
+                        ReplacementNode.encodable(AsyncNode(id: "id-1"))])
+                } else if count == 3 {
+                    return .singleNode(ReplacementNode.encodable(
+                        AssetPlaceholderNode(asset: PlaceholderNode(id: "text", type: "text", value: "3rd value in the multinode"))
+                    ))
+                }
+
+                return .singleNode(ReplacementNode.concrete(context?.evaluateScript("") ?? JSValue()))
+            }
```
sugarmanz, line 366: This isn't using the callback mechanism, right? This looks to be just using the return value approach towards resolving the values, which would only enable a single resolution. It looks like this handler is triggered multiple times, but for explicitly different async nodes (`ReplacementNode.encodable(AsyncNode(id: "id-1"))])`.

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -228,4 +228,64 @@ internal class AsyncNodePluginTest : PlayerTest() {
...
+                    count++
+                    update = asset
+                    if (count == 3) cont.resume(true) {}
+                }
+            }
+            player.start(asyncNodeFlowSimple)
+        }
+
+        Assertions.assertTrue(count == 3)
+        Assertions.assertTrue((update?.get("actions") as List<*>).isNotEmpty())
```
sugarmanz, line 268: Can we assert what the values are in the tree at this point? It's interesting to see the `callback` mechanism used in sequence with the return value. I'm interested in testing out that the different mechanisms are working as expected.

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -228,4 +228,64 @@ internal class AsyncNodePluginTest : PlayerTest() {
...
+                v?.hooks?.onUpdate?.tap { asset ->
+                    count++
+                    update = asset
+                    if (count == 2) cont.resume(true) {}
+                }
+            }
+            player.start(asyncNodeFlowSimple)
+        }
+        Assertions.assertTrue(count == 2)
+        Assertions.assertTrue((update?.get("actions") as List<*>).isNotEmpty())
```
brocollie08, line 289: similar comment here, but in this case, is an `emptyList` representative of `undefined`? and if so, should this be not empty?

plugins/async-node/ios/Sources/AsyncNodePlugin.swift
```diff
@@ -42,58 +43,70 @@ public class AsyncNodePlugin: JSBasePlugin, NativePlugin {
...
+
+              case .nullOrUndefinedNode(let replacementNode):
+                  switch replacementNode {
+                  case .encodable(let encodable):
+                      let encoder = JSONEncoder()
+                          let res = try encoder.encode(encodable)
+                          result = context.evaluateScript("(\(String(data: res, encoding: .utf8) ?? ""))") as JSValue
+                  case .concrete(let jsValue):
+                      return jsValue
+                  }
```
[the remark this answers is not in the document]
sakuntala-motukuri: Totally agree that its actually getting handled by other cases , so I was thinking across that do we actually need to handle this case explicitly since its already being handled !

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
sakuntala-motukuri: This actually lies inline to above comment that I was pretty thinking in the lines that be it (nil or null or undefined) these are already being handled , so I have tested it with the existing singlenode

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
sakuntala-motukuri: same as above comment

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
sakuntala-motukuri: this needs an update

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
brocollie08: I just realized this brings up another question, should this get a second update at all if you're resolving it with "`undefined`" right away?

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -228,4 +228,64 @@ internal class AsyncNodePluginTest : PlayerTest() {
...
         Assertions.assertTrue(true)
     }
+
+    @TestTemplate
+    fun `handle multiple updates through callback mechanism`() = runBlockingTest {
+        var update: Asset? = null
+        var updateCount = 0
+
+        plugin?.hooks?.onAsyncNode?.tap("") { _, node, callback ->
+            when (updateCount) {
```
brocollie08, line 238: Hmm I don't think this makes sense from a functional POV.

This tap is for the original resolve, where you are given the async node and a callback to be stored for reference later in case an update is needed.

The code here actually uses the callback to update the original node within the "first" tap, which means line 252 doesn't ever get hit (that return value is supposed to be the resolve for the original async node)

Furthermore, updateCount 0 conceptually doesn't click with me either, because the updateCount should be 1 even before the async node is resolved (from first tree update before it's resolved by the consumer)

You can reference the core test [here](https://github.com/player-ui/player/pull/404/files#diff-fbc47a3e301bdf595a6d686ded538fe6767e534af8821c421f97dc56264d1332R146) to see how the update was intended to be used

plugins/async-node/ios/Sources/AsyncNodePlugin.swift
```diff
@@ -42,58 +43,58 @@ public class AsyncNodePlugin: JSBasePlugin, NativePlugin {
...
+               print("Value of callback \(callback)")
+               // hook value is the original node
+               guard let asyncHookHandler = self.asyncHookHandler else {
+                   return JSValue()
+               }
+
+               let replacementNode = try await (asyncHookHandler)(node)
+               var result: JSValue?
+
+               switch replacementNode {
```
nancywu1, line 63: for lines 63 to 95 probably want to put this inside a function that takes a `AsyncNodeHandlerType` type so that the user can pass in those types from the `callback` in the test when we call `callback.call(withArguments: [` we can reuse this function

jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/Promise.kt
```diff
@@ -178,7 +178,7 @@ public val Runtime<*>.Promise: Promise.Api get() = getObject("Promise")?.let { p
 } ?: throw PlayerRuntimeException("'Promise' not defined in runtime")
 
 /** Helper to bridge complex [Promise] logic with the JS promise constructor */
-public fun <T : Any> Runtime<*>.Promise(block: suspend ((T) -> Unit, (Throwable) -> Unit) -> Unit): Promise {
+public fun <T : Any?> Runtime<*>.Promise(block: suspend ((T) -> Unit, (Throwable) -> Unit) -> Unit): Promise {
```
brocollie08, line 181: is there a reason now it's nullable?

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -228,4 +232,162 @@ internal class AsyncNodePluginTest : PlayerTest() {
...
+            updateContent = callback
+            val result = suspendCoroutine { cont ->
+                deferredResolve = { value ->
+                    cont.resume(value)
+                }
+            }
+            BailResult.Bail(result)
+        }
+
+        var viewUpdateContinuation: Continuation<Asset?>? = null
```
brocollie08, line 254: hmm is this assigned anywhere? what's the thinking behind this, I only see this being assigned as null

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -228,4 +232,162 @@ internal class AsyncNodePluginTest : PlayerTest() {
...
+            "asset" to mapOf(
+                "id" to "asset-1",
+                "type" to "action",
+                "value" to "New asset!",
+            ),
+        )
+
+        deferredResolve!!.invoke(listOf(viewObject))
+
+        Assertions.assertEquals(1, count)
```
brocollie08, line 291: I _think_ I'm following this, can you tell me why this is 1 and not 2 at this point

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -228,4 +232,162 @@ internal class AsyncNodePluginTest : PlayerTest() {
...
+
+            BailResult.Bail(result)
+        }
+
+        var viewUpdateContinuation: Continuation<Asset?>? = null
+        var count = 0
+        player.hooks.view.tap { v ->
+            v?.hooks?.onUpdate?.tap { asset ->
+                count++
+                println("Update after callback undefined node $count: $asset") // Debug statement
```
brocollie08, line 342: remove possibly

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -228,4 +232,162 @@ internal class AsyncNodePluginTest : PlayerTest() {
...
+                viewUpdateContinuation?.resume(asset)
```
brocollie08, line 343: same comment here as above with this continuation

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
sakuntala-motukuri: This is used [here](https://github.com/player-ui/player/pull/488/files/691749196a68e7fedfade8be8ac931f085c95698#diff-ac6259632cb11fb85e6116a391926ea88917bc5f23faba6c0657f014a09d74c0R258) When an update is received, the continuation is resumed with the new asset, allowing the coroutine to continue.

brocollie08: that linked spot, is it not `null` at that point? I'm curious because I can't seem to find a place where this is assigned to something other than null

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
brocollie08: > The assertion is placed after the first update event has occurred, so the count should be 1. If the count were 2, it would indicate that two update events have occurred,

Is this `deferredResolve` not triggering the second update? First update being the initial view without any async resolve?

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
```diff
@@ -292,6 +292,126 @@ class AsyncNodePluginTests: XCTestCase {
...
+
+        let plugin = AsyncNodePlugin { _ in
+            return .singleNode(.concrete(JSValue()))
+        }
+        plugin.context = context
+
+        let replacementResult = self.handleReplacement(plugin: plugin, context: context, count: count)
+        args = [plugin.handleAsyncNodeReplacement(replacementResult, context: context) ?? JSValue()]
+
+        callback.call(withArguments: args)
```
nancywu1, line 348: shouldnt be invoking the actual `callback`  function in the handler here from this comment https://github.com/player-ui/player/pull/488#discussion_r1720166394 this handler we are in is for the  original resolve, we are just using the handler to assigning the callback for use at a later time to update the original node. 
You can just return a single node with the value "new node from the hook 1" and then for the callback return a single node "new node from the hook 2" which we expect to replace the old value

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
sakuntala-motukuri: removed , its just for testing

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
sakuntala-motukuri: I have removed the callback in the handler and just invoking the callbackFunction after the original node update. Also "**You can just return a single node with the value "new node from the hook" and then for the callback return a single node "new node from the callback" which we expect to replace the old value**" is already being returned by the variable **args** which being passed as an argument to the callback function at a later point

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
```diff
@@ -292,6 +292,125 @@ class AsyncNodePluginTests: XCTestCase {
...
+
+    player.start(flow: .asyncNodeJson, completion: { _ in})
+
+    wait(for: [handlerExpectation, textExpectation], timeout: 5)
+
+    XCTAssert(count == 2)
+    XCTAssertEqual(expectedMultiNode1Text, "new node from the hook 1")
+
+    if let callbackFunction = callbackFunction {
+            callbackFunction.call(withArguments: args)
```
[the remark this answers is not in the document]
sakuntala-motukuri: updated

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
```diff
@@ -292,6 +292,147 @@ class AsyncNodePluginTests: XCTestCase {
         XCTAssert(count == 3)
         XCTAssertEqual(expectedMultiNode2Text, "new node from the hook 2")
     }
+
+    func handleEmptyNode() {
```
nancywu1, line 296: this it possible to change this test to have similar logic to the android test https://github.com/player-ui/player/pull/488/files#diff-ac6259632cb11fb85e6116a391926ea88917bc5f23faba6c0657f014a09d74c0R320?

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
```diff
@@ -292,6 +292,147 @@ class AsyncNodePluginTests: XCTestCase {
...
+    
+    replacementResult = AsyncNodeHandlerType.emptyNode
+    
+    args = [plugin.handleAsyncNodeReplacement(replacementResult) ?? JSValue()]
+    
+    _ = callbackFunction?.call(withArguments: args)
+
+    XCTAssert(count == 4)
+    // asset that the value at index 0 for the object
+    XCTAssertEqual(expectedMultiNode3Text, "undefined")
```
nancywu1, line 433: hmm i would expect `textString3` corresponding to `expectedMultiNode3Text`  to have a value of the first node in the original mulitnode of the `action` 

this one 

``` {
               "asset": {
                 "id": "action",
                 "type": "action",
                 "exp": "{{count}} = {{count}} + 1",
                 "label": {
                   "asset": {
                     "id": "test",
                     "type": "text",
                     "value": "test"
                   }
                 }
               }
             },
 ```
             
you can check for the actions label value of `test` but you'll need to change the accessing of the value on line 380

**Review by sugarmanz — DISMISSED** [review]

File: plugins/async-node/ios/Tests/AsynNodePluginTests.swift (hunk 17 above)
Comment (sakuntala-motukuri) [reply plugins/async-node/ios/Tests/AsynNodePluginTests.swift]: updated

File: plugins/async-node/ios/Tests/AsynNodePluginTests.swift [hunk 19]
```diff
@@ -292,6 +292,147 @@ class AsyncNodePluginTests: XCTestCase {
...
+                    let newText4 = val
+                        .objectForKeyedSubscript("values")
+                        .objectAtIndexedSubscript(1)
+                        .objectForKeyedSubscript("asset")
+                        .objectForKeyedSubscript("value")
+                    guard let textString4 = newText4?.toString() else { return XCTFail("newText was not a string") }
+
+                    expectedMultiNode3Text = textString3
+                    expectedMultiNode4Text = textString4
+                    textExpectation2.fulfill()
```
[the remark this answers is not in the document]
sakuntala-motukuri: updated

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
sakuntala-motukuri: updated

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
```diff
@@ -292,6 +292,226 @@ class AsyncNodePluginTests: XCTestCase {
...
+    
+    args = [plugin.handleAsyncNodeReplacement(replacementResult) ?? JSValue()]
+    
+    _ = callbackFunction?.call(withArguments: args)
+
+    XCTAssert(count == 4)
+    // asset that the value at index 0 for the object
+    XCTAssertEqual(expectedMultiNode3Text, "test")
+    XCTAssertEqual(expectedMultiNode4Text, "undefined")
+}
```
nancywu1, line 514: to fix indent formatting, you can select everything in the file or just the lines you want to indent and press ctrl + i

ios/core/Sources/Types/Hooks/Hook.swift
```diff
@@ -176,3 +176,47 @@ public class AsyncHook<T>: BaseJSHook where T: CreatedFromJSValue {
          self.hook.invokeMethod("tap", withArguments: [name, JSValue(object: tapMethod, in: context) as Any])
      }
 }
+
+/**
+ This class represents an object in the JS runtime that can be tapped into
+ and returns a promise that resolves when the asynchronous task is completed
+ */
```
nancywu1, line 183: should update comment to mention the 2 parameters, like the comment for Hook2

plugins/async-node/ios/Sources/AsyncNodePlugin.swift
```diff
@@ -42,57 +40,62 @@ public class AsyncNodePlugin: JSBasePlugin, NativePlugin {
...
+                } catch {
+                    return nil
+                }
+            case .concrete(let jsValue):
+                return jsValue
+            }
 
+        case .emptyNode:
             return nil
+        }
```
nancywu1, line 81: needs indenting fix here too

plugins/async-node/ios/Sources/AsyncNodePlugin.swift
```diff
@@ -42,57 +40,62 @@ public class AsyncNodePlugin: JSBasePlugin, NativePlugin {
...
-                        }
-                    }
-                })
-
-                return context.objectForKeyedSubscript("Array").objectForKeyedSubscript("from").call(withArguments: [jsValueArray])
-
-            case .singleNode(let replacementNode):
-                switch replacementNode {
-
+func handleAsyncNodeReplacement(_ replacementNode: AsyncNodeHandlerType) -> JSValue? {
```
nancywu1, line 43: im thinking a better name for this function can be "replacementNodeToJSValue" or "convertReplacementNodeToJSValue"?
also doc comment before this function would be good

plugins/async-node/ios/Tests/AsynNodePluginTests.swift
sakuntala-motukuri: done

plugins/async-node/ios/Sources/AsyncNodePlugin.swift
```diff
@@ -7,124 +7,149 @@
...
+            let jsValueArray = replacementNodes.compactMap { node in
+                switch node {
+                    
+                case .encodable(let encodable):
+                    return convertReplacementNodeToJSValue(.encodable(encodable), context: context)
+                case .concrete(let jsValue):
+                    return convertReplacementNodeToJSValue(.concrete(jsValue), context: context)
+                    
+                }
+            }
```
hborawski, line 88: this switch is not needed, since `convertReplacementNodeToJSValue` already switches inside

plugins/async-node/ios/Sources/AsyncNodePlugin.swift
```diff
@@ -7,124 +7,149 @@
...
+            
+        case .singleNode(let replacementNode):
+            switch replacementNode {
+                
+            case .encodable(let encodable):
+                return convertReplacementNodeToJSValue(.encodable(encodable), context: context)
+            case .concrete(let jsValue):
+                return convertReplacementNodeToJSValue(.concrete(jsValue), context: context)
+                
+            }
```
hborawski, line 99: this switch is not needed, since `convertReplacementNodeToJSValue` already switches inside

plugins/async-node/ios/Sources/AsyncNodePlugin.swift
```diff
@@ -7,124 +7,149 @@
...
+    
+    /**
+     Converts a given `ReplacementNode` to a `JSValue` that can be used in the JavaScript context.
+     
+     - Parameters:
+     - replacementNode: The `ReplacementNode` to be converted.
+     - context: The `JSContext` in which the `JSValue` will be used.
+     - Returns: A `JSValue` representing the given `ReplacementNode`, or `nil` if the conversion fails.
+     */
+    func convertReplacementNodeToJSValue(_ replacementNode: ReplacementNode, context: JSContext) -> JSValue? {
```
hborawski, line 51: this can be moved to an extension of `ReplacementNode` to simplify and remove the first parameter

you can also call it directly to unit test it and force the `catch` block

hborawski (approved):

sugarmanz (approved):

PR #512, opened 2024-09-20
revert async node flatten

cehan-Chloe:
1. Revert flatten nested node change
2. update tests for all platforms

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs


brocollie08:
might need to change test on other platforms too

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -144,7 +144,7 @@ internal class AsyncNodePluginTest : PlayerTest() {
             player.start(asyncNodeFlowSimple)
         }
         Assertions.assertTrue(count == 2)
-        Assertions.assertEquals(3, update?.getList("actions")?.size)
+        Assertions.assertEquals(2, update?.getList("actions")?.size)
```
brocollie08, line 147: if you're changing this to 2, you should probably check the nested one is also of size 2

brocollie08 (approved):

PR #519, opened 2024-09-27
iOS: Expand AnyType to handle deeply nested AnyType for beacon encoding

hborawski:
`AnyType` decoded as a part of the asset tree did not have `AnyTypeDecodingContext` to allow for `[String: Any]` decoding, that context is now added in `BaseRegistry` when decoding the asset tree

Including complex metadata for beacons would not reencode when passing the beacon from Swift to JSC in all scenarios. 

```json
{
  "asset": {
    "id": "action-1",
    "type": "action",
    "exp": "{{count}} = {{count}} + 1",
    "label": {
      "asset": {
        "id": "action-label-1",
        "type": "text",
        "value": "Clicked {{count}} times"
      }
    },
    "metaData": {
      "beacon": {
        "array": [{
          "key": "value"
        }],
        "string": "value"
      }
    }
  }
}
```

In this scenario, this metadata would be [decoded by the ActionAsset](https://github.com/player-ui/player/blob/main/plugins/reference-assets/swiftui/Sources/SwiftUI/ActionAsset.swift#L25), which includes `AnyType` for the [beacon metadata](https://github.com/player-ui/player/blob/main/ios/core/Sources/Types/Assets/Wrappers.swift#L89). When sending this decoded type through `BeaconContext.beacon`, the encoding was failing for the nested keys, as the entire `beacon` key would be decoded as `AnyType` with `.anyDictionary([String: Any])`. Encoding the `Any` values failed.

`CustomEncodable` did not work against `Any`, so instead, I decode `Any` as `AnyType` which is then encodable, and can be used as the backing type for encoding

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/ios/core/Sources/Types/Generic/AnyType.swift
+++ b/ios/core/Sources/Types/Generic/AnyType.swift
@@ -34,6 +34,8 @@ public enum AnyType: Hashable {
             hasher.combine(data)
         case .anyDictionary(let data):
             hasher.combine(data as NSDictionary)
+        case .anyArray(let data):
+            hasher.combine(data as NSArray)
         case .unknownData:
             return
         }
@@ -73,6 +75,13 @@ public enum AnyType: Hashable {
      */
     case anyDictionary(data: [String: Any])
 
+    /**
+     The underlying data was an array of varied value types
+
+     **This requires the decoder to add `AnyTypeDecodingContext` to the decoders userInfo**
+     */
+    case anyArray(data: [Any])
+
     /// The underlying data was not in a known format
     case unknownData
 }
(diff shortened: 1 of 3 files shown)


ios/core/Sources/Types/Generic/AnyType.swift
```diff
@@ -154,6 +167,12 @@ struct CustomEncodable: CodingKey {
         self.stringValue = key
         if let encodable = encodable as? Encodable {
             self.data = encodable
+        } else if
+            let encodable,
+            let data = try? JSONSerialization.data(withJSONObject: encodable, options: .fragmentsAllowed),
```
hborawski, line 172: `.fragmentsAllowed` lets us encode `String` here as a top level type, as we dont know what the value type will be

nancywu1 (approved):

nancywu1 (approved):

Issue #523, opened 2024-10-02
enhance iOS tapable logger to handle variadic args from the JS log


nancywu1:
**Is your feature request related to a problem? Please describe.**
Currently in iOS we only capture the first message from JS log which allows multiple arguments to be logged

**Describe the solution you'd like**
change the hooks type to [Any] from String so it can pass all the arugments https://github.com/player-ui/player/blob/main/ios/logger/Sources/TapableLogger.swift#L174-L180 

the log function passed in to be used in JavaScriptCore should collect the args into an array to be handled in iOS

PR #524, opened 2024-10-02
ios: enhance tapable logger to handle variadic args from the JS log

nancywu1:
https://github.com/player-ui/player/issues/523

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

# Release Notes
Prefixing message with the format `[Player] [\(logLevel)]:`  moved out of the Tapable logger and moved to PrintLoggerPlugin. Any consumers using their own LoggerPlugin will need to append the logLevel if they want to print it 

Before
```swift
public class CustomLoggingPlugin: NativePlugin {
    
    public let pluginName = "CustomLoggingPlugin"

    public func apply<P>(player: P) where P: HeadlessPlayer {
        guard let player = player as? SwiftUIPlayer else { return }
        player.logger.logLevel = .trace
        player.logger.hooks.trace.tap(name: pluginName, { print("Custom message")\(($0))" ) })
        ...
    }
```

After
```swift
public class CustomLoggingPlugin: NativePlugin {
    
    public let pluginName = "CustomLoggingPlugin"

    public func apply<P>(player: P) where P: HeadlessPlayer {
        guard let player = player as? SwiftUIPlayer else { return }
        player.logger.logLevel = .trace
        let prefixedMessage = "[Player] [trace]: "
        player.logger.hooks.trace.tap(name: pluginName, { print("\(prefixedMessage) Custom message \(($0))" ) })
        ...
    }
```

## Breaking Changes

Any usage of the `player.logger.hooks` taps will have breaking changes in the callback because the calls have been changed to provide a `[Any]` type instead of `String` so it can be returned in the form of messages instead of a single message. Unless nothing is done in the callback to access the value but just to print it, there should be breaking changes

Example:
```swift

// this should be no breaking change
player.logger.hooks.trace.tap(name: "log", { print("\(($0))" ) })

// if `values` should be accessed in anyway, i.e want the first value, or want to seperate the values
 player.logger.hooks.debug.tap(name: "log") { values in
            // values is of type [Any], if you want to print only the first value
            print("\((message as? [String])?.first))" )
        }
```

<details>
  <summary>📦 Published PR as canary version: <code>0.9.2--canary.524.17782</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.9.2--canary.524.17782
</details>

### DIFF
--- a/plugins/console-logger/ios/Sources/PrintLoggerPlugin.swift
+++ b/plugins/console-logger/ios/Sources/PrintLoggerPlugin.swift
@@ -35,10 +35,14 @@ public class PrintLoggerPlugin: NativePlugin {
     public func apply<P>(player: P) where P: HeadlessPlayer {
         player.logger.logLevel = logLevel
         player.logger.hooks.prefixMessage.tap(name: pluginName) { .bail("[Player] [\($0.description)]: ") }
-        player.logger.hooks.trace.tap(name: pluginName, { print($0) })
-        player.logger.hooks.debug.tap(name: pluginName, { print($0) })
-        player.logger.hooks.info.tap(name: pluginName, { print($0) })
-        player.logger.hooks.warn.tap(name: pluginName, { print($0) })
-        player.logger.hooks.error.tap(name: pluginName, { print($0 ?? "", $1?.localizedDescription ?? "") })
+
+        let prefixedMessage = player.logger.hooks.prefixMessage.call(logLevel) ?? ""
+
+        player.logger.hooks.trace.tap(name: pluginName, { print("\(prefixedMessage)\(($0))" ) })
+        player.logger.hooks.debug.tap(name: pluginName, { print("\(prefixedMessage)\(($0))" ) })
+        player.logger.hooks.info.tap(name: pluginName, { print("\(prefixedMessage)\(($0))" ) })
+        player.logger.hooks.warn.tap(name: pluginName, { print("\(prefixedMessage)\(($0))" ) })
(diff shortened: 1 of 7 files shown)


plugins/console-logger/ios/Sources/PrintLoggerPlugin.swift
```diff
@@ -35,10 +35,14 @@ public class PrintLoggerPlugin: NativePlugin {
     public func apply<P>(player: P) where P: HeadlessPlayer {
         player.logger.logLevel = logLevel
         player.logger.hooks.prefixMessage.tap(name: pluginName) { .bail("[Player] [\($0.description)]: ") }
-        player.logger.hooks.trace.tap(name: pluginName, { print($0) })
-        player.logger.hooks.debug.tap(name: pluginName, { print($0) })
-        player.logger.hooks.info.tap(name: pluginName, { print($0) })
-        player.logger.hooks.warn.tap(name: pluginName, { print($0) })
-        player.logger.hooks.error.tap(name: pluginName, { print($0 ?? "", $1?.localizedDescription ?? "") })
+
+        let prefixedMessage = player.logger.hooks.prefixMessage.call(logLevel) ?? ""
```
hborawski, line 39: i think we can remove the `prefixMessage` hook from `TapableLogger`, and just have that be opinionated behavior of `PrintLoggerPlugin`

hborawski (approved):

PR #563, opened 2024-12-18
Hermes reference tracking

brocollie08:
No longer overloading the global JNI table, tracking references in C++ instead to avoid over-usage crash

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.10.1--canary.563.18922</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.10.1--canary.563.18922
</details>


jvm/hermes/src/main/jni/JHermesRuntime.h
```diff
@@ -40,47 +41,41 @@ class JHermesRuntime : public HybridClass<JHermesRuntime, JJSIRuntime> {
...
-        //       weak_ref was out of scope, it was already released. But this doesn't seem to be true,
-        //       otherwise we wouldn't be getting segfaults. Making them global does prevent the segfaults,
-        //       reaffirming the above. Since all the wrapper classes are holding are pointers
-        //       this probably isn't a huge deal, but is certainly an inefficiency we should look
-        //       to remove. I'd like to just create weak_ptrs on the JSI values themselves,
-        //       which requires the holders to use shared_ptrs and expose a getter for accessing.
-        //       Then we'd also need a way to hold arbitrary weak_ptrs. We could also maybe try
-        //       to get weak refs to _just_ the CXX parts of the hybrid class to avoid leaking things
-        //       on the JVM side?
-        // TODO: Maybe we could do a periodic pruning of the vector to remove obsolete refs?
```
sugarmanz, line 72: Beautiful - removing longstanding paragraphs of TODOs are always great to see.

jvm/hermes/src/main/jni/JHermesRuntime.h
```diff
@@ -40,47 +41,41 @@ class JHermesRuntime : public HybridClass<JHermesRuntime, JJSIRuntime> {
...
+            ptr.second.reset();
+        }
+        for (auto& ptr : *runtimeScope_->functionScope) {
+            ptr.second.reset();
+        }
+        for (auto& ptr : *runtimeScope_->arrayScope) {
+            ptr.second.reset();
+        }
+        for (auto& ptr : *runtimeScope_->symbolScope) {
+            ptr.second.reset();
```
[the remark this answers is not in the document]
sugarmanz: Secondarily, this should also probably clear the keys too - as we discovered earlier.

jvm/hermes/src/main/jni/JJSIValue.cpp
```diff
@@ -39,8 +39,7 @@ void JJSIPreparedJavaScript::registerNatives() {
 }
```
sugarmanz, line 39: Can't put a comment on a diff that isn't modified - comment at the top of this file can now be removed as well:
https://github.com/player-ui/player/blob/dc3c3ab6fbc032b6dc44bdda553cfb5b616d93bf/jvm/hermes/src/main/jni/JJSIValue.cpp#L6-L15

jvm/hermes/src/main/jni/JHermesRuntime.h
```diff
@@ -3,6 +3,7 @@
 #include <hermes/hermes.h>
 #include <fbjni/fbjni.h>
 
+#include "RuntimeScope.h"
```
sugarmanz, line 6: Shouldn't need this include as it'd come from the following `#include "JJSIValue.h"` include

jvm/hermes/src/main/jni/JJSIValue.h
```diff
@@ -4,9 +4,11 @@
 #include <jsi/jsi.h>
 #include <fbjni/fbjni.h>
 #include <fbjni/ByteBuffer.h>
+#include "RuntimeScope.h"
 
 using namespace facebook::jni;
 using namespace facebook::jsi;
+using namespace std;
```
sugarmanz, line 11: If we're going to be `using namespace std`, we should actually remove all the `std::` references.

Probably would help to set up a linter.

jvm/hermes/src/main/jni/JJSIValue.h
```diff
@@ -115,9 +112,9 @@ class JJSIValue : public JJSIValueHybridClass {
     static constexpr auto kJavaDescriptor = "Lcom/intuit/playerui/jsi/Value;";
     static void registerNatives();
 
-    static local_ref<jhybridobject> fromBool(alias_ref<jclass>, bool b);
-    static local_ref<jhybridobject> fromDouble(alias_ref<jclass>, double d);
-    static local_ref<jhybridobject> fromInt(alias_ref<jclass>, int i);
+    static local_ref<jhybridobject> fromBool(alias_ref<jclass>, alias_ref<JRuntimeThreadContext>, alias_ref<JJSIRuntime::jhybridobject> jRuntime, bool b);
+    static local_ref<jhybridobject> fromDouble(alias_ref<jclass>, alias_ref<JRuntimeThreadContext>, alias_ref<JJSIRuntime::jhybridobject> jRuntime, double d);
+    static local_ref<jhybridobject> fromInt(alias_ref<jclass>, alias_ref<JRuntimeThreadContext>, alias_ref<JJSIRuntime::jhybridobject> jRuntime, int i);
```
sugarmanz, line 117: Primitive `Value`s (applies to boolean, double, int) should also not need to be tracked because they're not backed by runtime data. Meaning, they shouldn't need either runtime thread context or the runtime args themselves.

This'd result in reverting the API changes for the primitive methods too, as they won't need the runtime thread context or runtime args.

jvm/hermes/src/main/jni/JJSIValue.h
```diff
@@ -129,7 +126,14 @@ class JJSIValue : public JJSIValueHybridClass {
     static local_ref<jhybridobject> createFromJsonUtf8(alias_ref<jclass>, alias_ref<JRuntimeThreadContext>, alias_ref<JJSIRuntime::jhybridobject> jRuntime, alias_ref<JByteBuffer> json);
     static bool strictEquals(alias_ref<jclass>, alias_ref<JRuntimeThreadContext>, alias_ref<JJSIRuntime::jhybridobject> jRuntime, alias_ref<jhybridobject> a, alias_ref<jhybridobject> b);
 
-    explicit JJSIValue(Value&& value) : HybridClass(), value_(std::make_unique<Value>(std::move(value))) {}
+    explicit JJSIValue(std::shared_ptr<RuntimeScope> scope, Value&& value) : HybridClass(), scope_(scope) {
+        // internally creates unique ptr
+        scope->trackValue(this, std::move(value));
+    }
```
sugarmanz, line 132: If/when there is a singular API to track arbitrary runtime types, we should try to capture this as a common _super_ constructor for `JJSIValueHybridClass` to enforce for any subclass (there are some new ones that may exist in the future).

jvm/hermes/src/main/jni/JJSIValue.h
```diff
@@ -148,21 +152,37 @@ class JJSIValue : public JJSIValueHybridClass {
     local_ref<JJSIObject_jhybridobject> asObject(alias_ref<JRuntimeThreadContext>, alias_ref<JJSIRuntime::jhybridobject> jRuntime);
     std::string toString(alias_ref<JRuntimeThreadContext>, alias_ref<JJSIRuntime::jhybridobject> jRuntime);
 
+    ~JJSIValue() override {
+        release();
+    }
```
sugarmanz, line 157: Is this still necessary?

jvm/hermes/src/main/jni/JJSIValue.h
```diff
@@ -148,21 +152,37 @@ class JJSIValue : public JJSIValueHybridClass {
...
+
     void release() override {
-        value_.reset();
+        if (!tracked && value_) value_.reset();
+        if (scope_) scope_->clearRef(this);
```
sugarmanz, line 161: _Ah_, I get why this is necessary - but maybe the `!tracked && value_` case could just make it's own `RuntimeScope` so that all the APIs still go through a scope and you don't need explicit logic to handle both cases.

jvm/hermes/src/main/jni/JJSIValue.h
```diff
@@ -148,21 +152,37 @@ class JJSIValue : public JJSIValueHybridClass {
...
     Value& get_value() const {
-        if (value_) return *value_;
+        if (!tracked && value_) {
+            return *value_;
+        }
+        if (scope_) {
+            if (auto ref = scope_->getValue((void *)this)) {
+                return *ref;
+            } else {
+                throw std::runtime_error("Value reference cleaned up and is no longer in RuntimeScope");
```
sugarmanz, line 176: 1. This should probably just go through `throwNativeHandleReleasedException("Value");` from below
2. If you don't go through the helper function, you need to tell the JNI to expect an exception, like [`throwNativeHandleReleasedException` does](https://github.com/player-ui/player/blob/dc3c3ab6fbc032b6dc44bdda553cfb5b616d93bf/jvm/hermes/src/main/jni/JJSIValue.cpp#L24-L25) (and should also probably wrap in a `PlayerException` - like `throwNativeHandleReleasedException` does)

jvm/hermes/src/main/jni/RuntimeScope.cpp
```diff
@@ -0,0 +1,95 @@
...
+void RuntimeScope::trackSymbol(void *ptr, facebook::jsi::Symbol value) {
+    symbolScope->insert({ptr, make_unique<Symbol>(std::move(value))});
+}
+
+Value* RuntimeScope::getValue(void* ptr) {
+    auto it = valueScope->find(ptr);
+    if (it != valueScope->end()) {
+        return valueScope->at(ptr).get();
+    }
+    throw std::runtime_error("Value was not found in the RuntimeScope");
```
sugarmanz, line 34: Why throw here? The return type is nullable, it should just return `nullptr`.

Otherwise, change return type to `Value&` to be more accurate (although, my preference would be keep nullable return type to match standard map APIs and let the caller handle `nullptr` explicitly.

brocollie08: this would apply to all the pointers, right?

PR #566, opened 2024-12-23
Only mutate ManagedPlayerViewModel.loadingState on main thread

ap-for-work:
Published properties in SwiftUI models should not be mutated on background threads.

Refactor `ManagedPlayerViewModel.next` to ensure all paths that update `loadingState` do so from the main thread.

Add unit tests to verify loadingState changes arrive on the main thread when next is called.

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs


nancywu1 (approved):
looks good to me, thanks for adding the tests!

hborawski (approved):

PR #571, opened 2025-01-10
[WIP] JVM and Android Coverage

KetanReddy:
Collect coverage for JVM and Android test targets

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

TODO:
- [ ] Update getting started docs for JDK 17 requirement
- [ ] Collect coverage for `//jvm/hermes/..` targets

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,4 +1,4 @@
-common --enable_bzlmod
+common --enable_bzlmod --nolegacy_important_outputs --experimental_remote_build_event_upload=minimal
 test --test_output=errors
 coverage --combined_report=lcov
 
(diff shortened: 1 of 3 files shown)


KetanReddy:
So it looks like enabling ABI Jars is required to make coverage work for targets that use inline functions (which is a [known issue](https://github.com/bazelbuild/rules_kotlin/issues/509) in rules kotlin) but for some reason that causes the jvm core and hermes tests to fail. It seems like the `kt_abi_plugin_incompatible` tag isn't being propagated correctly or it needs to be specified on the source packages.

PR #575, opened 2025-01-16
 Example asset and transform for async content streaming

cehan-Chloe:
### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

# Release Notes
Player streaming enhancement
1. Changes
- Added new `chat-message` asset in `ReferenceAssetsPlugin` as async streaming example 
- Added generic function in `AsyncNodePlugin` to generate async asset transform
2. New functionality
Instead of adding `{async: true}` in content, replace it with a special type of asset, eg. `chat-message` associated with a transform function to add asyncNode


core/player/src/view/builder/index.test.ts
```diff
@@ -34,6 +34,26 @@ describe("multiNode", () => {
...
+  expect(result2.id).toBe("2");
+  expect(result2.flatten).toBe(true);
+});
+
+test("asset wrapper", () => {
+  const result = Builder.assetWrapper({ id: "asset", type: "text" });
+
+  expect(result.type).toBe(NodeType.Value);
+  expect(result.children?.[0]?.value.type).toBe("asset");
+  expect(result.children?.[0]?.value.value.id).toBe("asset");
```
brocollie08, line 54: nit: change the id so failure is more clear

core/player/src/view/builder/index.ts
```diff
@@ -36,7 +43,7 @@ export class Builder {
    * @param values - the value or applicability nodes to put in the multinode
```
brocollie08, line 43: doc update

core/player/src/view/builder/index.ts
```diff
@@ -36,7 +43,7 @@ export class Builder {
...
    */
   static multiNode(
-    ...values: (Node.Value | Node.Applicability)[]
+    ...values: (Node.Value | Node.Applicability | Node.Asset | Node.Async)[]
```
brocollie08, line 46: is this builder tested

core/player/src/view/parser/types.ts
```diff
@@ -116,6 +116,10 @@ export declare namespace Node {
     id: string;
     /** The value representing the node */
     value: Node;
+    /**
+     * Should this list be flattened in resolver
```
brocollie08, line 120: don't think async always guaranteed to be a list

docs/storybook/src/reference-assets/ChatMessage.mdx
```diff
@@ -0,0 +1,9 @@
+import * as ChatMessageStories from "./ChatMessage.stories";
+
+# ChatMessage Asset
+
+The `ChatMessage` asset is used as an example of async node.
```
brocollie08, line 5: semantics, the `ChatMessage` itself isn't an async node, but it's an example of an asset that supports further async loading of future assets

core/player/src/view/resolver/index.ts
```diff
@@ -408,7 +411,27 @@ export class Resolver {
         );
 
         if (mTree.value !== undefined && mTree.value !== null) {
-          childValue.push(mTree.value);
+          if (
```
brocollie08, line 414: maybe add some code comments for this section so whoever looks at this in the future knows what this is (discussion on where the flatten is, what it means for parent/children) 

there are a lot of checks here, maybe clarify what each case is

plugins/reference-assets/core/src/assets/chat-message/__tests__/transform.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    const ref = runTransform(
+      "chat-message",
+      chatMessageTransform,
+      {
+        type: "chat-message",
+        id: "1",
+        value: "Hello World!",
+      },
+      [
+        new AsyncNodePlugin({
```
brocollie08, line 20: is this plugin needed to run just the transform?

plugins/reference-assets/core/src/assets/chat-message/__tests__/transform.test.ts
```diff
@@ -0,0 +1,40 @@
...
+          plugins: [new AsyncNodePluginPlugin()],
+        }),
+      ],
+    );
+
+    expect(ref.current).toMatchObject({
+      id: expect.any(String),
+      type: "collection",
+      values: [
```
brocollie08, line 29: it's probably not possible to verify the async node placeholder here, but is there a way to check for that to exist after the transform

tools/asset-testing-library/core/src/index.ts
```diff
@@ -17,7 +17,7 @@ export function runTransform<
   TransformedAssetType extends Asset = BaseAssetType,
 >(
   type: string,
-  transform: TransformFunction<BaseAssetType, TransformedAssetType>,
+  transform: TransformFunctions,
```
brocollie08, line 20: i might be missing something, what's this change

plugins/async-node/core/src/__tests__/index.test.ts
```diff
@@ -548,6 +549,285 @@ describe("view", () => {
...
+    plugin.hooks.onAsyncNode.tap("test", async (node) => {
+      return new Promise((resolve) => {
+        deferredResolve = resolve;
+      });
+    });
+
+    let updateNumber = 0;
+
+    const player = new Player({
+      plugins: [plugin, new ReferenceAssetsPlugin()],
```
brocollie08, line 569: nit: just group this together with the async one, pass in an array

plugins/async-node/core/src/__tests__/index.test.ts
```diff
@@ -548,6 +549,285 @@ describe("view", () => {
...
+        },
+      },
+    };
+
+    player.start(chatMessageContent as any);
+
+    let view = (player.getState() as InProgressState).controllers.view
+      .currentView?.lastUpdate;
+
+    expect(view).toBeDefined();
```
brocollie08, line 613: hmmm where's the first textview

plugins/async-node/core/src/__tests__/index.test.ts
```diff
@@ -548,6 +549,285 @@ describe("view", () => {
...
```
brocollie08, line 712: same here, how are we suddenly jumping from nothing to 2 texts

plugins/async-node/core/src/__tests__/index.test.ts
```diff
@@ -548,6 +549,285 @@ describe("view", () => {
...
+    });
+
+    view = (player.getState() as InProgressState).controllers.view.currentView
+      ?.lastUpdate;
+
+    expect(view?.values[0].asset.type).toBe("text");
+    expect(view?.values[0].asset.value).toBe("Hello World!");
+    expect(view?.values[1].asset.type).toBe("text");
+    expect(view?.values[1].asset.value).toBe("async content");
+  });
```
brocollie08, line 830: are you able to add a test that builds on top of this, that resolves the second chat-message async content as well

core/player/src/view/parser/types.ts
cehan-Chloe: How about changing it to `Should the new content streamed in be flattened during resolving`?

core/player/src/view/resolver/index.ts
cehan-Chloe: i'm looking at the first check, which is the case that consumer streams in multi-node, but with the change in this pr, can they still stream in multi-node or should just limit to special async asset? do you think we'd better still keep the functionality or remove it to make this resolver part more clear?
```
  mValue.type === NodeType.Async &&
  mValue.flatten &&
  Array.isArray(mTree.value)
```

plugins/reference-assets/core/src/assets/chat-message/__tests__/transform.test.ts
cehan-Chloe: > it's probably not possible to verify the async node placeholder here

Async node placeholder is not going to be part of final currentView but in the resolver.ASTMap. [resolver](https://github.com/player-ui/player/blob/async-streaming-core/core/player/src/view/view.ts#L85) is a private member of view instance. we can have a test getter for it.

> is there a way to check for that to exist after the transform

`runTransform` func starts the player instance so we can't really check the beforeResolve result.

plugins/reference-assets/core/src/assets/chat-message/__tests__/transform.test.ts
cehan-Chloe: not needed. removed

tools/asset-testing-library/core/src/index.ts
cehan-Chloe: before, this util was only testing `TransformFunction` which is afterResolve. but in chat-message we test beforeResolve as well and it's the same way to register it in `AssetTransformPlugin` so i changed to `TransformFunctions ` [interface](https://github.com/player-ui/player/blob/0359f49cf83991b9c7cf8c4dc31c3f50b9c4055a/core/player/src/controllers/view/types.ts#L24) for both before and after resolve

plugins/async-node/core/src/__tests__/index.test.ts
cehan-Chloe: here checked the length of view values is 1 which is the first textview. line 647 tested the type. i can move it here to make it more clear.

core/player/src/view/resolver/index.ts
brocollie08: streaming in a multi-node is fine, gives them the flexibility

plugins/async-node/core/src/transform.ts
```diff
@@ -0,0 +1,28 @@
+import { Builder } from "@player-ui/player";
+import type { AsyncTransformFunc } from "./types";
+
+export const asyncTransform: AsyncTransformFunc = (
```
KetanReddy, line 4: Can we add a comment here (or somewhere) around how to use this?

KetanReddy: Also can we write a test for this?

plugins/reference-assets/mocks/chat-message/chat-message-basic.json
```diff
@@ -0,0 +1,27 @@
+{
```
KetanReddy, line 1: Our mocks should be using the DSL

KetanReddy:
Can we add some docs to the main site about how to author async assets?

plugins/reference-assets/mocks/chat-message/chat-message-basic.json
cehan-Chloe: Just to confirm, should it be DSL only or keep both json and DSL? i'm seeing action and flow-manager have dsl and json

cehan-Chloe:
> Can we add some docs to the main site about how to author async assets?

@KetanReddy ty for reviewing! i'm planning to have a separate pr for doc update including core, react, iOS and Android. as this pr only contains change for core and react. does it sound good to you?

KetanReddy:
> > Can we add some docs to the main site about how to author async assets?
> 
> @KetanReddy ty for reviewing! i'm planning to have a separate pr for doc update including core, react, iOS and Android. this pr only contains change for core and react.

What is the work remaining on the other platforms? I see there is an iOS PR open but from my understanding this should mostly be a core PR.

KetanReddy (dismissed):

plugins/reference-assets/core/src/assets/chat-message/__tests__/transform.test.ts
```diff
@@ -0,0 +1,61 @@
...
+          asset: {
+            id: "1",
+            type: "text",
+            value: "chat message",
+          },
+        },
+      ],
+    });
+
+    expect(test).toHaveBeenCalled();
```
brocollie08, line 59: 🎉

brocollie08 (approved):

brocollie08 (approved):

PR #578, opened 2025-01-22
Check scope active for awaiting player

brocollie08:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.10.2--canary.578.19394</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.10.2--canary.578.19394
</details>

### DIFF
--- a/android/player/src/main/java/com/intuit/playerui/android/lifecycle/PlayerViewModel.kt
+++ b/android/player/src/main/java/com/intuit/playerui/android/lifecycle/PlayerViewModel.kt
@@ -80,7 +80,7 @@ public open class PlayerViewModel(flows: AsyncFlowIterator) : ViewModel(), Andro
             deferredPlayer.getCompleted()
         } else {
             runBlocking {
-                if (viewModelScope.isActive) deferredPlayer.await() else throw IllegalAccessError("Accessing Player instance when ViewModelScope is no longer active")
+                deferredPlayer.await()
             }
         }
     }
@@ -162,11 +162,7 @@ public open class PlayerViewModel(flows: AsyncFlowIterator) : ViewModel(), Andro
             }
         }
 
-        try {
-            release()
-        } catch (e: IllegalAccessError) {
-            Log.e("AndroidPlayer", e.message ?: "IllegalAccessError in AndroidPlayer")
-        }
+        release()
     }
 
     public fun recycle() {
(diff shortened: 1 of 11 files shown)


android/player/src/main/java/com/intuit/playerui/android/lifecycle/PlayerViewModel.kt
```diff
@@ -78,7 +80,7 @@ public open class PlayerViewModel(flows: AsyncFlowIterator) : ViewModel(), Andro
             deferredPlayer.getCompleted()
         } else {
             runBlocking {
-                deferredPlayer.await()
+                if (viewModelScope.isActive) deferredPlayer.await() else throw IllegalAccessError("Accessing Player instance when ViewModelScope is no longer active")
```
sugarmanz, line 83: 1. Can we do the check outside the `runBlocking` to avoid the perf overhead of creating coroutines and blocking the thread on that coroutine context?

2. If the scope that `deferredPlayer` executes in changes, this check would need to be updated to. It'd probably be better to do `deferredPlayer.isActive` if that's possible to check "activeness" explicitly against the `Deferred` we're trying to access.

android/player/src/main/java/com/intuit/playerui/android/lifecycle/PlayerViewModel.kt
```diff
@@ -160,7 +162,11 @@ public open class PlayerViewModel(flows: AsyncFlowIterator) : ViewModel(), Andro
             }
         }
 
-        release()
+        try {
+            release()
+        } catch (e: IllegalAccessError) {
+            Log.e("AndroidPlayer", e.message ?: "IllegalAccessError in AndroidPlayer")
+        }
```
sugarmanz, line 169: Do you think this might swallow real `IllegalAccessError`s? Can/should we be more targeted with what we're wrapping in try/catch, or maybe even not throw anything at all? i.e.:
```kotlin
    public fun release() {
        if (deferredPlayer.isCompleted) {
            player.release()
        }
    }
```

android/player/src/main/java/com/intuit/playerui/android/lifecycle/PlayerViewModel.kt
brocollie08: moved to release so no longer needed to check and return or throw here

PR #579, opened 2025-01-23
[WIP] Initial Pass at Runtime Flag Architecture 

KetanReddy:
# Background

Player strives to be as configurable as possible and primarily achieves this through it's plugin architecture. While this does allow almost any aspect of Player's behavior to be configured, there are still some parts of Player's internal working that users of Player don't have access to. These are things like how strict Player is while running and how it should recover from state errors. For the most part the Player team has made assumptions on how users of Player would want Player to work but in pursuit of fulfilling Player's goal of customizability and configurability we should allow this behavior to be customizable. 

# Implementation Details 
[WIP]

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

## Release Notes
Add the ability to statically/dynamically configure Player's runtime behavior via "flags". Flags are held in a `FlagController` class that can either get flag configuration via Player's initial configuration, dynamically at runtime via a function to set new flags, or even more dynamically by tapping a hook on the controller to override the value of any specific flag(s).

### DIFF
--- a/core/player/src/controllers/flags/types.ts
+++ b/core/player/src/controllers/flags/types.ts
@@ -1,0 +1,17 @@
+import { Severity } from "../../logger";
+
+/** Configuration for runtime Player behavior */
+export interface PlayerFlags {
+  /** What log level duplicate ID errors during view resolution should be raised as */
+  duplicateIDLogLevel: Severity;
+
+  /** Log level for when there are cache conflicts in view resolution (usually related to duplicate IDs) */
+  cacheConflictLogLevel: Severity;
+}
+
+export const DefaultFlags: PlayerFlags = {
+  duplicateIDLogLevel: "error",
+  cacheConflictLogLevel: "info",
+};
+
+export type Flag = keyof PlayerFlags;
--- a/core/player/src/view/resolver/index.ts
+++ b/core/player/src/view/resolver/index.ts
@@ -9,12 +9,13 @@ import type {
   Updates,
 } from "../../data";
 import { DependencyModel, withParser } from "../../data";
-import type { Logger } from "../../logger";
+import type { Logger, Severity } from "../../logger";
 import type { Node } from "../parser";
 import { NodeType } from "../parser";
 import { caresAboutDataChanges, toNodeResolveOptions } from "./utils";
 import type { Resolve } from "./types";
 import { getNodeID } from "../parser/utils";
+import { FlagController } from "../../controllers";
 
 export * from "./types";
 export * from "./utils";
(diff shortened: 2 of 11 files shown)


core/player/src/controllers/flags/types.ts
```diff
@@ -0,0 +1,17 @@
+import { Severity } from "../../logger";
+
+/** Configuration for runtime Player behavior */
+export interface PlayerFlags {
+  /** What log level duplicate ID errors during view resolution should be raised as */
+  duplicateIDLogLevel: Severity;
+
+  /** Log level for when there are cache conflicts in view resolution (usually related to duplicate IDs) */
+  cacheConflictLogLevel: Severity;
+}
```
KetanReddy, line 10: Open to feedback on the flag format. For simplicity this initial pass was built to use flat flags primarily because nested flags would take more logic to merge and operate on and pulling in another dependency would be preferable to writing that logic custom but something like that would add a few kb to Player's size

core/player/src/view/resolver/index.ts
```diff
@@ -167,10 +212,28 @@ export class Resolver {
...
+      message = `Cache conflict: Found Asset/View nodes that have conflicting ids: ${id}, may cause cache issues.`;
+    } else if (type === NodeType.Value) {
+      loglevel = this.options.flagController.getFlag("cacheConflictLogLevel");
+      message = `Cache conflict: Found Value nodes that have conflicting ids: ${id}, may cause cache issues. To improve performance make value node IDs globally unique.`;
+    }
+
+    if (message && loglevel) {
+      this.logger?.[loglevel]?.(message);
+    }
+  }
```
KetanReddy, line 235: While most usages of flags I can think of right now are around logging but in theory some runtime functionality could depend on flags (like errors during expression parsing). Based on the current model a reference to the `FlagController` would have to be passed around but we could expose some dynamic logging functionality in the logger to take a flag and a message and let it internally figure out the severity to log with which would remove the need to pass around the reference as much.

core/player/src/controllers/flags/types.ts
adierkens: We have the `LocalDataStore` and the binding format we could use for nesting. Then do something like: 
```ts
logs.level.duplicateId
```

KetanReddy: I was considering using that but it seemed like overkill complexity-wise and potentially adding some overhead to get a flag if we need to go through binding parsing (even if its cached). I can write some benchmarks for it though to see what the impact would be.

PR #584, opened 2025-02-03
Update  to explicitly call out the default value for dataPath

A1shK:
The docs used the word "enable" that made us think that if a `dataPath` isnt provided, the constants will not be overridable through content. However, looking at the [code](https://github.com/player-ui/player/blob/7e2c1a691ea8e0aae1f2c412d145dcb4749d7164/plugins/shared-constants/core/src/index.ts#L45) thats not true.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

## Release Notes

Updated docs for Shared Constants Plugin to highlight the default dataPath used for overriding values

### DIFF
--- a/docs/site/src/content/docs/plugins/multiplatform/shared-constants.mdx
+++ b/docs/site/src/content/docs/plugins/multiplatform/shared-constants.mdx
@@ -23,7 +23,7 @@ const constantsPlugin = new ConstantsPlugin({
     prop2: "B",
   },
   namespace: "constants",
-  dataPath: "data.constants",
+  dataPath: "data.props",
 });
 
 const player = new Player({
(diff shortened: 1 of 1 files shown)


docs/site/src/content/docs/plugins/multiplatform/shared-constants.mdx
```diff
@@ -42,12 +42,12 @@ constantsPlugin.getConstants("prop1"); // 'A'
 
 ### Overriding Values in Content
 
-The `dataPath` configuration option enables content to override specific values for a particular flow:
+The `dataPath` configuration option allows content to override specific values for a particular flow:
```
KetanReddy, line 45:
```suggestion
By default, data can be provided in the `constants` path of the `data` object to override the constants that the plugin was initialized with. This can be overridden though via the `dataPath` configuration option. In the above example, `dataPath` was initialized with the path `data.props` therefore the following code snippet could be used to override the `prop1` constant:
```

KetanReddy (approved):

KetanReddy:
Hey @A1shK, approved the PR in principle. Can you write release notes for though?

KetanReddy (approved):

PR #585, opened 2025-02-04
[WIP] Update FlowManager in Managed Player state if a new one is passed in

KetanReddy:
If a new `FlowManager` instance is passed in while a `MangedPlayer` instance is running, when managed player transitions it will be using the old flow manager instance.  

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

## Release Notes
TBD

### DIFF
--- a/react/player/src/manager/managed-player.tsx
+++ b/react/player/src/manager/managed-player.tsx
@@ -243,7 +243,7 @@ export const usePersistentStateMachine = (options: {
     return unsub;
   }, []);
 
-  return { managedState, state };
+  return { state, managedState };
 };
 
 /**
@@ -252,7 +252,9 @@ export const usePersistentStateMachine = (options: {
  *
  * `suspense` must be enabled to wait for results in flight.
  */
-export const ManagedPlayer = (props: ManagedPlayerProps) => {
+export const ManagedPlayer = (
+  props: ManagedPlayerProps,
+): React.ReactElement | null => {
   const { withRequestTime, RequestTimeMetricsPlugin } = useRequestTime();
 
   const { state, managedState } = usePersistentStateMachine({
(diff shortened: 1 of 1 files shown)


react/player/src/manager/managed-player.tsx
```diff
@@ -284,6 +286,12 @@ export const ManagedPlayer = (props: ManagedPlayerProps) => {
     };
   }, [props.manager, state?.context.reactPlayer.player, state?.value]);
 
+  React.useEffect(() => {
+    if (managedState.state) {
+      managedState.state.context.manager = props.manager;
+    }
+  }, [props.manager]);
```
adierkens, line 293: IMO this feels like we should be terminating the old managed-player before picking up with the new one?

KetanReddy: Yeah this was just me hacking things together to test if this was the source of the issue. In theory a new Flow Manager should only be passed when managed player is in a `completed` state before the `pending` state builds the `next` function called in the `pending` state. If the parent app wants to override the initial Flow Manager it started player with, it should only be able to do so between flows.

react/player/src/manager/managed-player.tsx
```diff
@@ -223,12 +235,24 @@ export const usePersistentStateMachine = (options: {
...
-  const managedState =
-    managedPlayerStateMachines.get(keyRef.current) ??
-    new ManagedState({ middleware: options.middleware });
-  managedPlayerStateMachines.set(keyRef.current, managedState);
+  const initialManagedState = getOrCreateNewManagedState(
+    keyRef.current,
+    options.middleware,
+  );
+
+  const [managedState, setManagedState] = React.useState(initialManagedState);
```
adierkens, line 243: Do we need this kept in state, or could it be a `ref` instead?

KetanReddy: Since it should trigger a re-render when we create a new managed state I figured it should be kept in state. I guess technically because things will re-render anyways when the new flow comes in we don't need to do that twice. I can move it over to a ref.

PR #586, opened 2025-02-07
Moved AssetTest and Assertions into an android testutils package

mrigankmg:
### Change Type (required)
Moved AssetTest and Aseertions into an android testutils package

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


android/testutils/src/main/java/com/intuit/playerui/android/testutils/asset/AssetTest.kt
```diff
@@ -46,16 +46,16 @@ import org.robolectric.annotation.Config
 @RunWith(AndroidJUnit4::class)
 @Config(sdk = [28])
 @OptIn(ExperimentalCoroutinesApi::class)
-abstract class AssetTest(val group: String? = null) {
+public abstract class AssetTest(private val group: String? = null) {
 
     @get:Rule
-    val name = TestName()
+    public val name: TestName = TestName()
```
brocollie08, line 52: I think it'd be better to have these as `protected` so only child classes can access them

mrigankmg: Made most properties `protected` except for `name`, and `beforeEach` and `afterEach` funcs because they having them as `protected` leads to a `RuntimeExceptionError`.

android/testutils/BUILD
```diff
@@ -0,0 +1,36 @@
...
+kt_android_library(
+    name = "testutils_android",
+    srcs = glob(["src/main/java/**/*.kt"]),
+    custom_package = "com.intuit.playerui.android.testutils",
+    manifest = ":src/main/AndroidManifest.xml",
+    resource_files = glob(["src/main/res/**"]),
+    deps = main_deps,
+)
+
+android_library(
```
brocollie08, line 16: is this wrap necessary

brocollie08 (approved):

android/testutils/BUILD
mrigankmg: I think I need it for the specific maven coordinates?

PR #588, opened 2025-02-10
Making Node.snapshot `public`

A1shK:
This PR marks `Node.snapshot` as `public`. Since it was already displayed marked as an experimental API, I decided not to add any docs around it as I am not sure if its intended to be advertised.  

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

## Release Notes
Making `Node.Snapshot`  public. This utility can be used to decode a `Node` entirely as a snapshot of its current data.


KetanReddy:
In my opinion if it’s going to be public there should be docs around it, even if the docs mention it’s a potential unstable API just to make sure that’s disclosed. I’ll defer to @sugarmanz or @brocollie08 though

A1shK:
Closing because I dont think we need this right now.

PR #591, opened 2025-02-13
Expose metrics plugin hooks android

mrigankmg:
### Change Type (required)
Expose onFlowBegin, onFlowEnd, onRenderEnd metrics plugin hooks for Android

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/plugins/metrics/jvm/src/main/kotlin/com/intuit/playerui/plugins/metrics/MetricsPlugin.kt
+++ b/plugins/metrics/jvm/src/main/kotlin/com/intuit/playerui/plugins/metrics/MetricsPlugin.kt
@@ -1,20 +1,30 @@
 package com.intuit.playerui.plugins.metrics
 
 import com.intuit.playerui.core.bridge.Node
+import com.intuit.playerui.core.bridge.NodeWrapper
 import com.intuit.playerui.core.bridge.getInvokable
+import com.intuit.playerui.core.bridge.hooks.NodeSyncHook1
+import com.intuit.playerui.core.bridge.hooks.NodeSyncHook3
 import com.intuit.playerui.core.bridge.runtime.Runtime
 import com.intuit.playerui.core.bridge.runtime.ScriptContext
 import com.intuit.playerui.core.bridge.runtime.add
+import com.intuit.playerui.core.bridge.serialization.serializers.NodeSerializableField
+import com.intuit.playerui.core.bridge.serialization.serializers.NodeWrapperSerializer
 import com.intuit.playerui.core.player.Player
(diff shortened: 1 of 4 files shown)


plugins/metrics/jvm/src/test/kotlin/com/intuit/playerui/plugins/metrics/MetricsPluginTest.kt
```diff
@@ -45,6 +48,48 @@ internal class MetricsPluginTest : PlayerTest() {
...
+        val result = flow.await()
+
+        assertEquals("DONE", result.endState.outcome)
+        assertTrue(onFlowEndTapped)
+    }
+
+    @TestTemplate
+    fun `should trigger onRenderEnd hook`() {
+        var onRenderEndTapped = false
+        plugin?.hooks?.onRenderEnd?.tap("test") { _, _, _ ->
```
nancywu1, line 83: do we want to assert the 3 values from the handler or assert not null?

mrigankmg: I think asserting on the type should be good enough, since the value assertion should be handled in core?

plugins/metrics/jvm/src/main/kotlin/com/intuit/playerui/plugins/metrics/MetricsPlugin.kt
```diff
@@ -32,6 +42,22 @@ public class MetricsPlugin(
...
+
+    @Serializable(Hooks.Serializer::class)
+    public class Hooks internal constructor(override val node: Node) : NodeWrapper {
+        public val onFlowBegin: NodeSyncHook1<PlayerFlowMetrics>
+            by NodeSerializableField(NodeSyncHook1.serializer(PlayerFlowMetrics.serializer()))
+
+        public val onFlowEnd: NodeSyncHook1<PlayerFlowMetrics>
+            by NodeSerializableField(NodeSyncHook1.serializer(PlayerFlowMetrics.serializer()))
+
+        public val onRenderEnd: NodeSyncHook3<Timing, RenderMetrics, PlayerFlowMetrics>
```
brocollie08, line 57: I think `RenderMetrics` is missing `updates`

mrigankmg: Added!

PR #594, opened 2025-02-14
Convert Flow to NodeWrapper

mrigankmg:
### Change Type (required)
Convert Flow to NodeWrapper

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


jvm/core/src/main/kotlin/com/intuit/playerui/core/flow/Flow.kt
```diff
@@ -1,20 +1,72 @@
...
+                override val values: Collection<Any?> = paramsMap.values
+                override fun containsKey(key: String): Boolean = paramsMap.containsKey(key)
+                override fun containsValue(value: Any?): Boolean = paramsMap.containsValue(value)
+                override fun get(key: String): Any? = paramsMap[key]
+                override fun isEmpty(): Boolean = paramsMap.isEmpty()
+                override fun isReleased(): Boolean = false
+                override fun isUndefined(): Boolean = false
+                override fun nativeReferenceEquals(other: Any?): Boolean = other is Node && this.entries == other.entries
+                override val runtime: Runtime<*> get() = throw UnsupportedOperationException()
+            })
```
sugarmanz, line 69: There is a `Map` backed `Node` here:
https://github.com/player-ui/player/blob/d2cd18120b8527d2b84fb56c0442a0a183f9b7dc/jvm/testutils/src/main/kotlin/com/intuit/playerui/testutils/Node.kt#L19

mrigankmg: That's awesome! Would it be of concern to use a testutil in a non test class though?

brocollie08: I'm leaning towards moving away from the usage of a primary constructor. It was a container class for data in the first place, not like you can create a flow to start Player with. 

I'd say make it a node backed object so custom properties can be retrieved through the `Node` API, and add to the companion object that creates a default flow object through deserializing a json or something like that

jvm/testutils/src/main/kotlin/com/intuit/playerui/utils/test/TestMocks.kt
```diff
@@ -20,3 +23,12 @@ public val simpleFlow: Flow by lazy {
     Json { ignoreUnknownKeys = true }
         .decodeFromString(Flow.serializer(), simpleFlowString)
 }
+
+@Serializable
+public data class Flow(
```
brocollie08, line 28: forgot, mark this internal

brocollie08 (dismissed):

brocollie08 (approved):

PR #600, opened 2025-02-21
Async streaming android

sakuntala-motukuri:
### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

Release Notes
Player streaming enhancement for android

Changes

Update async-node-plugin and reference-assets-plugin tests

image

 https://app.buildbuddy.io/invocation/57b17939-c5b5-4bbe-84e0-360d0f1b8043


plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -387,4 +463,183 @@ internal class AsyncNodePluginTest : PlayerTest() {
...
+    }
+
+
+    @TestTemplate
+    fun `chat-message asset - resolve single chat-message`() = runBlockingTest {
+
+        var update: Asset? = null
+        plugin?.hooks?.onAsyncNode?.tap("") { _, node, callback ->
+            BailResult.Bail(
+                listOf(
```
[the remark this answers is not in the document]
sakuntala-motukuri: So I think this wouldn't be considered as a list unless we pass multiple arrays , so I took a reference of this where they pass in asset of type "text" https://github.com/player-ui/player/blob/2d8521e90a937cfd90108c87f0069a8bb516c21f/plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt#L119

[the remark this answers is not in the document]
cehan-Chloe: Multi-node and async asset flatten logic are different. 

The test you refer to means you can stream in a list with one text asset. it will be a multi-node as well and get flatten. i saw it already covered in this pr by the `chat-message asset replaces async nodes with multi node flattened`. 

The content streamed in here is using `listOf`, so it will be pass in as `[chat-message-asset]` then considered as multinode. 

You can compare result by removing the `listof` and only streaming in the asset itself. 

[Ref](https://github.com/player-ui/player/pull/575/files#diff-9979d7cc1989ec2431a108b837ae516132f687e6824a9b0aadbf44d28d9e1109R414-R436): flatten logic in resolver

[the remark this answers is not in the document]
sakuntala-motukuri: Ok so the problem here is the one your suggesting me is to pass a single object of **

> [chat-message-asset]

** but unfortunately , the deferred node is of type List<Map<String, Any?>>? as in [here](https://github.com/player-ui/player/blob/2d8521e90a937cfd90108c87f0069a8bb516c21f/plugins/async-node/jvm/src/main/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePlugin.kt#L24) , and also if I go ahead and change the deferredResolve to Pair<String, Map<String, String>> , that would impact other test cases which wouldn't be supported as their data type is currently List<Map<String, Any?>>? as per asyncNodeUpdate , I will need to check what else can be done to not consider this as ListOf and I highly doubt if this is the actual root cause causing this issue. Will keep you posted.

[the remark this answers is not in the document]
sakuntala-motukuri: If changing the deferred node is a solution , then that would impact the BailResult hook that is being used here as it accepts only only a list of maps in the current implementation [here](https://github.intuit.com/player/hooks/blob/a3c69074f4e2304e207e6e1e9de2dcf469be81b7/hooks/src/main/kotlin/com/intuit/hooks/SyncBailHook.kt#L3) and expose it as a Map | List in player hooks and I would need to go ahead and change the type of asyncnodeupdate to accept Map along with a list [here](https://github.com/player-ui/player/blob/2d8521e90a937cfd90108c87f0069a8bb516c21f/plugins/async-node/jvm/src/main/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePlugin.kt#L24) , I need to see if there is any hook in android that supports this kind of behaviour.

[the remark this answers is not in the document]
cehan-Chloe: could you help me to understand why BailResult hook only accept a list of maps? and my confusion is why Android has a different behaviour with other platforms and core and is there a way to change it. 🤔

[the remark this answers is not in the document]
sakuntala-motukuri: Ok , so if you look at my test [here](https://github.com/player-ui/player/pull/600/files#diff-ac6259632cb11fb85e6116a391926ea88917bc5f23faba6c0657f014a09d74c0R533) the result of passing an asset [here](https://github.com/player-ui/player/pull/600/files#diff-ac6259632cb11fb85e6116a391926ea88917bc5f23faba6c0657f014a09d74c0R534) resolves to a List<Map<String, Any?>>?. which is my deferredResolve but in this use-case as per the core if we need to make the deferredResult to accept both List and Map as such we would either need to expose BailResult hook which uses [SyncBailHook](https://github.intuit.com/player/hooks/blob/a3c69074f4e2304e207e6e1e9de2dcf469be81b7/hooks/src/main/kotlin/com/intuit/hooks/SyncBailHook.kt#L9) that accepts only a List in current functionality to accept either List or a Map implementation. The **BailResult** hook currently accepts only a list of maps because the type asyncNodeUpdate is defined as List<Map<String, Any?>>?. To allow it to accept both a List and a Map, we need to change the typealias of asyncNodeUpdate to Any? and follow up changes to the AsynNodePlugin.kt and NodeAsyncParallelBailHook2.kt and update the BailResult hook accordingly. 

For this 

> Android has a different behaviour with other platforms and core and is there a way to change it.

I'm. not sure why the behaviour is not consistent across platforms , probably worth discussing it with @brocollie08

plugins/async-node/jvm/src/main/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePlugin.kt
```diff
@@ -16,12 +16,9 @@ import com.intuit.playerui.core.plugins.JSScriptPluginWrapper
 import com.intuit.playerui.core.plugins.findPlugin
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.builtins.ListSerializer
-import kotlinx.serialization.builtins.MapSerializer
-import kotlinx.serialization.builtins.nullable
 import kotlinx.serialization.builtins.serializer
 
-public typealias asyncNodeUpdate = List<Map<String, Any?>>?
+public typealias asyncNodeUpdate = Any?
```
brocollie08, line 21: is there a reason why this is changed to `Any`?

brocollie08: maybe a bit more context is needed here. This was initially made to be a list of maps to represent a list of assets (represented by maps), forming a multinode, which then gets flattened by core logic.

you may need to work with @cehan-Chloe on that, maybe this is why we weren't seeing the flattening being applied?

sakuntala-motukuri: according to this conversation https://github.com/player-ui/player/pull/600#discussion_r1968220954 and also core defers the return type to any https://github.com/player-ui/player/blob/main/plugins/async-node/core/src/index.ts#L49

sakuntala-motukuri: Yes @brocollie08 , that was the reason why the flattening was not happening

brocollie08: Ok I discussed with Chloe on the old/new rationale, I think `Any` is fine in theory but not so much in practice. Idea behind that is just to give the users some more context clues on what to provide. 

A list of maps representing objects makes sense for users if they had context on the plugin, but `Any` doesn't really give any guard rails or clues on what they're supposed to pass in.

Kotlin doesn't officially support union types but we can sort of mimic it with sealed classes, that way we can guide the user to pass in either a list or a single map instead of `Any`

sakuntala-motukuri: Sure , thank you @brocollie08 , do we have any references which support this behaviour in any of the plugins ? Just for me to get a better idea.

cehan-Chloe: Besides, to avoid the confusion for inconsistent behavior of flattening multi-node streamed in, i'll remove the flatten multi-node logic in core resolver. for this android pr, the test of `chat-message asset replaces async nodes with multi node flattened` will no longer be needed.

brocollie08: if this is how react does it, maybe it's fine for now. I'd add a TODO though and come back to it later

plugins/reference-assets/android/src/androidTest/java/com/intuit/playerui/android/reference/assets/chatmessageasset/ChatMessageAssetTest.kt
```diff
@@ -0,0 +1,28 @@
...
+
+        val collectionValues = currentView?.findViewById<LinearLayout>(R.id.collection_values)
+            ?: throw AssertionError("current view is null")
+
+        collectionValues[0].shouldBeView<TextView> {
+            assertEquals("Hello World!", text.toString())
+        }
+
+        currentState.shouldBePlayerState<InProgressState>()
+    }
```
brocollie08, line 27: I know it probably works from the unit tests, but could you add another test here as well to verify the UI

brocollie08 (approved):
can merge after CI pass

PR #610, opened 2025-03-04
Expose FlowInstance afterTransition hook for jvm

mrigankmg:
### Change Type (required)

Expose FlowInstance afterTransition hook for jvm

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/jvm/core/src/test/kotlin/com/intuit/playerui/core/flow/FlowControllerIntegrationTest.kt
+++ b/jvm/core/src/test/kotlin/com/intuit/playerui/core/flow/FlowControllerIntegrationTest.kt
@@ -7,6 +7,7 @@ import com.intuit.playerui.utils.test.PlayerTest
 import com.intuit.playerui.utils.test.runBlockingTest
 import com.intuit.playerui.utils.test.simpleFlowString
 import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNotNull
 import org.junit.jupiter.api.Assertions.assertNull
 import org.junit.jupiter.api.TestTemplate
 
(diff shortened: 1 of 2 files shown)


jvm/core/src/test/kotlin/com/intuit/playerui/core/flow/FlowControllerIntegrationTest.kt
```diff
@@ -47,6 +55,7 @@ internal class FlowControllerIntegrationTest : PlayerTest() {
         assertEquals("*", pendingTransition?.second)
         assertEquals("VIEW_1", completedTransition?.first?.name)
         assertEquals("END_Done", completedTransition?.second?.name)
+        assertEquals(flowInstanceId, flowInstanceIdAfterTransition)
```
brocollie08, line 58: i'm not really understand what we're verifying here, that the flow id after transition is the same? Is this a meaningful check or is this just to make sure we're getting something back in the hook?

If the latter, can we just verify the existence of data

mrigankmg: Yeah, pretty much just verifying that the current `FlowInstance` is what's being passed into the `afterTransition` call.

nancywu1: we could verify the statetype of the flow instance like in here 
https://github.com/player-ui/player/blob/12c6b5209c5e30e5218723ce13fbe228de5a28ee/plugins/external-action/swiftui/ViewInspector/ExternalActionViewModifierPluginTests.swift#L180

brocollie08 (approved):

Issue #590, opened 2025-02-12
Implement Template API for Explicit Expansion Behavior


KetanReddy:
### Changes to the Template API
Template elements would have a new property (to be decided during implementation) that would take information on whether the expanded data from the template should be appended (added to the end of the any existing elements) or prepended (added before any existing elements) in the location its being expanded to. Because the order of elements in an array is persevered and easy to ascertain at a glance, it is possible for content authors to accurately predict the order of multiple templates being expanded into the same slot with the same expansion behavior (e.g if there are two templates both set to append, the first template will be expanded followed by the second template). 

### Changes to Template Processing
During template expansion in the [TemplatePlugin](https://github.com/player-ui/player/blob/main/core/player/src/view/plugins/template.ts), the newly added field should be used to determine how the expanded template nodes should be inserted to the tree during parsing. In order to not break existing behavior, if no value is specified for how the template should be expanded the existing logic should be followed. As part of this change, we should introduce a warning message that a value was not specified to allow Player to require explicitly setting this at a later date.

Issue #540, opened 2024-11-06
Better Document Template Behavior for Cases When There Are Values Already Present in the Output Property


KetanReddy:
Right now our docs around `Template` usage calls out cases for nested templates and multiple templates but there is a gap for when the template is expanding into a property that already has a value. For example, we should document the behavior of the the following case:

```json
{
  "asset": {
    "id": "top-level",
    "type": "collection",
    "values": [
      {
        "asset": {
          "id": "top-level-value-0",
          "type": "text",
          "value": "Frodo"
        }
      },
    ],
    "template": [
      {
        "data": "list.of.names",
        "output": "values",
        "value": {
          "asset": {
            "id": "value-_index_",
            "type": "text",
            "value": "{{list.of.names._index_}}"
          }
        }
      }
    ]
  }
}
```

PR #614, opened 2025-03-07
Template ordering

kharrop:
Closes https://github.com/player-ui/player/issues/590
Closes https://github.com/player-ui/player/issues/540

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

# Release Notes

Adds support so that content authors can pass in `placement: "append" | "prepend"` to templates so that they can specify the order they should expand in relation to non-template data when everything is transformed into JSON.

This logic works for both static and dynamic templates.

<details>
  <summary>📦 Published PR as canary version: <code>0.11.0--canary.614.21793</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.11.0--canary.614.21793
</details>


core/player/src/view/plugins/template.ts
```diff
@@ -37,7 +39,12 @@ export type TemplateSubstitutionsFunc = (
 export default class TemplatePlugin implements ViewPlugin {
   private readonly options: Options;
 
-  hooks = {
+  hooks: {
+    resolveTemplateSubstitutions: SyncWaterfallHook<
+      [TemplateSubstitution[], TemplateItemInfo],
+      Record<string, any>
+    >;
+  } = {
```
spentacular, line 47: @KetanReddy is there a way we can turn off verbose for these? This kinds sucks to redeclare

kharrop: Yea VS Code was complaining about this; wasn't sure if I needed to do this or not.

KetanReddy: Unfortunately I think that comes from isolated declarations in tsc which we _should_ be using.

core/player/src/view/plugins/__tests__/__snapshots__/asset.test.ts.snap
```diff
@@ -197,6 +197,7 @@ exports[`asset > template 1`] = `
                   },
                 },
               ],
+              Symbol(template): undefined,
```
KetanReddy, line 200: Should we delete the symbol after we use it? In theory I don't think there would be a problem with returning the symbol along with the view but I'm not sure if it would cause issues on mobile? @nancywu1 @brocollie08

core/player/src/view/resolver/index.ts
```diff
@@ -376,6 +376,21 @@ export class Resolver {
...
+              );
+              resolved = setIn(resolved, child.path, arr);
+            } else if (childNode.placement === "prepend") {
+              const arr = addFirst(
+                dlv(resolved, child.path as any[], []),
+                childValue,
+              );
+              resolved = setIn(resolved, child.path, arr);
+            }
+          }
```
KetanReddy, line 393: Do we still need this if we're sorting the dynamic template in the parser?

docs/site/src/content/docs/content/assets-views.mdx
```diff
@@ -259,6 +259,73 @@ Within a template, the `_index_` string can be used to substitute the array-inde
 }
 ```
 
+### Template ordering
+
+When writing template content, the position of the template in relation to non-template content is not guaranteed when it gets transformed into its final JSON output. Because of this, templates support a `placement` property that specifies whether the template should expand before or after pre-existing content.
```
KetanReddy, line 264: Technically the DSL does preserve the ordering, but JSON itself  doesn't guarantee preserving the key ordering.

docs/site/src/content/docs/content/assets-views.mdx
```diff
@@ -259,6 +259,73 @@ Within a template, the `_index_` string can be used to substitute the array-inde
...
+            "type": "text",
+            "value": "{{list.of.names._index_}}"
+          }
+        }
+      }
+    ]
+  }
+}
+```
+
```
KetanReddy, line 328: Even though we support `placement` now, we should document the fallback lexicographical effects on template expansion

KetanReddy: That would also close #540

core/player/src/view/resolver/index.ts
kharrop: TBH I forgot this was a thing - I just reverted these changes and everything still passes 👍

core/player/src/view/plugins/__tests__/__snapshots__/asset.test.ts.snap
kharrop: So this is actually a little trickier than it looks - I at first tried adding `delete node[TemplateSymbol]` after the placement logic but the result would either be only direct children would have Symbols removed and nested children would still contain the Symbol, or they would all be removed correctly but the placement logic would break.

Even when I recursively looked to remove the Symbol, the undefined ones still remained in the snapshot interestingly enough.

Looking more into the snapshots I was receiving, it seemed to only be an issue with undefined for the placement since interestingly enough, the Symbols that did contain placement information wasn't showing up in snapshots to begin with.

As a result, I just ended up adding a conditional to not include the property if placement wasn't defined and that seemed to address the unwanted Symbols appearing within the object.

This makes me wonder if we still do need to ada delete - and even if so, I'm concerned it's not actually removing all the Symbols if the snapshots aren't picking it up, meaning it makes it challenging to test/confirm.

kharrop: Just kidding - I added in a test and it's an issue of the Symbol still being returned. I'll take another stab at the delete logic but seeing as it seems fine in Android/iOS... it could be okay?

kharrop: After more investigation, I tried a bunch of things but every time I removed the symbol (after the sorting logic), the placement tests would fail. I believe it's due to how the symbol is tied to placement itself so removing it at any stage breaks the sorting logic.

I tried multiple things like moving it in various stages - after sorting within the same hook, adding a separate hook, trying it in the beforeResolve hook; either the Symbols would be remove and tests would break, or the Symbols wouldn't be removed.

Since it seems like Symbols are okay for iOS/Android, this seems like it would be okay? I think having it not included if it's undefined is an improvement over all.

core/player/src/view/plugins/template.ts
sugarmanz: I wonder if defining types for `hooks` means you could omit the type parameters from `resolveTemplateSubstitutions: new SyncWaterfallHook()` below

core/player/src/view/plugins/template.ts
```diff
@@ -122,6 +134,52 @@ export default class TemplatePlugin implements ViewPlugin {
...
+                return 0; // Same placement, no preference
+              }
+              // "prepend" should come before "append"
+              return aPlacement === "prepend" ? -1 : 1;
+            }
+            return 0;
+          }
+
+          // Trigger more sorting for nested assets
+          return a.path.join().localeCompare(b.path.join());
```
sugarmanz, line 176: @brocollie08 have we validated this on Android Hermes?

tmarmer: Since we've been seeing issues with `localeCompare` on Android, we could likely simplify this to just use a more direct string comparison like:
```ts
const aPath = a.path.join();
const bPath = b.path.join();

if (aPath === bPath) {
  return 0;
}

return aPath > bPath ? 1 : -1;
```

We don't need the accuracy of `localeCompare` here imo since this isn't sorting for the sake of anything user facing, but instead to make sure equal paths are grouped and ordered correctly compared to each other.

core/player/src/view/plugins/template.ts
```diff
@@ -104,12 +111,17 @@ export default class TemplatePlugin implements ViewPlugin {
       type: NodeType.MultiNode,
       override: false,
       values,
-    };
+    } as Node.MultiNode;
```
KetanReddy, line 114: is there any reason to do `as Node.MultiNode` here if `result` is already typed that way?

core/player/src/view/plugins/template.ts
```diff
@@ -122,6 +134,55 @@ export default class TemplatePlugin implements ViewPlugin {
       return node;
     });
 
+    parser.hooks.onCreateASTNode.tap("template-sort", (node) => {
```
KetanReddy, line 137: The tap name on this hook should probably be the same throughout the plugin

docs/site/src/content/docs/content/assets-views.mdx
```diff
@@ -259,6 +259,77 @@ Within a template, the `_index_` string can be used to substitute the array-inde
...
+**_Example_**
+
+The following content will:
+
+- Place the expanded static template first, since it has `placement: prepend`
+- Then display content from the `values` array
+- Finally place the expanded dynamic template last, since it has `placement: append`
+
+```json
```
KetanReddy, line 288: Should we give a DSL example of this instead of a JSON one?

KetanReddy (approved):

PR #630, opened 2025-04-03
 Renew hydration scope, wrap composableAsset and apply the modifiers to all assets through the compose call

brocollie08:
### Change Type (required)
1. renews hydration scope in the `RenderableAsset.compose` call. This is because in the case where a ComposableAsset is created within another ComposableAsset, we don't end up going through here.
2. Always uses `WRAP_CONTENT` for the `ComposeView` in `ComposableAsset`
3. Exposes [`updateProvidedValues`](https://github.com/player-ui/player/pull/630/files#diff-42db4b43820d4143f6d1be976bd533ba432a4b787fb918387dc4fbf9cbfef140R221) publicly to allow consumers to set styles for Compose assets from parent XML assets

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.11.0--canary.630.22590</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.11.0--canary.630.22590
</details>


android/player/src/main/java/com/intuit/playerui/android/compose/ComposableAsset.kt
```diff
@@ -35,8 +42,14 @@ public abstract class ComposableAsset<Data> (
     assetContext: AssetContext,
     serializer: KSerializer<Data>,
 ) : SuspendableAsset<Data>(assetContext, serializer) {
+    /** Set to false for assets that take full width. Defaults to true */
+    public open fun wrapContent(): Boolean = true
```
brocollie08, line 46: does this need to be a function?

android/player/src/main/java/com/intuit/playerui/android/compose/ComposableAsset.kt
```diff
@@ -35,8 +42,14 @@ public abstract class ComposableAsset<Data> (
...
+
+    public lateinit var composeHydrationScope: CoroutineScope
```
sugarmanz, line 48: Why do we need a different hydration scope?

android/player/src/main/java/com/intuit/playerui/android/compose/ComposableAsset.kt
sugarmanz: Yeah, the distinction to me is somewhat smelly - for Android View based assets, it's done w/ another interface [`ViewportAsset`](https://github.com/player-ui/player/blob/09a2373c7076ec7b9f06a906d7d21243fb066be3/android/player/src/main/java/com/intuit/playerui/android/asset/RenderableAsset.kt#L266-L270), even though that's slightly different, we should follow the same approach if possible.

A1shK: ah yes, I can use that. The only concern would be that we can't make it optional but that's easy to work around

android/player/src/main/java/com/intuit/playerui/android/compose/ComposableAsset.kt
```diff
@@ -92,7 +111,16 @@ private fun RenderableAsset.composeAndroidView(
     modifier: Modifier = Modifier,
     styles: Styles? = null,
 ) {
+    renewHydrationScope("Creating compose view")
+    val scope = rememberCoroutineScope()
     AndroidView(factory = ::FrameLayout, modifier) {
-        render(styles) into it
+        scope.launch {
+            val renderedView = withContext(Dispatchers.IO) {
```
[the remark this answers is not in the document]
A1shK: Seeing some crashing in my project when using `Dispatchers.Default` 
Will continue to investigate with a canary

android/player/src/main/java/com/intuit/playerui/android/compose/ComposableAsset.kt
A1shK: Good question - looked into it and all I need to do is renew the hydration scope 😅 
removed the `composeHydrationScope`

android/player/src/main/java/com/intuit/playerui/android/compose/ComposableAsset.kt
A1shK: Updated to use ViewPortAsset
I realized that this can be mitigated at the asset registry level

plugins/reference-assets/android/src/main/java/com/intuit/playerui/android/reference/assets/action/Action.kt
```diff
@@ -24,7 +23,7 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import kotlinx.serialization.Serializable
 
-class Action(assetContext: AssetContext) : ComposableAsset<Action.Data>(assetContext, Data.serializer()) {
+class Action(assetContext: AssetContext) : ComposableAsset<Action.Data>(assetContext, Data.serializer()), RenderableAsset.ViewportAsset {
```
sugarmanz, line 26: Hmm, I don't think `Action` shouldn't be a `ViewportAsset`. Or at least how that interface was used for Android View assets.

android/player/src/main/java/com/intuit/playerui/android/compose/ComposableAsset.kt
```diff
@@ -42,10 +42,9 @@ public abstract class ComposableAsset<Data> (
 ) : DecodableAsset<Data>(assetContext, serializer) {
 
     override suspend fun initView(data: Data) = ComposeView(requireContext()).apply {
-        layoutParams = if (asset is ViewportAsset) {
-            ViewGroup.LayoutParams(MATCH_PARENT, MATCH_PARENT)
-        } else {
-            ViewGroup.LayoutParams(WRAP_CONTENT, WRAP_CONTENT)
+       layoutParams = ViewGroup.LayoutParams(WRAP_CONTENT, WRAP_CONTENT)
+       setContent {
+            compose(data = data)
```
A1shK, line 47: @brocollie08 why was this removed earlier?

brocollie08: We want to compose with data as soon as view is initialized?

A1shK: hmm i guess not. There are no cases where the hydrate would be skipped on first render right?

nancywu1 (approved):

PR #627, opened 2025-03-28
Update ESLint to v9

KetanReddy:
Update ESLint and related rules to v9. The goal was to maintain broadly similar lint rules with the migration but a few rules have changed since so there were some fixes required. 

### Change Type (required)
Indicate the type of change your pull request is:

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`

### Does your PR have any documentation updates?
- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs
<!--
In an effort to standardize our process and code, please make sure you include documentation and/or update any existing documentation.
Please refer to our site https://player-ui.github.io/latest/about, and include any neccesary information that would be helpful to coders, developers, and learners.

If you are unable to update the current documents, please create an issue for us to get back to it.

## Release Notes
Internal - Update ESLint to v9

PR #632, opened 2025-04-13
Hermes Build Fixes and Upgrade, Better Cache Uses, and ESLint v9 Upgrade

KetanReddy:
This PR does three primary things. For some background this is more or less a cleaned up version of #627 that was opened with the initial goal of updating ESLint to v9 but attempted a _quick_ fix to other Hermes related build issues we were encountering. Attempts to fix the Hermes issues spiraled a bit and for cleanliness so I thought it was best to just start over. 

## Hermes Build Fixes
A few weeks ago all pipelines started failing due to the job being killed in the test phase by Circle. The indication was a socket close error most likely due to resources. After a lot of debugging it was determined that the Hermes build was causing excessive resource consumption which was causing the early termination. While not confirmed, the current theory is taht changes to how CircleCI runs Docker jobs was causing the C++ compilation stages to incorrectly set their own resource utilization limits. Ultimately the fix for this was explicitly set the parallelization via the `CMAKE_BUILD_PARALLEL_LEVEL` flag. 

## Better Cache Uses
Along the way of fixing the Hermes build, it was noticed that the remote cache was not being used as effectively as it could be. The primary culprit was setting `workspace_status_command` for _all_ builds which would cause cache not to be reused across runs in CI nor pull from local runs (when enabled). Additionally because BuildBuddy disabled anonymous access the Android visual test stages weren't using any cache, causing 2x longer builds. Additionally, limiting Bazel logging during build was found to have performance benefits as well.  

## ESLint v9 Upgrade 
Comparatively a minor change and despite it being the original purpose of this work relatively minor. We were on a fairly old version of ESLint so going through the upgrade as a hygiene effort. 

## React Native Hermes Updates

Update RN JSI and Hermes binaries, compiled with `HERMES_UNICODE_LITE` flag to avoid Java `localeCompare` impl

Change log for RN from [here](https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0781) to [here](https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0742) 

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

## Release Notes
Update React Native JSI and Hermes binaries, compiled with `HERMES_UNICODE_LITE` flag to avoid Java `localeCompare` implementation. Change log for React Native from [here](https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0781) to [here](https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0742) 

<details>
  <summary>📦 Published PR as canary version: <code>0.10.5--canary.632.22183</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.10.5--canary.632.22183
</details>


.bazelrc
```diff
@@ -44,19 +54,24 @@ coverage --remote_download_toplevel
...
+# We want to find a compromise between printing often enough to show that the build isn't stuck,
+# but not so often that we produce a long log file that requires a lot of scrolling.
+# https://bazel.build/reference/command-line-reference#flag--show_progress_rate_limit
+common:ci --show_progress_rate_limit=60
+
 # Circle macos.m1.large.gen1 8 cores 12gb ram
 common:ci-mac --config=ci --local_cpu_resources=4 --local_ram_resources=7000
 
-build:release --stamp --workspace_status_command=./scripts/workspace-status.sh
+common:release --config=ci --stamp --workspace_status_command=./scripts/workspace-status.sh
```
sugarmanz, line 74: Amazing

.circleci/config.yml
```diff
@@ -231,11 +232,11 @@ jobs:
       - attach_workspace:
           at: ~/player
 
-      - run: bazel test --config=ci -- //third_party/hermes/hello:test
+      - run: bazel build --config=ci -- $(bazel query 'kind(".*_test", //...) except filter("ios|swiftui", //...)') -//android/demo:android_instrumentation_test
```
sugarmanz, line 235: Hmm, I'm not understanding the change here, why are we building more things at this stage that aren't related to what this specific run step was for (just building `@hermes//:host` in isolation). If that's no longer required, can we just remove this step?

.circleci/config.yml
```diff
@@ -231,11 +232,11 @@ jobs:
...
 
       - run: bazel test --config=ci -- $(bazel query 'kind(".*_test", //...) except filter("ios|swiftui", //...)') -//android/demo:android_instrumentation_test
 
-      - run: bazel test --config=ci --//android/player:runtime=hermes -- $(bazel query 'kind(".*_test", //...) except filter("ios|swiftui", //...)') -//android/demo:android_instrumentation_test
+      - run: bazel test --config=ci --//android/player:runtime=hermes -- $(bazel query 'kind(".*_test", //...) except filter("ios|swiftui|core|react", //...)') -//android/demo:android_instrumentation_test
```
sugarmanz, line 239: Nit: I understand _why_ it makes sense to add the additional filters.. but does it actually matter? Any tests that don't have an input on `//android/player:runtime=hermes` shouldn't be rerun anyways.

third_party/fbjni/BUILD
```diff
@@ -45,6 +45,7 @@ cmake(
         "CMAKE_BUILD_TYPE": "$(CMAKE_BUILD_TYPE)",
         # Resolves to fbjni/jni/include
         "JNI_HEADER": "$(rlocationpath :jni_include)",
+        "CMAKE_BUILD_PARALLEL_LEVEL": "4"
```
sugarmanz, line 48: Is this just a broad strategy for any cmake target? My understanding was this only happened for `@hermes//:host`. If it's a broad strategy, could we pull this out to config set by command line, or set by something in `.bazelrc` to make it easily configurable to the constraints of where it's actually building?

.bazelrc
```diff
@@ -6,6 +6,16 @@ coverage --combined_report=lcov
 build --@rules_kotlin//kotlin/settings:jvm_emit_jdeps=False
 build --cxxopt=-std=c++17
 
+# Java Specific Configs
+build --java_language_version=17 
+build --tool_java_language_version=17 
+build --java_runtime_version=remotejdk_17 
+build --tool_java_runtime_version=remotejdk_17 
+build --repo_env=JAVA_HOME=../bazel_tools/jdk
```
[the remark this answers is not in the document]
KetanReddy: I don't think we can set it in the docker image because these are bazel config flags

WORKSPACE.bzlmod
```diff
@@ -60,11 +61,12 @@ git_repository(
...
     remote = "https://github.com/facebook/react-native",
-    tag = "v{}".format(RN_VERSION),
+    commit = "31f9d45074f069151c155836335e0edff87f70bd"
 )
 
 http_archive(
     name = "fbjni_artifact",
     build_file = "//third_party/fbjni:artifact.BUILD",
     url = "https://repo1.maven.org/maven2/com/facebook/fbjni/fbjni/0.7.0/fbjni-0.7.0.aar",
+    sha256 = "7e319ae110ac5e5ef18904170aea5c3e753e915d196699d7fd39d36c8e1dfe36"
```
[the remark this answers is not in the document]
KetanReddy: I pulled these from the CI when I noticed the errors. It doesn't throw an integrity error so I I think they're right?

.circleci/config.yml
KetanReddy: This and your other comment about the filters are related to the same theoretical change. In essence what this does is same same thing as the existing `bazel test --config=ci -- //third_party/hermes/hello:test` invocation which is to build hermes without any potential flags which may cause bazel to invalidate its cached build. When both test invocations were the same minus setting the android runtime flag it was causing Hermes to rebuild on the subsequent run. This also splits up the actual building of the tests and their execution just to be extra safe around how much is actively running. There is probably a better way to do this though so I am open to suggestions.

third_party/fbjni/BUILD
KetanReddy: Yeah its broadly for every cmake target. The `@hermes//:host` target definitely needs it now but out of future looking caution, since cmake in bazel seems to be having problems identifying the correct concurrency limit I want to apply it everywhere just so we don't have to think about it. It looks like potential fixes for this are in 8 or coming in 9 so we may be able to do away with this when we are able to upgrade our bazel version.

.bazelrc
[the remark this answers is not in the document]
brocollie08: even `JAVA_HOME`?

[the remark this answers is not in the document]
KetanReddy: This is setting it only for repository rules ([docs](https://bazel.build/reference/command-line-reference#common_options-flag--repo_env)). If we set that in the regular env _any_ java call would try and use that jdk which is probably not what we want.

third_party/rn/react_native.BUILD
```diff
@@ -18,6 +18,9 @@ cmake(
     generate_args = ["-G Ninja"],
     lib_name = "jsi",
     lib_source = "jsi_sources",
+    env = {
```
brocollie08, line 21: i don't think this is even being used right now

.bazelrc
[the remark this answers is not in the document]
adierkens: Should this be behind a `build:ci` config? or is changing the JAVA_HOME for all local builds safe too?

third_party/rn/react_native.BUILD
KetanReddy: Like this step at all?

brocollie08: no harm leaving it in though, probably eventually get to use this, but brings up this point
> could we pull this out to config set by command line, or set by something in .bazelrc

when we remove it after upgrade we just remove it in one place

KetanReddy: I don't think we can, I wasn't able to find a way to set args for certain targets via the CLI. If it was a bit harder to find and untangle I would agree but since this is pretty easy to search and remove I am not too worried about it.

spentacular (approved):

PR #636, opened 2025-04-25
Re-add stamping support to builds

adierkens:
Fixes: https://github.com/player-ui/rules_player/issues/77

### Change Type (required)
- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`


react/player/src/player.tsx
```diff
@@ -19,8 +19,9 @@ import type { ReactPlayerProps } from "./app";
 import PlayerComp from "./app";
 import OnUpdatePlugin from "./plugins/onupdate-plugin";
 
-const WEB_PLAYER_VERSION = "__VERSION__";
-const COMMIT = "__GIT_COMMIT__";
+// Variables injected at build time
+const WEB_PLAYER_VERSION = typeof __VERSION__ !== "undefined" ? __VERSION__ : "__VERSION__";
```
KetanReddy, line 23:
```suggestion
const REACT_PLAYER_VERSION = typeof __VERSION__ !== "undefined" ? __VERSION__ : "__VERSION__";
```

react/player/src/player.tsx
```diff
@@ -19,8 +19,9 @@ import type { ReactPlayerProps } from "./app";
...
+const COMMIT = typeof __GIT_COMMIT__ !== "undefined" ? __GIT_COMMIT__ : "__GIT_COMMIT__";
```
KetanReddy, line 24: Do we need to define them here or can we just pull them from the core player object since they should all have the same version and be built from the same commit

adierkens: That's true, but they might be different at runtime -- if your resolutions are pulling a different version of the core dep (or semver ranges resolve to something else)

KetanReddy: Ahh, good point.

KetanReddy (approved):

PR #640, opened 2025-04-30
Shared Flow Expressions Plugin

adierkens:
The Flow Expressions Plugin allows you to define expressions at the flow level and use them as operators in content

### Change Type (required)
- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -60,6 +60,7 @@ plugins/reference-assets/react/node_modules
 plugins/shared-constants/core/node_modules
 plugins/stage-revert-data/core/node_modules
 tools/asset-testing-library/core/node_modules
+plugins/flow-expressions/core/node_modules
 docs/storybook/node_modules
 plugins/reference-assets/mocks/node_modules
 tools/storybook/node_modules
--- a/docs/site/src/content/docs/plugins/core/flow-expressions.mdx
+++ b/docs/site/src/content/docs/plugins/core/flow-expressions.mdx
@@ -1,0 +1,55 @@
+---
+title: Flow Expressions Plugin
+description: A plugin that makes flow-level expressions available as operators in the expression evaluator
+---
+
+# Flow Expressions Plugin
+
+The Flow Expressions Plugin allows you to define expressions at the flow level and use them as operators in your content. This is useful for creating reusable expressions that are specific to a particular flow.
+
+Consider using flow expressions for complex calculations or formatting that would be used multiple times in a flow.
+
+## Usage
+
+```typescript
+import { Player } from "@player-ui/player";
+import { FlowExpressionsPlugin } from "@player-ui/flow-expressions-plugin";
+
+const player = new Player({
+  plugins: [new FlowExpressionsPlugin()]
+});
+```
+
+## Configuration
+
+The plugin automatically reads expressions from your flow definition and makes them available as operators. No additional configuration is required.
+
+## Example
+
+Here's an example of how to use flow expressions in your flow:
+
+```json
+{
+  "id": "test-flow",
+  "expressions": {
+    "isValid": "{{data.value}} > 0",
+    "formatValue": "{{data.value}} + ' units'"
+  },
+  "views": [
+    {
+      "id": "main-view",
+      "type": "view",
+      "title": "Value: @[formatValue()]@",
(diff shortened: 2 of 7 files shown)


rafbcampos (commented):
That's certainly an improvement over the current behaviour, but also an opportunity to discuss if we should prevent inlining expressions and require them to be provided by the registry (through a manifest that describes the player instance and its plugins).
Doing that would allow to no only push towards good practice (separation of concerns between UI and business logic) but also improve developer experience with proper static checks for the expressions and its arguments.
If we only handle references to registered expressions, we don’t need to support Javascript expressions evaluation and allow expressions per platform that run native code. Instead of a string we'd use a common reference type ( Expression { handle: string, args: Array<{name: string, type: int64 | string | boolean | Record <...>}, platform: ‘ios’}>.

KetanReddy:
> if we should prevent inlining expressions and require them to be provided by the registry 
Isn't that what we basically already do by having the expressions be defined by the plugin? It not arbitrary script code that can be executed. If we wanted to put every permutation of every expression that could be called into a registry that seems a bit extreme (unless I am misunderstanding) 
> improve developer experience with proper static checks for the expressions and its arguments
FWIW we do have a POC of this working with the XLR based expressions LSP

To the original point of the PR though, is this functionally any different than composing and reusing expressions in the DSL?

adierkens:
> To the original point of the PR though, is this functionally any different than composing and reusing expressions in the DSL?

The intent is to leverage this to reduce the amount of duplication that the composed expressions produce in content; and also make it easier to start generating more complex functions via expressions that can be independently tested rather than having them be part of the views directly (or exposed via [computed-properties](https://player-ui.github.io/next/plugins/core/computed-properties/) when they shouldn't be

rafbcampos:
Player-UI's expression evaluator parses and executes string-based expressions leveraging JavaScript-specific semantics for operators and syntax.

It supports a fair amount of expressions besides pre-registered function calls:

- Binary operators
- Unary operators
- Model references
- Member expressions (object.property)
- Assignment expressions
- Conditional expressions
- Array expressions
- Object expressions with attributes

Allowing string-based expression is error-prone, and leads to a poor developer experience:

- No autocomplete for expression functions
- No parameter type validation until runtime
- No indication of which expressions are available
- No compile-time error checking

Consider this example:

```ts
exp={e`const x = {"count": ${data.count} = ${data.count} + 1}["count"]`}
```
This would fail at runtime with `Expression type: Compound is not supported`, but shows no errors during authoring and even compiles using the DSL.

Instead of inline expressions, I propose a strongly-typed registry approach:

```ts
interface Expression<T extends string, K extends undefined> {
  handle: T;
  args?: K;
  platform?: 'web' | 'ios' | 'android';
}

// Registration
const incrementCounter: Expression<'incrementCounter', { count: 'number' }> = {
  handle: 'incrementCounter',
  args: {
    count: 'number'
  }
};

// Usage
expression('incrementCounter').args({ count: data.count })
```
There's the idea of having a manifest generated by a Player instance that can inform about the registered expressions and we could use XLR to communicate their types.

That gives us:

1. Platform Independence: Native implementations for each platform
2. Type Safety: Compile-time checking of:
  - Expression existence
  - Argument types
  - Platform availability
3. Better Developer Experience:
  - Autocomplete for available expressions
  - Parameter hints and validation
  - Clear error messages during development
4. Separation of Concerns:
  - Expression logic is defined by plugins/platform
  - Content only references expressions by name with typed args
  - No mixing of UI and business logic
5. Enhanced Tooling:
  - LSP can validate all expression references
  - Generation of documentation from registry
 
> "Isn't that what we basically already do by having the expressions be defined by the plugin?"

Plugins define expressions, but content can still use inline JavaScript via @[...]@ syntax or template literals. This creates an escape hatch from the type system and re-introduces all the problems above.

> "If we wanted to put every permutation of every expression that could be called into a registry that seems a bit extreme"

We wouldn't register permutations, just the base expressions. Complex logic would be composed by calling multiple expressions or creating higher-order expressions that take other expressions as parameters.

adierkens:
We do have utilities that expose expressions to the validator and auto-complete in-editor: 
https://github.com/player-ui/tools/pull/20
https://github.com/player-ui/vscode-player

PR #644, opened 2025-05-07
Enable support async expressions to be processed during flow transitions

adierkens:
### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`


core/player/src/expressions/evaluator.ts
```diff
@@ -417,4 +457,175 @@ export class ExpressionEvaluator {
...
+    options: AsyncHookOptions,
+  ): Promise<unknown> {
+    const { resolveNode, model } = options;
+
+    const expressionContext: ExpressionContext = {
+      ...options,
+      evaluate: (expr) => this.evaluate(expr, options),
+    };
+
+    if (node.type === "BinaryExpression" || node.type === "LogicalExpression") {
```
KetanReddy, line 473: I know the complexity here is having to await the subNode execution but is there any way we could not duplicate the per-node handling logic or abstract that so its reusable somehow?

core/player/src/expressions/evaluator.ts
```diff
@@ -119,12 +131,21 @@ export class ExpressionEvaluator {
...
+  } = {
+    binary: new Map<string, BinaryOperator>(
+      Object.entries(DEFAULT_BINARY_OPERATORS),
+    ),
+    unary: new Map<string, UnaryOperator>(
+      Object.entries(DEFAULT_UNARY_OPERATORS),
     ),
+    expressions: new Map<string, ExpressionHandler<any, any>>([
+      ...Object.entries(DEFAULT_EXPRESSION_HANDLERS),
+      ["await", DEFAULT_EXPRESSION_HANDLERS.waitFor],
```
KetanReddy, line 147: I know core Player is js based but I am hesitant to use the same keywords that javascript uses. On one hand it does make sense because it is in effect the same functionality in the same language but because core is supposed to be cross platform it just feels a bit weird.

adierkens: It's not just JS that uses the `await` keyword/handing -- Python, C#, (I think Dart might as well); though I get where your head is at.

core/player/src/expressions/evaluator.ts
adierkens: Yeah there's def stuff we can pull out and share -- just have to be careful to not over-complicate it w/ avoiding the async/await

core/player/src/expressions/evaluator.ts
KetanReddy: Good point, didn't know about Python and C#. If its a common enough keyword I think it should be fine.

PR #661, opened 2025-06-04
React 18 cleanup

spentacular:
### Change Type (required)

Updating dependencies so React 18 is used everywhere. Applies mostly to `@testing-library` since we were already on React 18 for other parts.

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`


package.json
```diff
@@ -9,9 +9,9 @@
   },
   "engines": {
     "node": "^20.14.0",
-    "pnpm": "^9.7.0"
+    "pnpm": "^10.11.1"
```
KetanReddy, line 12: You'll also need to update the pnpm version in the MODULE.bazel file otherwise things might be wonky when it's run from clean in bazel

PR #665, opened 2025-06-10
update core async node plugins to catch and handle errors

tmarmer:
- Update the `AsyncNodePlugin` to include a new hook `onAsyncNodeError`
  - The intent of this hook is to catch all errors in `onAsyncNode` hook and allow for a fallback node to be inserted in case of an error.
- Update the `AsyncNodePluginPlugin` to call the `onAsyncNodeError` hook when it fails to handle the async node.
- If the call to `onAsyncNodeError` does not return a new node or `null`, then the error is used to fail the current player flow.
  - This is different from the existing behaviour, where the error is left uncaught and just ends up in the browser's console.
  - Even if the error is handled, the error is logged to ensure it doesn't get lost.

##### TODO:
- [ ] Update documentation to add "Error Handling" section to usage of async node docs.
- [ ] Add hook definitions to the mobile wrappers.

### Change Type (required)
- [X] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [X] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/docs/site/src/content/docs/plugins/multiplatform/async-node.mdx
+++ b/docs/site/src/content/docs/plugins/multiplatform/async-node.mdx
@@ -166,6 +166,22 @@ const player = new ReactPlayer({
   </Fragment>
   <Fragment slot='ios'>
 
+### Error Handling
+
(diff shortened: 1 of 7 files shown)


tmarmer:
Moving PR off of fork and onto main repo.

PR #671, opened 2025-06-29
feat(common-expressions): add substr and split string functions

tusharagrawa:
## Description
Adds two new string manipulation functions to the Common Expressions Plugin:

- **substr**: Extracts substrings with support for negative indices and optional length
- **split**: Splits strings with custom separators and optional limit

## Changes
- Implemented `substr` and `split` functions following existing patterns
- Added comprehensive test coverage (13 new tests, 52 total passing)
- Updated documentation with TypeScript signatures

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

## Release Notes
Added `substr` and `split` functions to common expressions plugin that mirror their javascript implementation


plugins/common-expressions/core/src/expressions/index.ts
```diff
@@ -98,6 +98,63 @@ export const sentenceCase = withoutContext(
...
+    }
+
+    if (separator === undefined || separator === null) {
+      return str;
+    }
+
+    const separatorStr = String(separator);
+
+    // Handle empty separator - return the original string as a single element
+    if (separatorStr === "") {
```
KetanReddy, line 114: I would say lets align this functionality to be similar to JavaScript's where an empty string separator splits at every character

plugins/common-expressions/core/src/expressions/index.ts
```diff
@@ -98,6 +98,63 @@ export const sentenceCase = withoutContext(
...
+    }
+
+    const startIndex = typeof start === "number" ? start : toNum(start);
+    if (startIndex === null || startIndex === undefined) {
+      return str;
+    }
+
+    // Handle negative indices by converting to positive indices from the end
+    const actualStartIndex =
+      startIndex < 0 ? str.length + startIndex : startIndex;
```
KetanReddy, line 144: If we're going to add additional functionality on top of the base language implementation of these functions, can we document those decisions?

plugins/common-expressions/core/src/expressions/index.ts
```diff
@@ -98,6 +98,63 @@ export const sentenceCase = withoutContext(
   ifString((str) => str.replace(/\b[a-zA-Z]/, (word) => word.toUpperCase())),
 );
 
+export const split = withoutContext(
+  (str: unknown, separator: unknown, limit?: unknown) => {
```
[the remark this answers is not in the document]
tusharagrawa: Yes if we need to solve only for string cases yes we can remove it then..it will be inline to String's method

plugins/common-expressions/core/src/expressions/index.ts
tusharagrawa: i will update the javadoc and inline comments to clearly portray the implementation, let me know if we are maintaining any other usage doc.

docs/site/src/content/docs/plugins/core/common-expressions.mdx
```diff
@@ -147,6 +147,30 @@ function replace(
...
+
+```ts
+function split(
+  value: string,
+  separator: string,
+  limit?: number,
+): Array<string>;
+```
+
+### substr
```
spentacular, line 162: Potentially support `substring` as well, and maybe all the common JS standard lib functions

KetanReddy: https://github.com/player-ui/player/issues/672

PR #639, opened 2025-04-28
Asset deprecation

brocollie08:
Step towards consolidating the various levels of assets, start by deprecating `SuspendableAsset` to make all `DecodableAsset` suspendable

### Change Type (required)
Indicate the type of change your pull request is:

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

### Does your PR have any documentation updates?
- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.10.5--canary.639.22481</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.10.5--canary.639.22481
</details>

PR #676, opened 2025-07-08
add async hydration tracker plugin as default

sugarmanz:
Since `SuspendableAsset` was deprecated (in #639), consumers will need to add a `AsyncHydrationTrackerPlugin` to enable tracking of asset rendering state. This plugin is used directly in `DecodableAsset` code, and if it's not found, will log a warning. It's not a _real_ warning, and should be disabled to limit noise in logs. We could enable conditional logging in `findPlugin`, or make `AsyncHydrationTrackerPlugin` a default plugin — for the sake of simplicity, this PR opts for the latter.

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

# Release Notes

`AsyncHydrationTrackerPlugin` is now a default `AndroidPlayerPlugin`.


android/player/src/main/java/com/intuit/playerui/android/asset/DecodableAsset.kt
```diff
@@ -114,9 +114,13 @@ public abstract class DecodableAsset<Data>(assetContext: AssetContext, private v
...
-            androidPlayer.onUpdate { _, _ ->
-                synchronized(trackedHydrations) {
-                    trackedHydrations.clear()
+            androidPlayer.hooks.viewController.tap { viewController ->
+                viewController?.hooks?.view?.tap { view ->
+                    view?.hooks?.onUpdate?.tap { _ ->
+                        synchronized(trackedHydrations) {
+                            trackedHydrations.clear()
+                        }
+                    }
```
sugarmanz, line 123: Unfortunately, `androidPlayer.onUpdate` triggers asset decoding logic, which isn't needed for this plugin and causes some redundant logs. This is more verbose, but more targeted and trims redundant operations. Really, it would be great to encapsulate the asset decoding lifecycle as another hook to tap.

KetanReddy:
Are we aware of why the deprecation in #639 was a performance detractor?

KetanReddy (approved):

PR #680, opened 2025-07-10
add bench test for streaming player

tmarmer:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

## Release Notes
- Add initial bench test to async-node plugin
- Add step to run bench tests in `just` and in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ executors:
     docker:
       - image: docker.io/playerui/bazel-docker:13
     working_directory: ~/player
-    resource_class: medium
+    resource_class: medium+
     environment:
       TZ: "/usr/share/zoneinfo/America/Los_Angeles"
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -378,6 +378,14 @@ jobs:
           at: ~/player
       - auto_shipit
 
+  bench_js:
+    executor: base
```
KetanReddy, line 382: For now, since the benchmarks are relatively small can/should we use a smaller executor?

.circleci/config.yml
```diff
@@ -517,6 +525,15 @@ workflows:
...
             - build-ios-fork
 
+      - bench_js:
+          name: js-benchmark-tests
+          filters:
+            branches:
+              ignore:
+                - /pull\/.*/
+          requires:
+            - build-trunk
```
KetanReddy, line 535: Should we also run this for forks?

MODULE.bazel
```diff
@@ -33,6 +33,7 @@ node.toolchain(node_version = "20.14.0")
 ###### Start iOS ######
 ## Rule Dependencies
 bazel_dep(name = "rules_apple", version = "3.5.1", repo_name = "build_bazel_rules_apple")
+# local_path_override(module_name = "rules_apple", path = "/Users/tmarmer/Repos/bazel_rules/rules_apple")
```
KetanReddy, line 36: I would remove this instead of commenting it out

PR #689, opened 2025-07-15
remove DefaultViewPlugins and move implementation into ViewController

tmarmer:
- Remove `DefaultViewPlugins`
- Add all view plugins into the `ViewController` and ensure they get added before the `this.hooks.view` is called
  - This should ensure a consistent tap order for built in view plugins

### Change Type (required)

- [X] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


core/player/src/controllers/view/controller.ts
```diff
@@ -23,16 +33,21 @@ export interface ViewControllerOptions {
   flowController: FlowController;
 }
 
+type ViewControllerHooks = {
```
KetanReddy, line 36: IMO this type should be exported

core/player/src/view/view.ts
```diff
@@ -72,14 +72,24 @@ class CrossfieldProvider implements ValidationProvider {
   }
 }
 
+type ViewHooks = {
```
KetanReddy, line 75: Same thing here. I would export this type

KetanReddy (approved):

PR #649, opened 2025-05-12
Async exp

adierkens:
Re-open https://github.com/player-ui/player/pull/644 on the main branch

### Change Type (required)
Async expression support for action nodes

PR #701, opened 2025-08-06
iOS: calling onupdate in main thread

nancywu1:
Fix for a crash by running flows that use async expressions on iOS 
after the changes in https://github.com/player-ui/player/pull/649/files to support async expressions 
when an async expression causes a view update the[ onViewController](https://github.com/player-ui/player/blob/7207eb0a553c3c1659cc822691d6b2a496f4ce2f/ios/swiftui/Sources/SwiftUIPlayer.swift#L83) tap is being called on a background thread when previously it was called on the main thread

onViewController's view hook called onView which calls the [onUpdate](https://github.com/player-ui/player/blob/7207eb0a553c3c1659cc822691d6b2a496f4ce2f/ios/swiftui/Sources/SwiftUIPlayer.swift#L147) hook which after the async expression changes is being fired on a background thread which triggers asset decode off the main thread which triggers an assertion failure

a work around is to call the onUpdate on the main thread however ideally the [viewcontroller callback]([this callback](https://github.com/player-ui/player/blob/7207eb0a553c3c1659cc822691d6b2a496f4ce2f/ios/swiftui/Sources/SwiftUIPlayer.swift#L83) would always happen on the main thread) before the onUpdate should be called on main thread but when we try to wrap that with main actor nothing is rendered

followup ticket will be made to look into why the async expression is causing viewController.hooks to be called on background thread but the current solution solves the crash and doesnt introduce side effects

one test failure as the result of the change is addressed by delaying the assertions
### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.12.1--canary.701.25487</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.12.1--canary.701.25487
</details>


KVSRoyal (commented):
Lgtm if it passes! I rebased it and added the `patch` label.

PR #703, opened 2025-08-12
support boolean values in binding queries

tmarmer:
- Allow boolean values in binding queries
- Remove tests using `ebnf` or `parsimmon` from `parser.test.ts` since custom parser implementation does not handle strings or booleans the same as those parsers.

### Change Type (required)

- [X] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [X] No Update needed
- [ ] Unable to update docs

## Release Notes
- Update binding parser to allow for boolean value nodes
- Only use boolean values in query values as parts of the path should always be strings or numbers


core/player/src/binding-grammar/__tests__/test-utils/ast-cases.ts
```diff
@@ -54,6 +54,23 @@ export const VALID_AST_PARSER_TESTS: Array<[string, PathNode]> = [
       toQuery(toValue("Month"), toValue("New[.]]Mo,nth.")),
     ]),
   ],
+  // Boolean values only treated as bools for query compare values
+  [
```
KetanReddy, line 58: Can we add a test for comparison with a binding?

core/player/src/binding-grammar/custom/index.ts
```diff
@@ -91,6 +91,15 @@ export const parse: Parser = (path) => {
       value += ch;
     }
 
+    if (allowBoolValue) {
+      if (value === "true") {
+        return toValue(true);
+      }
+      if (value === "false") {
+        return toValue(false);
+      }
```
KetanReddy, line 100: Are we sure we want to cast

PR #708, opened 2025-08-22
update createAsyncTransform to allow for different node types

tmarmer:
### Change Type (required)

- [X] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [X] No Update needed
- [ ] Unable to update docs

## Release Notes
- Update the `createAsyncTransform` util to consider different incoming node types coming from `getNestedAsset` option. Originally, this assumed that it would always be an `asset` node type and would asset-wrap the node, but to allow for some additional flexibility, the following rules are applied:
  - Asset nodes and Applicability nodes with a value that is an asset node get asset wrapped like before.
  - multi-nodes have their values spread into the initialized multi-node to keep the structure flat
  - everything else gets added to the initialized multi-node as-is.

### DIFF
--- a/core/player/src/view/builder/index.ts
+++ b/core/player/src/view/builder/index.ts
@@ -39,11 +39,9 @@ export class Builder {
    * Creates a multiNode and associates the multiNode as the parent
    * of all the value nodes
    *
-   * @param values - the value, applicability or async nodes to put in the multinode
+   * @param values - the nodes to put in the multinode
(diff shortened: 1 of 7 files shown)


KetanReddy:
Looks good! Just one small think about mocking in tests.

KetanReddy (approved):

PR #716, opened 2025-09-05
fix caching issues with resolved async nodes.

tmarmer:
### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


plugins/async-node/core/src/index.ts
```diff
@@ -192,6 +186,8 @@ export class AsyncNodePluginPlugin implements AsyncNodeViewPlugin {
 
       const resolvedNode = context.nodeResolveCache.get(node.id);
       if (resolvedNode !== undefined) {
+        const existingResolvedNodes = node.asyncNodesResolved ?? [];
+        resolvedNode.asyncNodesResolved = [...existingResolvedNodes, node.id];
```
KetanReddy, line 190: Can we append here instead of making a new array?

PR #732, opened 2025-10-16
refactor: replace default exports with named exports

### LINKED ISSUE #669: Remove all default exports from Internal JS Player plugins
We have moved away from using default exports from the first party player plugins (under the `/plugins` directory) that we provide however internally there still are a handful of plugins that still use default exports. For the most part these are in the logger and view submodules but there is one (the `OnUpdatePlugin` plugin) in the react package that should also be updated.

dtychshenko:
All default exports have been replaced with named exports across core packages, plugins, and React components. External dependencies have been verified to be not using the affected default exports that changed. Closes #669. 

### Change Type (required)

- [ ] `patch`
- [X] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [X] No Update needed
- [ ] Unable to update docs

# Release Notes
All default exports have been replaced with named exports across core packages, plugins, and React components. 

Affected packages:
- `@player-ui/make-flow`: identify function
- `@player-ui/partial-match-registry`: createObjectMatcher function
- `@player-ui/player`: Logger classes and view plugins
- `@player-ui/react`: ReactPlayer component and OnUpdatePlugin
- `@player-ui/metrics-plugin`: MetricsCorePlugin class
- `@player-ui/auto-scroll-manager-plugin-react`: scrollIntoViewWithOffset

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+# 0.14.1-next.2 (Thu Oct 16 2025)
+
+### Release Notes
+
+#### replace parallel hook with series hook in async node ([#718](https://github.com/player-ui/player/pull/718))
+
+- make the `onAsyncNode` hook of `AsyncNodePlugin` an `AsyncSeriesBailHook` to allow taps to return `undefined` in order to skip resolving the node.
+- Remove `asyncNode` hook from `AsyncNodePluginPlugin` because it never gets called or tapped.
+
+---
+
+#### 🐛 Bug Fix
+
+- replace parallel hook with series hook in async node [#718](https://github.com/player-ui/player/pull/718) ([@tmarmer](https://github.com/tmarmer))
+
+#### Authors: 1
+
+- Thomas Marmer ([@tmarmer](https://github.com/tmarmer))
+
+---
+
 # 0.14.1-next.1 (Wed Oct 08 2025)
 
 ### Release Notes


dtychshenko:
> @sugarmanz any reason these changes could cause a timeout on the android emulation stage?

I suspect it may have something to do with the `android-test-fork` job itself. I noticed that all the PRs that are being built and merged successfully are running from the main repo branches that execute `android-test-trunk`, while I don't have access to the main repo and using a forked repo.

KetanReddy:
> > @sugarmanz any reason these changes could cause a timeout on the android emulation stage?
> 
> I suspect it may have something to do with the `android-test-fork` job itself. I noticed that all the PRs that are being built and merged successfully are running from the main repo branches that execute `android-test-trunk`, while I don't have access to the main repo and using a forked repo.

Ah, that could be it. Forked jobs don't have access to our build cache so there might be some issue with testing if the the test is already cached where it is failing instead of returning false. Let me pull this branch into the repo and run it through CI from there.

KetanReddy:
> > @sugarmanz any reason these changes could cause a timeout on the android emulation stage?
> 
> It did time out due to no output, but that doesn't mean it wasn't building things behind the scenes. I guess I shouldn't have muted all the logs — was just trying to make it apparent that we were checking for cached results vs actually building/testing. I'll open up a PR to let Bazel log normally during the cache-check.

Ah, I know circle will kill a job if it has no log output for 10m so if it was doing something it probably didn't show as such so it got killed.

KetanReddy:
I think the biggest question here @player-ui/reviewers is if this is going to be a major release? If so we probably need to stage this while we burn down the pre first major list.

PR #750, opened 2025-11-03
Handle edge cases in markdown asset

shahabdsh:
When the value provided to markdown is undefined or an empty string, it errors. This changes makes it return null nodes in such cases.

Also took the chance to refactor `markdownParser.ts` to have clearer code paths. Otherwise, I would have had to repeat the empty handling logic in quite a few places.

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

# Release Notes
Updated `markdown` to handle edge cases (undefined or empty string values)

### DIFF
--- a/plugins/markdown/core/src/utils/markdownParser.ts
+++ b/plugins/markdown/core/src/utils/markdownParser.ts
@@ -41,7 +41,7 @@ export function parseAssetMarkdownContent({
   }
 
   // Map all children to their transformed content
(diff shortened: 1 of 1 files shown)


KetanReddy:
Should we default to an empty text asset? If we were going to provide an opinionated behavior I would say it should just return nothing. We could provide a similar configuration for an empty value.

PR #758, opened 2025-11-13
Set timeout of publish stage to 40m

KetanReddy:
Set timeout to 40m. This is now required because of removing the `-vv` arg from auto which silences the script output during the publish stage which was causing cirlcle to preemptively kill the job assuming nothing was happening. 

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs


KetanReddy:
This timeout is 10m longer than we've seen it run but just adding a bit more time for headroom

PR #648, opened 2025-05-12
Revert "Merge pull request #646 from player-ui/ops/merge-636-and-637"

KetanReddy:
This reverts commit 6ed308100ab3dfe18d439335dc038bf68f8c1bc2, reversing changes made to 6fc1db4ebf87259073d72f33d675d0519823bae6.

### Change Type (required)
Indicate the type of change your pull request is:

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

### Does your PR have any documentation updates?
- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.10.5--canary.648.22787</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.10.5--canary.648.22787
</details>

PR #761, opened 2025-11-19
Add Back in Stamping for JavaScript Builds

KetanReddy:
#636 added support for properly stamping the Core and React JS bundles but had to be rolled back in #648 because the artifacts it produced were broken. After digging into the original issue it _seems_ like it might have been because of a race condition in the source maps that were generated for the `.mjs` and `index.legacy-esm.js` builds as only one was getting the additional substitutions (because of a non-cascading object spread) causing their map files to be slightly different. This applies the substitutions to all js builds including the IIFE for native. This PR also cleans up and marks some legacy artifacts for deprecation from when there were different versions associated with the Core and React player code bases. 

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

## Release Notes
Fix stamping in core/react builds for Player version and commit used to build the bundle

<details>
  <summary>📦 Published PR as canary version: <code>0.14.2--canary.761.29359</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.14.2--canary.761.29359
</details>

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,10 +5,10 @@ module(
 
 bazel_dep(name = "rules_player")
 archive_override(
-    module_name = "rules_player",
-    integrity = "sha256-90jZ8zCyOrUIwYMPH5kb4renSuDeZQ8vXciKhYC7khM=",
-    strip_prefix = "rules_player-2.1.1",
-    urls = ["https://github.com/player-ui/rules_player/archive/refs/tags/v2.1.1.tar.gz"],
+  module_name = "rules_player",
+  strip_prefix = "rules_player-2.2.1",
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -30,7 +30,7 @@ build --remote_cache=grpcs://remote.buildbuddy.io
 build --remote_cache_compression
 build --remote_cache_async
 build --remote_download_toplevel
-build --remote_timeout=3600
+build --remote_timeout=60
```
sugarmanz: What's this do??

rafbcampos (approved):

PR #762, opened 2025-11-23
fix(react-player): expose viewUpdateSubscription to render alt comp's

### LINKED ISSUE #441: React 18 Suspense issue for nested reactPlayer.hooks.webComponent
When leveraging the `reactPlayer.hooks.webComponent` to potentially render an alternative component, React can fall into a suspended state forever because the thrown promise is removed from the tree without ever resolving [here](https://github.com/player-ui/player/blob/bc65945e3b46555f0d4c0a617c0017809a97c49a/react/player/src/player.tsx#L205).

This comes into play when using an External Action Plugin to render an alternative component when a specific `EXTERNAL` is navigated to. This will happen anytime the `reactPlayer.hooks.webComponent` renders something other than the `Component` argument as it's children, since that will cause the final `WebPlayerComponent` to be removed from the tree without the promise ever resolving, keeping it in a suspended state.

Ideally we can better expose the subscription in the react player so that anyone can `publish` to it and also leverage that same subscription for also choosing to render an alternative path. `publish` for this top level would function more closely to a redux action like:

```
reactPlayer.viewUpdateSubscription.publish({
  type: 'alternative-component',
  payload: { /* custom payload */ }
});
```

This way, others wouldn't have to keep their own local subscription to force a re-render of the `reactPlayer.hooks.webComponent`, they could just leverage the top level subscription then to render the alternative path:

```
reactPlayer.hooks.webComponent.tap(this.name, (Comp) => {
  return () => {
    if (reactPlayer.viewUpdateSubscription.get().type === 'alternative-component') {
      return <AlternativeComponent />
    }

    return <Comp />
  };
});
```

dtychshenko:
## Summary
Exposed `viewUpdateSubscription` as a public member of the ReactPlayer per the proposed solution in the issue description. This allows consumers to publish updates to this subscription. Publishing a value with an alternative component payload resolves the pending promise, allowing React to complete the render cycle and switch to the new component without hanging.

Closes #441. 

### Change Type (required)

- [ ] `patch`
- [X] `minor`
- [ ] `major`
- [ ] `N/A`

- [X] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

  # Release Notes
  Exposed `viewUpdateSubscription` to render alternative components from subscribed events:
  ```typescript
  applyReact(rp: ReactPlayer) {
    rp.hooks.webComponent.tap(this.name, (Comp) => {
      return () => {
        if (rp.viewUpdateSubscription.get().type === "alternative-component") {
          return <div>Alternative Component</div>;
        }

        return <Comp />;
      };
    });
  }
  ...
  rp.viewUpdateSubscription.publish({ type: "alternative-component" });
  ```

### DIFF
--- a/react/player/src/__tests__/player.test.tsx
+++ b/react/player/src/__tests__/player.test.tsx
@@ -1,61 +1,97 @@
 import { useSubscribedState } from "@player-ui/react-subscribe";
 import { act, render, screen } from "@testing-library/react";
 import React, { Suspense, type ComponentType } from "react";
-import { describe, expect, test } from "vitest";
+import { describe, expect, it } from "vitest";
 import { ReactPlayer, type ReactPlayerPlugin } from "../player";
 
 describe("ReactPlayer", () => {
-  test("switching from suspending component to an alternative component does not hang", async () => {
-    let rp: ReactPlayer;
+  const genericViewEvent = { type: "generic", id: "view-1" };
+  const altViewEvent = { type: "alternative-component", id: "view-2" };
 
-    const viewUpdateEvent = { type: "alternative-component", id: "1" };
+  it("should switch from suspending component to an alternative component without hanging", async () => {
+    let rp: ReactPlayer;
 
     const TestPlugin: ReactPlayerPlugin = {
       name: "test-plugin",
       applyReact: (reactPlayer) => {
         rp = reactPlayer;
-
-        // Register a dummy asset for the alternative component
-        // This prevents ActualPlayerComp from crashing if it renders before being unmounted
-        const dummyAsset = { type: viewUpdateEvent.type };
-        const dummyComponent = () => <div>Dummy Asset</div>;
-        reactPlayer.assetRegistry.set(dummyAsset, dummyComponent);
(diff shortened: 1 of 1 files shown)


react/player/src/__tests__/player.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const viewUpdateEvent = { type: "alternative-component", id: "1" };
+
+    const TestPlugin: ReactPlayerPlugin = {
+      name: "test-plugin",
+      applyReact: (reactPlayer) => {
+        rp = reactPlayer;
+
+        // Register a dummy asset for the alternative component
+        // This prevents ActualPlayerComp from crashing if it renders before being unmounted
+        const dummyAsset = { type: viewUpdateEvent.type };
```
spentacular: Could we do another test and have this be `type: 'generic'` or something, then render the base component followed by the alternative view to very suspense works between them?

Other than that everything looks 💯 . Thank you!

spentacular (changes requested):

PR #770, opened 2025-12-09
error controller

cehan-Chloe:
Define error controller types and hooks

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

# Release Notes

### Error Controller
Player now has a built-in ErrorController, instantiated automatically when a flow starts and accessible via controllers.error on all platforms.

#### Core API
captureError(error) accepts any error implementing the PlayerErrorMetadata interface (type, optional severity, optional metadata). It fires the onError bail hook, writes to errorState in the data model, and navigates via errorTransitions. Returns true if the error was handled (hook bailed or an errorTransitions route was navigated), false if the flow was failed.

onError is a bail hook — returning true takes ownership of the error, preventing errorState from being written and navigation from occurring. Additional state inspection methods: getCurrentError(), getErrors(), clearErrors(), clearCurrentError().

#### Error Navigation
Flows can define errorTransitions at the node or flow level to route errors to dedicated error views instead of failing the flow:

{ "errorTransitions": { "render": "RENDER_ERROR_VIEW", "\*": "GENERIC_ERROR_VIEW" } }
Node-level takes precedence over flow-level. The "\*" wildcard matches any unhandled type.

#### Render-Time Error Capture
Capture asset render errors. Each platform provides an AssetRenderError / AssetRenderException with the failing asset id and full parent path in the error message


core/player/src/controllers/error/controller.ts
```diff
@@ -0,0 +1,159 @@
...
+   * - Return undefined/false to continue to next handler
+   * - Once true is returned, no further plugins are called
+   */
+  onError: SyncBailHook<[PlayerError], boolean | undefined>;
+}
+
+export interface ErrorControllerOptions {
+  /** Optional logger for error operations */
+  logger?: Logger;
+  /** Required for setting data.errorState */
```
KetanReddy: Do we need to add middleware to prevent writing to that namespace from the Flow?

core/player/src/controllers/error/controller.ts
```diff
@@ -0,0 +1,159 @@
...
+    this.log?.debug("[ErrorController] All errors cleared");
+  }
+
+  /**
+   * Clear only current error, preserve history
+   */
+  public clearCurrentError(): void {
+    this.currentError = undefined;
+    this.log?.debug("[ErrorController] Current error cleared");
+  }
```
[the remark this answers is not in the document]
cehan-Chloe: beside, it leads to the question in middleware, should we allow only delete for data.errorState from error controller or just don't allow delete for `data.errorState`. I kinda lean toward the first one (keep clear functions in error controller, add delete `data.errorState` when clean up and limit delete in middleware only from error controller for `data.errorState`) to provide more flexibility for plugin development, can I get some thoughts? ty @KetanReddy

core/player/src/controllers/error/controller.ts
```diff
@@ -0,0 +1,159 @@
...
+
+/** Default error format for data.errorState  */
+interface FormattedError {
+  message: string;
+  name: string;
+  errorType?: string;
+  severity?: ErrorSeverity;
+  assetId?: string;
+  assetType?: string;
+  bindingPath?: string;
```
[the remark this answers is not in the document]
KetanReddy: I think just generalizing it to a `metaData section` that individual systems can define/extend should be fine. Another approach would be to have systems extend this and use a discriminated union + type guards to allow them to be narrowed when needed.

core/player/src/controllers/error/controller.ts
```diff
@@ -0,0 +1,229 @@
...
+    this.currentError = playerError;
+
+    this.options?.logger?.debug(
+      `[ErrorController] Captured error: ${error.message}`,
+      playerError.metadata,
+    );
+
+    // Notify listeners and check if navigation should be skipped
+    // Plugins can observe the error and optionally return true to bail
+    const shouldSkip = this.hooks.onError.call(playerError) ?? false;
```
KetanReddy: if we're using `false` here should the type on the hook be boolean without undefined?

core/player/src/controllers/error/controller.ts
```diff
@@ -0,0 +1,229 @@
...
+        "[ErrorController] No DataController available",
+      );
+      return;
+    }
+
+    try {
+      const formattedError = this.formatErrorForData(playerError);
+
+      // Temporarily allow writes to errorState
+      this.middleware?.enableWrites();
```
KetanReddy: Instead of having to set state to enable writes, can we use a private symbol from the error controller or something so the middleware can use that to check the origin of the write?

cehan-Chloe: do you mean adding a new flag in `DataModelOptions`?

KetanReddy:
Can you write release notes and maybe work with @tmarmer on adding a section to the 1.0 migration guide?

jvm/core/src/main/kotlin/com/intuit/playerui/core/error/ErrorController.kt
```diff
@@ -0,0 +1,116 @@
...
+ */
+@Serializable(with = ErrorController.Serializer::class)
+public class ErrorController internal constructor(
+    override val node: Node,
+) : NodeWrapper {
+    private val captureError: Invokable<Boolean>? by NodeSerializableFunction()
+    private val getCurrentError: Invokable<Throwable?>? by NodeSerializableFunction()
+    private val getErrors: Invokable<List<Throwable>?>? by NodeSerializableFunction()
+    private val clearErrors: Invokable<Unit>? by NodeSerializableFunction()
+    private val clearCurrentError: Invokable<Unit>? by NodeSerializableFunction()
```
sugarmanz: Can we describe the functions explicitly instead of using `Invokable` since that doesn't define the parameter types?

jvm/core/src/main/kotlin/com/intuit/playerui/core/error/ErrorController.kt
```diff
@@ -0,0 +1,116 @@
...
+    public const val VIEW: String = "view"
+    public const val ASSET: String = "asset"
+    public const val NAVIGATION: String = "navigation"
+    public const val VALIDATION: String = "validation"
+    public const val DATA: String = "data"
+    public const val SCHEMA: String = "schema"
+    public const val NETWORK: String = "network"
+    public const val PLUGIN: String = "plugin"
+    public const val RENDER: String = "render"
+}
```
sugarmanz, line 69: Can we make this an enum if it's meant to be a statically defined collection of constants? Or are we enabling any arbitrary string to be used for an error type? It does look like the [core definition only specifies a `string`](https://github.com/player-ui/player/blob/d4b72610239cc04808f1e31d75af25ab46d047e8/core/player/src/controllers/error/types.ts#L34), just I'd typically expect this to be an enum to encapsulate the collection of strings as a concrete type as [`ErrorTypes` in JS does](https://github.com/player-ui/player/blob/d4b72610239cc04808f1e31d75af25ab46d047e8/core/player/src/controllers/error/types.ts#L8-L21)

jvm/core/src/main/kotlin/com/intuit/playerui/core/error/ErrorController.kt
sugarmanz: Also, it looks like these functions are [required for the base instance](https://github.com/player-ui/player/blob/d4b72610239cc04808f1e31d75af25ab46d047e8/core/player/src/controllers/error/controller.ts#L81) — they should only be nullable if they're nullable in JS land.

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -200,12 +201,13 @@ internal class AsyncNodePluginTest : PlayerTest() {
...
-        val errorMessage = assertThrows<Exception> {
+        val err = assertThrows<JSErrorException> {
             runBlockingTest {
                 player.start(chatMessageContent).await()
             }
-        }.message
-        assertEquals("This is an error message from onAsyncNode", errorMessage)
+        }
+        assertEquals("Error: An error occured during async node resolution. See cause for details.", err.message)
+        assertEquals("This is an error message from onAsyncNode", err.node.getObject("cause")?.getString("message"))
```
sugarmanz, line 210: I think I didn't catch this from the other PR, but is there a way to get the cause message using JVM APIs? Like `err.cause?.message`? The `node` API works, but is a shortcut into interop with JS — I'd love to test against the API we'd expect consumers to use.

ios/core/Sources/Player/HeadlessPlayer.swift
```diff
@@ -54,6 +54,9 @@ public protocol CoreHooks {
     /// Fired when the DataController changes
```
KVSRoyal, line 54: Testing out the new `ios-review` skill. Here's the feedback:

This PR exposes the `ErrorController` to iOS, adds `HookWithResult` for bail hooks, makes `AnyType` `Sendable` by switching `anyDictionary`/`anyArray` to recursive `AnyType` values, adds a `ReadOnlyDataController`, exposes `errorController` on `CoreHooks`/`PlayerControllers`, and cleans up SWIFT_PACKAGE conditionals throughout.

## Violations

| Rule | File | Line | Issue |
|------|------|------|-------|
| Rule 3 | `ErrorController.swift` | 72–80 | Doc comment lists removed parameters (`errorType`, `severity`, `metadata`) |

### Rule 3 — Stale doc comment

[ErrorController.swift:72-80](ios/core/Sources/Types/Core/ErrorController.swift#L72-L80) — `captureError` doc lists parameters `errorType`, `severity`, `metadata` that were removed from the signature (now only takes `error: Error`). Remove the dead `- parameters` entries.

ios/core/Sources/Types/Core/ErrorController.swift
```diff
@@ -0,0 +1,130 @@
...
+    public static let view = "view"
+    public static let asset = "asset"
+    public static let navigation = "navigation"
+    public static let validation = "validation"
+    public static let data = "data"
+    public static let schema = "schema"
+    public static let network = "network"
+    public static let plugin = "plugin"
+    public static let render = "render"
+}
```
KVSRoyal: Since this is a finite set, can it just be a `public enum ErrorTypes: String`?

ios/core/Sources/Types/Core/ErrorController.swift
```diff
@@ -0,0 +1,130 @@
...
+        } else {
+            args.append(value.context.error(for: PlayerError.unknownResponse(error)) as Any)
+        }
+        
+        let result = value.invokeMethod("captureError", withArguments: args)
+        if let boolResult = result, boolResult.isBoolean {
+            return boolResult.toBool()
+        }
+        
+        return false
```
KVSRoyal: Can we do something like this to avoid using `var`? We prefer `let` where possible for ios stuff. Also, in a case like this, using `guard` instead of `if` makes the happy path clearer. E.g.

```swift
let convertibleError = (error as? JSConvertibleError & Error) ?? PlayerError.unknownResponse(error)
let args = [value.context.error(for: convertibleError) as Any]

guard let result = value.invokeMethod("captureError", withArguments: args), result.isBoolean else {
    return false 
}
return result.toBool()
```

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -9,6 +9,11 @@ import Foundation
 import JavaScriptCore
 
 extension JSValue {
+    
+    internal struct TryCatchResultKeys {
+        static let success = "success"
+        static let result = "result"
+    }
```
KVSRoyal: Nit: can we use an `enum` here for namespacing instead of a struct? It would otherwise stay the same. Follows this guideline: https://google.github.io/swift/#nesting-and-namespacing that we've aligned on for ios.

KVSRoyal (commented):
Just a few nits

ios/core/Sources/utilities/JSUtilities.swift
```diff
@@ -55,15 +55,41 @@ public class JSUtilities {
...
 }
 
+internal enum JSClass: String {
+    case error = "Error"
+    case promise = "Promise"
+}
+
 internal extension JSContext {
+    func getJSClass(_ jsClass: JSClass) -> JSValue {
+        objectForKeyedSubscript(jsClass.rawValue)
```
JunDangIntuit, line 71: return type should be optional JSValue: func getJSClass(_ jsClass: JSClass) -> JSValue?
as objectForKeyedSubscript(jsClass.rawValue) could be nil.

ios/core/Sources/utilities/JSUtilities.swift
```diff
@@ -72,3 +98,9 @@ public protocol JSConvertibleError {
     /// The description to use when send to JavaScriptCore
     var jsDescription: String { get }
 }
+
+public protocol ErrorWithMetadata : Error {
```
JunDangIntuit: 1. documentation is missed here.
2. JSConvertibleError and ErrorWithMetadata can be merged, so for every type that conforms to ErrorWithMetadata also conformed JSConvertibleError:
public protocol ErrorWithMetadata: Error, JSConvertibleError {
    var type: String { get }
    var severity: ErrorSeverity? { get }
    var metadata: [String: Any]? { get }
}
Then anything like this:
private struct ErrorWithAnyMetadata: Error, ErrorWithMetadata, JSConvertibleError {...}
could be simplified as:
private struct ErrorWithAnyMetadata: ErrorWithMetadata {...}
since ErrorWithMetadata now inherits from both Error and JSConvertibleError, so conformers get both.

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -23,37 +28,96 @@ extension JSValue {
...
-                       return e
+                       return {
+                           \(TryCatchResultKeys.result): fn(...args),
+                           \(TryCatchResultKeys.success): true
+                       }
+                   } catch(error) {
+                       return {
+                           \(TryCatchResultKeys.result): error,
+                           \(TryCatchResultKeys.success): false
+                       }
```
JunDangIntuit, line 36: This is an improvement compared to the previous version. just one suggestion, may we rename tryCatch method to a more meaningful name, like callWithErrorHandling(args:)?

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -23,37 +28,96 @@ extension JSValue {
...
 }
 
 /**
  Represents the different errors that occur when evaluating JSValue
  */
-public enum JSValueError: Error, Equatable {
-    case thrownFromJS(message: String)
-}
\ No newline at end of file
+public struct JSValueError: Error, CreatedFromJSValue, ErrorWithMetadata {
```
JunDangIntuit: no need to conform to Error since ErrorWithMetadata already conforms to Error

jvm/core/src/main/kotlin/com/intuit/playerui/core/error/ErrorController.kt
cehan-Chloe: ya. i'm enabling the arbitrary string here like core type using only string thinking of when consumers tap in onError hook, type is a required field. so this list is meant to be a collection of existing types in Player but not an exhaust list. in this case, is it still better practice to use enum?

cc. @KVSRoyal  same for ios, i was using struct in ios for the errorTypes.

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -23,37 +28,96 @@ extension JSValue {
...
+            isErrorWithMetadata = false
+            type = JSValueError.defaultType
+        }
+        
+        severity = ErrorSeverity(rawValue: jsErrorObject.objectForKeyedSubscript(JSKeys.severity)?.toString() ?? "")
+        metadata = jsErrorObject.objectForKeyedSubscript(JSKeys.metadata)?.toDictionary() as? [String: Any]
+    }
+}
+
+extension JSValueError: JSConvertibleError {
```
JunDangIntuit: once ErrorWithMetadata conforms to JSConvertibleError, it would be simplified as: extension JSValueError {}

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -23,37 +28,96 @@ extension JSValue {
...
+        static let metadata = "metadata"
+    }
+    
+    public static func createInstance(value: JSValue) -> JSValueError {
+        return JSValueError(value)
+    }
+    
+    public init(_ jsErrorObject: JSValue) {
+        originalJSError = jsErrorObject
+        if !jsErrorObject.isInstance(of: jsErrorObject.context.getJSClass(.error)) {
```
JunDangIntuit: this jsErrorObject.context could be nil and would have a potential crash if writng as line 91. suggest to do this:
guard
        let context = jsErrorObject.context, 
        jsErrorObject.isInstance(of: context.getJSClass(.error))
    else {
        message = JSValueError.defaultMessage
        type = JSValueError.defaultType
        severity = nil
        metadata = nil
        isErrorWithMetadata = false
        return
    }

ios/core/Sources/Types/Core/ErrorController.swift
```diff
@@ -0,0 +1,130 @@
...
+//
+//  Created by Player Team
+//
+
+import Foundation
+import JavaScriptCore
+
+/**
+ Severity levels for errors
+ */
```
JunDangIntuit: please change comment style of /** ... */ to /// for an iOS side alignment. moreover, /** ... */ is for multiple lines doc.

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
cehan-Chloe: I'm looking into how it's implemented in core and serialized in jvm. In `AsyncNodeError` [core](https://github.com/player-ui/player/blob/error-controller/plugins/async-node/core/src/AsyncNodeError.ts#L22) we have `cause` but in [JSErrorException](https://github.com/player-ui/player/blob/main/jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/JSErrorException.kt#L11) there is using `innerErrors`. I didn't find doc for this. it seems like our custom fields? should we change it to `cause`?

jvm/core/src/main/kotlin/com/intuit/playerui/core/error/ErrorController.kt
cehan-Chloe: Note: (error: Throwable) -> Boolean. deprecate invokables and update other controller
Related issue https://github.com/player-ui/player/issues/852

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
cehan-Chloe: Note: TODO added in `ThrowableSerializer`

ios/core/Sources/utilities/JSValue+Extensions.swift
cehan-Chloe: JSValueError already conformed to ErrorWithMetadata and ErrorWithMetadata conform to JSConvertibleError, do i still need extension JSValueError {}?

KVSRoyal (approved):

ios/core/Sources/Types/Core/ErrorController.swift
```diff
@@ -0,0 +1,146 @@
...
+/// A wrapper around the JS ErrorController in the core player
+public class ErrorController: CreatedFromJSValue {
+    /// Typealias for associated type
+    public typealias T = ErrorController
+
+    /**
+     Creates an instance from a JSValue, used for generic construction
+     - parameters:
+        - value: The JSValue to construct from
+     */
```
JunDangIntuit: please convert this to:
/// Creates an instance from a JSValue, used for generic construction
/// - Parameters:
///   - value: The JSValue to construct from

and every comments should use /// to align with the guidance. https://google.github.io/swift/

Image

tmarmer (approved):

JunDangIntuit (approved):

react/player/src/asset/index.tsx
```diff
@@ -104,9 +104,8 @@ export const ReactAsset = (
   return (
     <ErrorBoundary
       fallbackRender={(props) => {
-        const { error, resetErrorBoundary } = props;
+        const { error } = props;
 
-        resetErrorBoundary();
```
sugarmanz: Just calling this out to make sure its intended

sugarmanz (approved):

react/player/src/asset/index.tsx
sugarmanz: Oh, I guess I wasn't reviewing the latest 🤦 leaving unresolved to make sure at least

cehan-Chloe: This resetErrorBoundary is not needed. Removed in the latest.

core/player/src/controllers/error/__tests__/controller.test.ts
```diff
@@ -0,0 +1,359 @@
...
+      expect(errorController.getErrors()).toEqual([]);
+    });
+
+    it("should delete errorState from data model", () => {
+      errorController.captureError(
+        new ErrorWithProps("Test error", "test-error"),
+      );
+
+      // Reset mock to track only clearErrors call
+      vitest.clearAllMocks();
```
spentacular: Can we make this more specific instead of the root `vitest` for all of these?

cehan-Chloe: thanks. updated

PR #774, opened 2025-12-12
fix tests with broken context mocks in android/player

tmarmer:
Issue: Some unit tests in `android/player` were not running because `MockK` throwing errors when it could not create a mock of `android.content.Context`. The errors were being swallowed by something and the test runner would treat each of these files as having no tests rather than causing a failure.

Fix: 
- Move broken tests to be instrumented tests under `androidTest`.
- Migrate these tests from using junit5 to junit4
  - To support this, updated `io.mockk:mockk` to version `1.12.5`. `JUnit4` support was added in `1.12.4` ([notes](https://github.com/mockk/mockk/releases/tag/1.12.4)) so I updated to the latest non-breaking version I could find.
- use the application context instead of a mock context
- Fix any test failures after doing the above. These failures were
  - Some tests in `AndroidPlayerTest` checked the number of plugins. The list is now size 6 instead of 5
  - Some tests in `SimpleAssetTest` were failing with broken styles. Changed style used in these tests
  - `HydrationScopeTest` was failing the test `test hydration scope is cancelled on flow end` due to a race condition in the assertions. Added a `waitForCondition` to fix this.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [X] `N/A`

- [ ] Updated docs
- [X] No Update needed
- [ ] Unable to update docs


sugarmanz (approved):

android/player/BUILD
```diff
@@ -47,12 +47,21 @@ unit_test_deps = [
...
+    "//jvm/testutils",
     "//android/testutils:with-runtime",
+    artifact("io.mockk:mockk"),
+    artifact("org.jetbrains.kotlinx:kotlinx-coroutines-test"),
+    artifact("androidx.test:core"),
+    "//jvm:kotlin_serialization"
 ]
 
+# TODO: Update this to be updated automatically instead of maintaining a list manually
+instrumented_test_classes = ['com.intuit.playerui.android.AndroidPlayerTest', 'com.intuit.playerui.android.AssetContextTest', 'com.intuit.playerui.android.AssetContextAndroidTest', 'com.intuit.playerui.android.renderer.NestedAssetTest', 'com.intuit.playerui.android.renderer.HydrationScopeTest', 'com.intuit.playerui.android.renderer.SimpleAssetTest', 'com.intuit.playerui.android.renderer.BrokenAssetTest', 'com.intuit.playerui.android.extensions.IntoKtTest', 'com.intuit.playerui.android.registry.RegistryPluginTest']
```
brocollie08: do we need to do this in order to get all these tests picked up

sugarmanz: Yeah — there is logic [here](https://github.com/player-ui/rules_player/blob/081cc912fd050b4acb8f60fe901c07970932db60/kotlin/private/kt_android_library_and_test.bzl#L106-L121) for automatically compiling this list, but it picks up _every_ file. Ideally, we can expose that logic to let us filter out files (like not ending in `Test`) to build up the list here. We could also put the filter in the rule, but that's pretty opinionated.

tmarmer: Yeah, if I don't add them manually, it picks up every `.kt` file as a test and any of the util / non-test files end up creating test targets that just throw errors.

brocollie08 (approved):

PR #775, opened 2025-12-22
error controller navigation

cehan-Chloe:
When an error occurs in a state, ErrorController tries in order:

- Node-level errorState transition - Most specific, handles errors for this state only
- Flow-level errorState - General fallback, handles errors from any state in flow
- Reject flow promise - No error handling defined
Error with no node/flow errorState → flowResultDeferred.reject(error)

detail doc: https://github.com/player-ui/player/pull/770/commits/4481c719d9cf295a8a1b311bee970a2520d8bc43

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/core/types/src/index.ts
+++ b/core/types/src/index.ts
@@ -134,8 +134,7 @@ export interface NavigationFlow {
     | Expression
     | ExpressionObject
     | NavigationFlowState
-    | NavigationFlowTransition
-    | Record<string, string>;
+    | NavigationFlowTransition;
 }
 
 export type NavigationFlowTransition = Record<string, string>;


core/player/src/controllers/flow/flow.ts
```diff
@@ -150,6 +151,46 @@ export class FlowInstance {
     return this.flowPromise.promise;
   }
 
+  /**
+   * Navigate to flow-level errorState (fallback when node-level fails)
+   * Uses direct navigation, bypassing transition map lookup
+   * @param errorType Optional error type for dictionary-based error transition
+   * @returns true if navigation succeeded, false if errorState not defined or navigation failed
+   */
+  public transitionToErrorState(errorType?: string): boolean {
```
KetanReddy: Why do we need this function? Couldn't we just use the existing `transition` function?

cehan-Chloe: For node level, I still use the existing `transition` function. This is the flow-level error we don't have transitions map, so I use error type as the key. Do you think it's better to have consistency between flow and node? I can add `transitions` to flow level.
samples:
flow:   
```
"errorState": {
    "network": "NETWORK_ERROR_VIEW",
  }
```

node:
```
    "SHIPPING_VIEW": {
      "state_type": "VIEW",
      "ref": "shipping-form",
      "errorState": "node-error",
      "transitions": {
        "node-error": "NODE_ERROR_VIEW"
      }
    },
```

KetanReddy: I think so. Also if we can centralize some of the logic here and in `navigateToErrorState` and make them more utility oriented I think it might be cleaner long term as that logic needs to evolve

core/types/src/index.ts
```diff
@@ -116,16 +116,36 @@ export interface NavigationFlow {
   /** An optional expression to run when this Flow ends */
   onEnd?: Expression | ExpressionObject;
 
+  /**
+   * An optional flow-level transitions map (fallback when node-level is not defined).
+   * Used as a fallback when the current state doesn't have a transition defined.
+   */
+  transitions?: NavigationFlowTransition;
```
KetanReddy: we might want to rename this to make it apparent its for errors. Should we call it the same thing as the state level error transition map?

KetanReddy: Were we not going to differentiate between normal and error transitions or am I misunderstanding?

cehan-Chloe: Before I'm using different transition maps which is the `errorState` or can be called `errortransition`. It supports both string and map. I can revert this change by removing this extra transitions map in flow and reuse the flow controller transition func to have another param to tell which map to look for. Default is the `transitions` and in error controller pass in `errortransition`. Then the flow will be like, but it won't support string as it's not supported by existing transition map
```json
{
  "navigation": {
    "BEGIN": "FLOW_1",
    "FLOW_1": {
      "startState": "VIEW_1",
      "errorTransition": {
        "binding": "BINDING_ERROR_VIEW",
        "validation": "VALIDATION_ERROR_VIEW",
        "*": "GENERIC_ERROR_VIEW"
       }
      "VIEW_1": {
        "state_type": "VIEW",
        "ref": "main-view",
        "errorTransition": {
          "binding": "BINDING_ERROR_VIEW",
          "validation": "VALIDATION_ERROR_VIEW",
          "*": "GENERIC_ERROR_VIEW"
         }
        "transitions": {
          "*": "VIEW_2"
        }
      },
      "VIEW_2": {
        "state_type": "VIEW",
        "ref": "second-view",
        "transitions": {
          "*": "END_Done"
        }
      },
      "ERROR_VIEW": {
        "state_type": "VIEW",
        "ref": "error-view",
        "transitions": {
          "*": "END_Error"
        }
      }
    }
  }
}
```

core/types/src/index.ts
```diff
@@ -116,12 +116,26 @@ export interface NavigationFlow {
...
+
   [key: string]:
     | undefined
     | string
     | Expression
     | ExpressionObject
-    | NavigationFlowState;
+    | NavigationFlowState
+    | NavigationFlowTransition
+    | Record<string, string>;
```
KetanReddy: I think `NavigationFlowTransition` already covers this case?

PR #779, opened 2026-01-13
Use Xcode 26.4.0 in CI/CD

KVSRoyal:
# What this means for you

- You can now use Xcode 26.1-26.4 for local development! 🎉 
- Please ensure you have iOS 26.4 iPhone 17 Pro sim installed locally

# What Changed

> [!NOTE]
> There is an issue with SwiftLint & Xcode 26 that blocked this PR:
> `https://github.com/realm/SwiftLint/issues/6433`
> It is now resolved since CircleCi Xcode 26.4 runs on MacOS 26.4

Update our pipeline to use:

What | Changed to | Why
-- | -- | --
Xcode | 26.4 | Latest available from CircleCI
iOS sim | 26.4, iPhone 17 Pro | Latest available from CircleCI
rules_xcodeproj (bazel) | 3.6.0 | Latest compatible
rules_apple (bazel) | 4.3.3 | Latest compatible
apple_support (bazel) | 2.5.2 | Latest compatible
rules_swift (bazel) | 3.4.2 | Latest compatible
rules_swift_package_manager (bazel) | 1.11.1 | Stuck here — 1.12+ breaks @bazel_lib visibility from aspect_rules_js
swiftlint (bazel) | 0.63.2 | Old version incompatible with above dep changes
gazelle (bazel) | Removed | Not actually used; was interfering with Android tests
rules_shell (bazel) | 0.7.1 | Latest compatible
ViewInspector (SPM) | 0.10.3 | 0.10.2 doesn't support calling onTapGesture through a ViewModifier under Xcode 26 — fixes testLinkView

Also, add a `patch` per suggestion from @sugarmanz to fix this issue:
```bash
ERROR: /home/circleci/.cache/bazel/_bazel_circleci/e8362d362e14c7d23506d1dfa3aea8b8/external/rules_android+/src/tools/java/com/google/devtools/build/android/dexer/BUILD:15:12: @@rules_android+//src/tools/java/com/google/devtools/build/android/dexer:DexFileSplitter depends on @@rules_android++com_android_dex_extension+com_android_dex//:dex in repository @@rules_android++com_android_dex_extension+com_android_dex which failed to fetch. no such package '@@rules_android++com_android_dex_extension+com_android_dex//': java.io.IOException: Error downloading [https://android.googlesource.com/platform/dalvik/+archive/5a81c499a569731e2395f7c8d13c0e0d4e17a2b6.tar.gz] to /home/circleci/.cache/bazel/_bazel_circleci/e8362d362e14c7d23506d1dfa3aea8b8/external/rules_android++com_android_dex_extension+com_android_dex/temp8772008869777028469/5a81c499a569731e2395f7c8d13c0e0d4e17a2b6.tar.gz: GET returned 429 Too Many Requests
Use --verbose_failures to see the command lines of failed build steps.
```

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -18,8 +18,8 @@ common --android_platforms=@rules_android//:x86,@rules_android//:x86_64,@rules_a
 mobile-install --mobile_install_aspect=@rules_android//mobile_install:mi.bzl --mobile_install_supported_rules=android_binary
 
 # iOS Flags
-common --ios_simulator_device="iPhone 16 Pro"
-common --ios_simulator_version="18.5"
+common --ios_simulator_device="iPhone 17 Pro"
+common --ios_simulator_version="26.4"
 
 # Cache Flags
 common --action_cache_store_output_metadata --experimental_remote_cache_compression_threshold=100
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ executors:
     working_directory: ~/player
     resource_class: m4pro.medium
     macos:
-      xcode: 16.4.0
+      xcode: 26.4.0
     environment:
       TZ: "/usr/share/zoneinfo/America/Los_Angeles"
   android:
@@ -150,9 +150,9 @@ jobs:
             HOMEBREW_NO_AUTO_UPDATE=1 brew install bazelisk lcov
 
       - macos/preboot-simulator:
-          version: "18.5"
+          version: "26.4"
           platform: "iOS"
-          device: "iPhone 16 Pro"
+          device: "iPhone 17 Pro"
 
       - run: bazel build --config=ci-mac -- //ios/demo:PlayerUIDemo
 
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,11 +5,12 @@ module(
 
 bazel_dep(name = "rules_player")
 archive_override(
-  module_name = "rules_player",
-  strip_prefix = "rules_player-2.2.1",
-  urls = ["https://github.com/player-ui/rules_player/archive/refs/tags/v2.2.1.tar.gz"],
-  integrity = "sha256-vRyADsmF6vA865X9euZw5Z4KnGgpht5qzaa/vLxSkYQ="
+    module_name = "rules_player",
+    integrity = "sha256-+8W4qKFiv8zQy+NyXcWqULo49VqwTuf/DIMg2ikwvho=",
+    strip_prefix = "rules_player-2.2.3",
+    urls = ["https://github.com/player-ui/rules_player/archive/refs/tags/v2.2.3.tar.gz"],
 )
+
 bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "aspect_bazel_lib", version = "2.19.4")
 bazel_dep(name = "aspect_rules_js", version = "2.3.8")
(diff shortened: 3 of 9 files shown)


MODULE.bazel
```diff
@@ -101,9 +102,9 @@ build_constants(
 bazel_dep(name = "rules_java", version = "8.13.0")
 bazel_dep(name = "rules_kotlin", version = "2.1.8")
 archive_override(
-    module_name = "rules_kotlin", 
-    urls = ["https://github.com/bazelbuild/rules_kotlin/releases/download/v2.1.8/rules_kotlin-v2.1.8.tar.gz"],
+    module_name = "rules_kotlin",
```
KVSRoyal, line 104: This just got reorganized by `bazel mod tidy`. No actual change.

maven-lock.json
```diff
@@ -1,8 +1,11 @@
 {
   "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": "THERE_IS_NO_DATA_ONLY_ZUUL",
-  "__INPUT_ARTIFACTS_HASH": 939461660,
-  "__RESOLVED_ARTIFACTS_HASH": 829057949,
+  "__INPUT_ARTIFACTS_HASH": -1748226514,
+  "__RESOLVED_ARTIFACTS_HASH": -1956390637,
   "conflict_resolution": {
+    "com.google.code.gson:gson:2.10.1": "com.google.code.gson:gson:2.8.9",
+    "com.google.errorprone:error_prone_annotations:2.23.0": "com.google.errorprone:error_prone_annotations:2.5.1",
+    "com.google.guava:guava:33.0.0-jre": "com.google.guava:guava:32.0.1-jre",
```
KVSRoyal: Note to self: undo this change before review. Definitely does not seem right to be downgrading.

.bazelrc
```diff
@@ -18,8 +18,8 @@ common --android_platforms=@rules_android//:x86,@rules_android//:x86_64,@rules_a
 mobile-install --mobile_install_aspect=@rules_android//mobile_install:mi.bzl --mobile_install_supported_rules=android_binary
 
 # iOS Flags
-common --ios_simulator_device="iPhone 16 Pro"
-common --ios_simulator_version="18.5"
+common --ios_simulator_device="iPhone 17 Pro"
+common --ios_simulator_version="26.4"
```
sugarmanz, line 22: Can we move this to CI specific .bazelrc so it'll pick up whatever simulator I have running locally without needing to comment these out?

https://github.com/player-ui/player/blob/693615a5ff79c5d9329fd48fddd171c273583583/.circleci/config.yml#L93 or maybe just prefixing with `ci:` would work

KetanReddy:
Quick question on this

```
gazelle (bazel) 	Removed 	Not actually used; was interfering with Android tests
```

Was this interference just with these other upgrades or is this also happening on main?

.bazelrc
KVSRoyal: I can remove it, but it still probably won't use the simulator you expect if I do. If you remove this, typically it will pick the first available sim, which often is an old phone with old ios. I think this line is actually helpful to ensure that people that don't typically do ios development are testing with something similar.

I personally recommend running ios through Xcode, which lets you pick whatever sim you want.

However, what has been your experience with running the ios tests with this line commented out?

.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ executors:
     working_directory: ~/player
```
KVSRoyal, line 32: Unrelated to this line.
> Quick question on this
> 
> ```
> gazelle (bazel) 	Removed 	Not actually used; was interfering with Android tests
> ```
> 
> Was this interference just with these other upgrades or is this also happening on main? 

@KetanReddy Interference with the other updates. This is not happening on main.

.bazelrc
sugarmanz: All the Bazel repos, if I am invoking iOS tests or demo app, first thing I do is comment this out first thing cause typically I already have an iOS simulator running and if they aren't the same it doesn't work. We could take the strategy of requiring everyone to match CI environment, but that is a higher barrier to entry for contributing — even if someone is developing with an older simulator, CI will test it against the latest we have configured.

We did move it for [devtools](https://github.com/player-ui/devtools/blob/e6b5f20d5380928215afade6e3fdc584e5d4548f/.circleci/config.yml#L99-L101) — just want to be consistent.

KVSRoyal: Fair enough. Okay, moving to ci only.

KVSRoyal: I revise my opinion. I got rid of the lines and know it's running on an iPhone 6s locally no matter what I have open and breaking my stuff. I think this is objectively worse. I'd rather use a sim that's shipped with Xcode which should mean everyone will have it anyway.

This is what it's using for me locally, even after a clean:
```swift
Name: New-iPhone 6s Plus-26.4
OS: iOS 26.4
Type: iPhone 6s Plus
```

The sim/ios in the `.bazelrc` ships with Xcode 26.4 so everyone should have it or can get it. And if not, using Xcode is always good.

sugarmanz: Sounds good — should we add some line to contributing around how to test locally for other simulators? I'm imagining a scenario where an issue only affects a certain version and we need to verify. Or is Xcode UI the only way you think we should be able to do this?

KVSRoyal: I think Xcode UI is definitely the preferred way for any debugging in Swift, yeah.

nancywu1 (approved):

JunDangIntuit (approved):

.bazelrc
sugarmanz: > The sim/ios in the .bazelrc ships with Xcode 26.4

Does that mean we're requiring 26.4 for development then?

sugarmanz (approved):

.bazelrc
KVSRoyal: No. You can get that sim/ios in whichever Xcode you have. It just comes automatically with 26.4

PR #780, opened 2026-01-14
Docs: Plugin Architecture/Hooks + Landing Page Fix

kharrop:
- Fixing a small bug where a comment was unintentionally being displayed on the landing page.
- Adding plugin architecture documentation
- Adding a script that generates a `hooks.mdx` docs page

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,28 @@
+# 0.15.0-next.4 (Fri Jan 16 2026)
+
+### Release Notes
+
+#### Remove chakra/emotion and unused Storybook components ([#776](https://github.com/player-ui/player/pull/776))
+
+- Remove Chakra/Emotion dependencies in Storybook.
+
+---
+
+#### 🚀 Enhancement
+
+- add asset render errors to each platform [#765](https://github.com/player-ui/player/pull/765) ([@tmarmer](https://github.com/tmarmer))
+
+#### 🐛 Bug Fix
+
+- Remove chakra/emotion and unused Storybook components [#776](https://github.com/player-ui/player/pull/776) ([@KetanReddy](https://github.com/KetanReddy))
+
+#### Authors: 2
+
+- Ketan Reddy ([@KetanReddy](https://github.com/KetanReddy))
+- Thomas Marmer ([@tmarmer](https://github.com/tmarmer))
+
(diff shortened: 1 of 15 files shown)


intuit-svc:
## Docs Preview

A preview of your PR docs was deployed by CircleCI [#29952](https://circleci.com/gh/player-ui/player/29952) on `Wed, 14 Jan 2026 22:34:35 GMT`

### 📖 Docs ([View site](https://player-ui.github.io/pr/780/))
- [guides/writing-plugins](https://player-ui.github.io/pr/780/guides/writing-plugins)
- [index](https://player-ui.github.io/pr/780/)
- [player/about](https://player-ui.github.io/pr/780/player/about)
- [plugins/architecture](https://player-ui.github.io/pr/780/plugins/architecture)
- [plugins/index](https://player-ui.github.io/pr/780/plugins)

scripts/docs/generate-hooks-docs.mjs
```diff
@@ -0,0 +1,622 @@
+/**
```
KetanReddy: Is this something we should integrate into the rules so that is can also be run for plugins that have hooks?

package.json
```diff
@@ -12,6 +12,9 @@
     "pnpm": "^10.18.1"
   },
   "packageManager": "pnpm@10.18.1",
+  "scripts": {
+    "docs:hooks": "node scripts/docs/generate-hooks-docs.mjs"
+  },
```
KetanReddy: Do we need to actually run this manually or is this just for convenience?

KVSRoyal (dismissed):
I feel underqualified to review the new js code, but the new docs look good to me. These are really great for figuring out what hooks are available, which I appreciate.

package.json
kharrop: So, currently I have this manual, because it started getting a little complex with having it be more automated since that would mean it would be a non-committed file; which is fine because we could just document this. 

I didn't know if we wanted to keep it manual and alter automate or just automate it out of the box. Open to discussing which approach we want to take.

PR #788, opened 2026-01-29
Make AnyType Sendable

KVSRoyal:
> [!NOTE]
> Contribution originally by @JJ-Intuit 
> This is a clone of #782 that is a direct branch of player, rather than a fork. Forks don't run on CI/CD properly.

## What

This PR makes the `AnyType` enum conform to the `Sendable` protocol by removing all `Any`-based APIs and using only recursive `AnyType` values. This enables safe usage across Swift concurrency boundaries but introduces breaking changes for code that relied on `Any` interoperability.

See "Release Notes" for full description. (Moved there to avoid redundancy.)

## Why

With Swift 6's strict concurrency checking, `AnyType` needed to be `Sendable` to be used safely in concurrent contexts.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`; going into Player 1.0

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### Breaking Change: `.anyArray` and `.anyDictionary` use `AnyType`

The `AnyType` enum now conforms to `Sendable` protocol and is concurrency-safe. This required removing all APIs that depended on the `Any` type:

```swift
// Before
case anyDictionary(data: [String: Any])
case anyArray(data: [Any])

// After  
case anyDictionary(data: [String: AnyType])
case anyArray(data: [AnyType])
```

#### Migration
Use `AnyType` cases directly instead of converting from/to `Any`:

```swift
// Old (removed)
let anyType = AnyType.anyDictionary(data: ["key": "value"])
let anyArray = AnyType.anyArray(data: ["key", 2])

// New (Sendable-safe)  
let anyType = AnyType.anyDictionary(data: ["key": .string(data: "value")])
let anyArray = AnyType.anyArray(data: [.string(data: "key"), .number(data: 2)])
```

### Breaking Change: `AnyType` will only decode `.unknownData` with `AnyTypeDecodingContext`

Users are required to provide `AnyTypeDecodingContext` when attempting to decode `anyArray` or `anyDictionary`. The decoder will throw if we need the context but one is not provided. As a side effect of this, `.unknownData` will only be returned when a context is provided. For example:

```swift
// Before, decoding data = "null"
let anyType = try? JSONDecoder() // no context
    .decode(AnyType.self, from: data) // This will return .unknownData

// After, decoding data = "null"
let anyType = try? JSONDecoder() // no context
    .decode(AnyType.self, from: data) // This will throw an error

let anyType = try? AnyTypeDecodingContext(rawData: data) // with context
            .inject(to: JSONDecoder())
            .decode(AnyType.self, from: data) // This will return .unknownData
```

### Breaking Change: `AnyTypeDecodingContext.objectFor` is no longer `public`.

All decoding of `AnyType` is now handled internally. So `AnyTypeDecodingContext.objectFor(path: [CodingKey]) throws -> Any` is no longer exposed to users.

### New Features

Added `as<T>(_:)` convenience method and `AnyType` (`anyDictionary`) subscripting for type-safe value extraction:

```swift
// Before
if case let .anyDictionary(myDict) = anyType,
    let title = myDict["title"] as? String {
    // Do a thing
}

// After
let title: String? = anyType["title"]?.as(String.self)
```

### DIFF
--- a/ios/core/Sources/Types/Generic/AnyType.swift
+++ b/ios/core/Sources/Types/Generic/AnyType.swift
@@ -92,7 +92,7 @@ public enum AnyType: Hashable, Sendable {
      */
     case anyArray(data: [AnyType])
 
-    /// The underlying data was not in a known format
+    /// The underlying data was not in a known format. For example, it was a "null".
     case unknownData
 }
 
@@ -238,13 +238,17 @@ extension AnyType: Decodable {
         } else if let number = try? decoder.singleValueContainer().decode(Double.self) {
             self = .number(data: number)
             return
-        } else if decoder.userInfo[AnyTypeDecodingContext.key] is AnyTypeDecodingContext {
-            // Note: Mixed-type collections (anyDictionary and anyArray) cannot be decoded
-            // through standard JSON decoding due to their heterogeneous nature.
-            // Use AnyTypeDecodingContext when decoding JSON containing mixed types.
         }
-        self = .unknownData
-        return
+        
+        // anyArray, anyDictionary, or "null" (which becomes unknownData)
+        guard let context = decoder.userInfo[AnyTypeDecodingContext.key] as? AnyTypeDecodingContext else {
+            throw AnyTypeDecodingError.missingDecodingContext
+        }
+
+        /* Handle mixed-type collections using the context. We could handle every case with this,
+         but it would be slower because of the JSON serialization. So we only use this for anyArray
+         and anyDictionary */
+        self = try context.decode(path: decoder.codingPath)
     }
 }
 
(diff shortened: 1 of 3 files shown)


ios/core/Sources/Types/Generic/AnyType.swift
```diff
@@ -222,42 +300,58 @@ extension AnyType: Encodable {
...
+    }
+
+    /// Decodes the raw JSON data at the given coding path to AnyType
+    func decode(path: [CodingKey]) throws -> AnyType {
+        let obj = try objectFor(path: path)
+        return Self.decode(from: obj)
+    }
+
+    /// Retrieves the object at the given coding path from the raw JSON data
+    private func objectFor(path: [CodingKey]) throws -> Any {
```
KVSRoyal: Todo for self: add this becoming private to release notes with batch of PR review changes.

PR #802, opened 2026-02-12
add render-time and async node error recovery pattern

tmarmer:
- Update `captureError` to fail the player state if no error transition is available.
- Implement render-time error recovery functionality
  - iOS: Errors are handled during deserialization of asset data. Asset render functions do not support throwing so actual render-time issues do not need to be caught
  - Android: Errors are caught for each asset type and bubble up according to their needs.
    - RenderableAsset: Bubbles up to PlayerFragment error handling which calls `PlayerViewModel.fail` which now uses the error controller.
    - SuspendableAsset: Errors caught in the coroutine exception handler are sent to the error controller (except for cancellation exceptions)
    - ComposableAsset: Errors thrown during deserialization are sent to the error controller.
  - React: Top level ErrorBoundary on react player catches and sends errors to the ErrorController.
- Add `PlayerErrorMetadata` interface for errors to implement to include info required for the error controller to capture errors
- Add errors for specific areas where it can be brought to the error controller
  - AssetRenderError updated to include metadata required for the error controller
  - ResolverError added for errors thrown during computeTree 
- Update async-node plugin to make use of the above to call its onAsyncNodeError hook and replace broken content
- Update `JSValueError` to support serialization with JSValue. Update `ErrorState` to use `JSValueError`
- Update J2V8 `.toInvokable` to fix an issue where functions returning arrays of objects would deserialize those objects into a `V8Node` even if a `deserializationStrategy` was available for it.

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

## Release Notes
- Update `AsyncNodePlugin`'s `onAsyncNodeError` hook to depend on the `ErrorController`'s `onError` hook for finding errors.
- Capture render time errors using `ErrorController.captureError` and allow for error recovery on all platforms.
- Add `PlayerErrorMetadata` interface across platforms and ensure serialization of errors keeps additional metadata.
  - When throwing an error that matches this interface, the `ErrorController.captureError` will prefer data from the error object when calling the `onError` hook.

<details>
  <summary>📦 Published PR as canary version: <code>0.15.1--canary.802.31569</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.15.1--canary.802.31569
</details>


plugins/reference-assets/core/src/assets/throwing/types.ts
```diff
@@ -0,0 +1,9 @@
+import type { Asset } from "@player-ui/player";
+
+export interface ThrowingAsset extends Asset<"text"> {
```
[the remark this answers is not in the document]
tmarmer: Found a way to get the errors I need using bad content without introducing a whole new asset so all the throwing stuff is gone.

core/player/src/controllers/error/__tests__/middleware.test.ts
```diff
@@ -88,13 +104,11 @@ describe("ErrorStateMiddleware", () => {
     it("should allow writes when authorized with writeSymbol", () => {
       const binding = parser.parse("errorState");
```
KetanReddy: Should we make this path a const somewhere?

cehan-Chloe (commented):
.

core/player/src/controllers/flow/flow.ts
```diff
@@ -168,6 +168,22 @@ export class FlowInstance {
     return map[key] || map["*"];
   }
 
+  /** Check if the flow has a transition for the given error type in its current state. */
+  public hasTransitionForError(errorType: string): boolean {
```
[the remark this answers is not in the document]
tmarmer: Alternatively, maybe getting rid of `hasTransitionForError` and making `errorTransition` into `tryErrorTransition` that returns a boolean representing if it was successful might be a cleaner approach here.

KVSRoyal (approved):
iOS stuff lgtm---can you just double-check that everything `public` actually needs to be?

jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/JSErrorException.kt
```diff
@@ -2,15 +2,21 @@ package com.intuit.playerui.core.bridge
...
+import com.intuit.playerui.core.player.PlayerExceptionMetadata
 import kotlinx.serialization.builtins.ListSerializer
 import kotlinx.serialization.builtins.serializer
 
 public class JSErrorException(
     override val node: Node,
     cause: Throwable? = node.getSerializable("innerErrors", ListSerializer(ThrowableSerializer()))?.first(),
+    override val type: String = "",
+    override val severity: ErrorSeverity?,
+    override val metadata: Map<String, Any?>?,
```
sugarmanz: I don't think these belong on the base `JSErrorException` type, as that is supposed to mirror the base `Error` TS class. We probably want a subclass that matches the `PlayerErrorMetadata` one

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
```diff
@@ -92,17 +93,19 @@ public class HeadlessPlayer @ExperimentalPlayerApi @JvmOverloads public construc
...
-                    "Exception caught in Player scope: ${throwable.message}",
-                    throwable.stackTrace
-                        .joinToString("\n") {
-                            "\tat $it"
-                        }.replaceFirst("\tat ", "\n"),
-                )
+        coroutineExceptionHandler =
+            config.coroutineExceptionHandler ?: CoroutineExceptionHandler { _, throwable ->
+                if (state !is ReleasedState) {
+                    logger.error("[HeadlessPlayer]: Error has been found")
```
sugarmanz: Consolidate the two logs now — just pull the other one up if you think we should log it every time regardless of whether we can forward it to `captureError`.

jvm/testutils/src/main/kotlin/com/intuit/playerui/utils/test/PlayerTest.kt
```diff
@@ -29,8 +29,13 @@ public abstract class PlayerTest :
     }
 
     /** Helper method for setting a [player] with configurable [plugins] and [runtime] */
-    public fun setupPlayer(plugins: List<Plugin> = this.plugins + this, runtime: Runtime<*> = this.runtime) {
-        player = HeadlessPlayer(plugins, runtime, config = runtime.config)
+    public fun setupPlayer(
+        plugins: List<Plugin> = this.plugins,
+        runtime: Runtime<*> = this.runtime,
+        includeThisInPlugins: Boolean = true,
```
sugarmanz: Rather than an extra param, should we just do this automatically if `this` isn't contained in `plugins`?

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
```diff
@@ -200,16 +203,24 @@ internal class AsyncNodePluginTest : PlayerTest() {
             throw Exception("This is an error message from onAsyncNode")
         }
 
-        val errorMessage = assertThrows<Exception> {
+        val refPlugin = ReferenceAssetsPlugin()
+        refPlugin.apply(runtime)
+        if (player is HeadlessPlayer) {
+            val invokable = (player as HeadlessPlayer).node.getInvokable<Unit>("registerPlugin")
```
sugarmanz: We should just implement `registerPlugin` properly than hack it into a test.

sugarmanz (changes requested):

nancywu1 (approved):
iOS changes looks good, do we have any examples/mocks in the demo app with errorTransitions or error state structure, if not maybe can add them in later?

tmarmer:
> iOS changes looks good, do we have any examples/mocks in the demo app with errorTransitions or error state structure, if not maybe can add them in later?

@nancywu1 The only example I add here is an update to the `chat-ui` mock to show how the error recovery works for async nodes since that's primarily what this PR is for. I know @cehan-Chloe added docs for some of this stuff but I'm unsure if the base `error-controller` branch has any mocks for this case.

cehan-Chloe:
> > iOS changes looks good, do we have any examples/mocks in the demo app with errorTransitions or error state structure, if not maybe can add them in later?
> 
> @nancywu1 The only example I add here is an update to the `chat-ui` mock to show how the error recovery works for async nodes since that's primarily what this PR is for. I know @cehan-Chloe added docs for some of this stuff but I'm unsure if the base `error-controller` branch has any mocks for this case.

I added doc and tests for error controller. There weren't specific storybook/demo app examples for controllers, but since this one added a navigation errorState, so may worth adding it to storybook like [managedPlayer](https://player-ui.github.io/latest/storybook-demo/?path=/story/react-player-managed-player--content-error-handling). Thought?

jvm/core/src/main/kotlin/com/intuit/playerui/core/error/ErrorController.kt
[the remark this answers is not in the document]
sugarmanz: But if the [`error`](https://github.com/player-ui/player/blob/107cbfe940763b898f80139c0e4156d99f98f1f9/jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/JSErrorException.kt#L14-L16) already has this metadata, why do we have a wrapper holding the metadata? Seems like we have conflicting ways of representing the same information.

jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/JSErrorException.kt
```diff
@@ -2,21 +2,15 @@ package com.intuit.playerui.core.bridge
...
 import com.intuit.playerui.core.bridge.serialization.serializers.NodeSerializableField
 import com.intuit.playerui.core.bridge.serialization.serializers.ThrowableSerializer
-import com.intuit.playerui.core.error.ErrorSeverity
 import com.intuit.playerui.core.player.PlayerException
-import com.intuit.playerui.core.player.PlayerExceptionMetadata
 import kotlinx.serialization.builtins.ListSerializer
 import kotlinx.serialization.builtins.serializer
 
-public class JSErrorException(
+public open class JSErrorException(
```
sugarmanz: Let's make this only subclassable by us:
```suggestion
public open class JSErrorException internal constructor(
```

plugins/async-node/jvm/src/test/kotlin/com/intuit/playerui/plugins/asyncnode/AsyncNodePluginTest.kt
sugarmanz: Yeah, we rely on the reference assets/transforms for tests unfortunately. I know @KetanReddy wants to split them out of this repo, so we'll have to address it soon anyways.

android/player/src/main/kotlin/com/intuit/playerui/android/asset/SuspendableAsset.kt
```diff
@@ -52,7 +52,22 @@ public abstract class SuspendableAsset<Data>(
...
+        try {
+            initView(getData()).apply { setTag(R.bool.view_hydrated, false) }
+        } catch (exception: Throwable) {
+            // ignore cancellation exceptions because those are used to rehydrate the view
+            if (exception is CancellationException) {
+                throw exception
+            }
+
+            if (exception is AssetRenderException) {
+                exception.assetParentPath += assetContext
```
brocollie08, line 65: i'm a bit late here, this looks like it's supposed to bubble up from the asset level that threw the original error to show the full path? Does this mean the final message is going to be in reverse order from the originating asset?

android/player/src/main/kotlin/com/intuit/playerui/android/compose/ComposableAsset.kt
```diff
@@ -82,9 +98,12 @@ public abstract class ComposableAsset<Data>(
         styles: AssetStyle? = null,
         tag: String? = null,
     ) {
-        val assetTag = tag ?: asset.id
```
brocollie08, line 85: what's this tag logic change

jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/serialization/serializers/ThrowableSerializer.kt
```diff
@@ -72,29 +81,53 @@ public class ThrowableSerializer : KSerializer<Throwable> {
                     message = decodeNullableSerializableElement(descriptor, 1, String.serializer().nullable) ?: ""
                     stackTrace = decodeSerializedStackTrace()
                     cause = decodeNullableSerializableElement(descriptor, 4, nullable)
+                    type = decodeNullableSerializableElement(descriptor, 5, String.serializer().nullable)
+                    severity = decodeNullableSerializableElement(descriptor, 6, ErrorSeverity.serializer().nullable)
+                    metadata = decodeNullableSerializableElement(descriptor, 7, GenericSerializer()) as? Map<String, Any>
```
brocollie08, line 88: is the GenericSerializer okay here? this doesn't need a MapSerializer?

jvm/hermes/src/test/kotlin/com/intuit/playerui/jsi/serialization/serializers/ThrowableSerializerTest.kt
```diff
@@ -32,6 +47,31 @@ internal class ThrowableSerializerTest : HermesTest() {
...
+        exception as JSErrorExceptionWithMetadata
+        Assertions.assertEquals("Error: hello", exception.message)
+        Assertions.assertEquals(
+            """com.intuit.playerui.core.bridge.JSErrorExceptionWithMetadata: Error: hello
+""",
+            exception.stackTraceToString(),
+        )
+        Assertions.assertEquals("ErrorType", exception.type)
+        Assertions.assertEquals(ErrorSeverity.ERROR, exception.severity)
+        Assertions.assertEquals(mapOf("testProperty" to "testValue"), exception.metadata)
```
brocollie08, line 72: following previous comment, i'm a bit surprised this is passing, @sugarmanz what am i missing

android/player/src/main/kotlin/com/intuit/playerui/android/compose/ComposableAsset.kt
tmarmer: The code here was originally calling `withTag` on the asset context even if no `tag` was provided. This change was meant to prevent that, but I didn't want to change the `.testTag` call to avoid breaking tests. This was something @sugarmanz and I noticed when I was testing earlier iterations where the I was matching assets based on the `assetContext.id` but was getting the asset id duplicated in the string. This is out of scope of the PR though so if there are issues with this I can revert it and we can look into this more in depth another time.

jvm/core/src/main/kotlin/com/intuit/playerui/core/bridge/serialization/serializers/ThrowableSerializer.kt
tmarmer: I think the `GenericSerializer` was picking the `MapSerializer` for me. Wasn't aware of that specific serializer when I first put this together so I'll go make that change.

android/player/src/main/kotlin/com/intuit/playerui/android/asset/SuspendableAsset.kt
tmarmer: The list will be in reverse-order of the actual view tree, but this is used to create the error message [here](https://github.com/player-ui/player/blob/56881b6e8c9b42cc544828eb9e47db07263ad82d/android/player/src/main/kotlin/com/intuit/playerui/android/asset/AssetRenderException.kt#L18-L24). The reverse order lets us list the most relevant asset first and list where it and its parents were found in.

jvm/hermes/src/test/kotlin/com/intuit/playerui/jsi/serialization/serializers/ThrowableSerializerTest.kt
tmarmer: As per my other comment, I think `GenericSerializer` was picking the `MapSerializer` for me. I've updated `ThrowSerializer` to use `MapSerializer` instead.

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
+                       return {
+                           result: fn(...args),
+                           success: true
+                       }
                    } catch(e) {
-                       return e
+                       return {
+                           result: e,
+                           success: false
+                       }
```
tmarmer, line 39: This change is to ensure that all exceptions are handled as exceptions. Previously if the executed code did something like `throw false` our code wouldn't be able to differentiate between that and `return false` despite it being thrown. Using an object here lets us tell the difference and extract the result as needed.

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
+        let type = value.objectForKeyedSubscript("type")
+        if (type?.isString != true) {
+            return JSValueError.simpleJsError(jsError: value, message: message!.toString()!)
+        }
+        
+        let severity = ErrorSeverity(rawValue: value.objectForKeyedSubscript("severity")?.toString() ?? "")
+        let metadata = value.objectForKeyedSubscript("metadata")?.toDictionary() as? [String: Any]
+        return JSValueError.errorWithMetadata(jsError: value, message: message!.toString()!, type: type!.toString()!, severity: severity, metadata: metadata)
+    }
+}
```
tmarmer: Not sure if this is the best way to support deserializing from `JSValue`. Could use some advice here.

ios/core/Tests/HeadlessPlayerTests.swift
```diff
@@ -210,9 +216,9 @@ class HeadlessPlayerTests: XCTestCase {
                 XCTFail("should have failed")
             case .failure(let error):
                 switch error {
-                case .promiseRejected(let errorState):
-                    XCTAssertEqual(errorState.error, "undefined is not an object (evaluating \'o.navigation\')")
-                default: break
+                case .jsConversionFailure: break
+                default:
+                    XCTFail("Should throw PlayerError.jsConversionFailure")
```
tmarmer, line 216: This change looks breaking, but this test was previously passing because of the `default: break`. Caught this when the new `JSValueError` had a mistake in the message and it wasn't causing this test to fail. This check [here](https://github.com/player-ui/player/blob/6222f07d6fbf274ceb6ecd11a094456524557841/ios/core/Sources/Player/HeadlessPlayer.swift#L263) is what causes this to become a `jsConversionFailure` unlike the test above.

ios/core/Sources/Types/Core/CompletedState.swift
```diff
@@ -244,12 +244,13 @@ public class ErrorState: BaseFlowState, PlayerFlowExecutionData {
     public static func createInstance(from value: JSValue?) -> ErrorState? {
         guard
             let flow = value?.objectForKeyedSubscript("flow"),
-            let message = value?.objectForKeyedSubscript("error")?.objectForKeyedSubscript("message")?.toString()
+            let err = value?.objectForKeyedSubscript("error")
```
JunDangIntuit, line 247: Could we put all keys in an enum, not string literals in the code. I've noticed this pattern across several iOS files (CompletedState.swift, NavigationStates.swift, PlayerControllers, etc.): like objectForKeyedSubscript("flowResult"), objectForKeyedSubscript("controllers") etc. A typo in any of these would silently return undefined with no compiler warning. If it's out of scope for this PR, I'd be happy to take on a follow-up PR to address it.

ios/core/Sources/utilities/JSUtilities.swift
```diff
@@ -63,7 +63,21 @@ public class JSUtilities {
 
 internal extension JSContext {
     func error<E>(for error: E) -> JSValue? where E: Error, E: JSConvertibleError {
-        objectForKeyedSubscript("Error").construct(withArguments: [error.jsDescription])
+        if let jsError = error as? JSValueError {
+            // If the error originated in JS, just return the original object
+            return jsError.getJSErrorObject()
+        }
+        
+        let errObj = objectForKeyedSubscript("Error").construct(withArguments: [error.jsDescription])
```
JunDangIntuit: The key "Error" in objectForKeyedSubscript("Error") (the JS global Error constructor) vs "error" in objectForKeyedSubscript("error") (a property on the state object) are easy to confuse at a glance. As a newcomer to the codebase, I had to look them up to confirm they're intentionally different.

Would it be worth extracting JS keys into named constants? For example:

private enum JSGlobal {
    static let error = "Error"  // JS Error constructor
}
This way, objectForKeyedSubscript(JSGlobal.error).construct(...) makes the intent immediately clear without needing to know the JS convention. Not a blocker — just a readability thought for future contributors.

ios/core/Sources/utilities/JSUtilities.swift
```diff
@@ -63,7 +63,21 @@ public class JSUtilities {
...
+        if let errorWithMetadata = error as? ErrorWithMetadata, let err = errObj, errorWithMetadata.hasMetadata {
+            err.setValue(errorWithMetadata.type, forProperty: "type")
```
JunDangIntuit: may you define a struct Constant like 
struct SomeName {
static let type = "type"
static let severity = "severity"
static let metadata = "metadata"
}

ios/core/Sources/utilities/JSUtilities.swift
```diff
@@ -72,3 +86,10 @@ public protocol JSConvertibleError {
     /// The description to use when send to JavaScriptCore
     var jsDescription: String { get }
 }
+
+public protocol ErrorWithMetadata : Error {
+    var hasMetadata: Bool { get }
+    var type: String { get }
```
JunDangIntuit, line 103: Nit: since errorType is a String to allow plugin-defined custom types, consider adding a doc comment on captureError(errorType:) noting that callers should prefer ErrorTypes constants when applicable, but custom strings are accepted for plugin-specific error types.

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
+            return jsError
+        }
+    }
+}
+
+extension JSValueError: CreatedFromJSValue {
+    public static func createInstance(value: JSValue) -> JSValueError {
+        if (!value.isInstance(of: value.context.objectForKeyedSubscript("Error"))) {
+            return JSValueError.unknownError(jsError: value)
+        }
```
KVSRoyal: nit: Swift doesn't use brackets around the conditionals. E.g. 
```swift
if !value.isInstance(of: value.context.objectForKeyedSubscript("Error")) { }
```

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
 /**
  Represents the different errors that occur when evaluating JSValue
  */
-public enum JSValueError: Error, Equatable {
-    case thrownFromJS(message: String)
-}
\ No newline at end of file
+public enum JSValueError: Error {
+    case simpleJsError(jsError: JSValue, message: String)
+    case errorWithMetadata(jsError: JSValue, message: String, type: String, severity: ErrorSeverity?, metadata: [String: Any]?)
```
JunDangIntuit: errorWithMetadata has too many associated values.  A better pattern would be to put these values into a struct. keep jsError: JSValue, message: String, and then type: String, severity: ErrorSeverity?, metadata: [String: Any]? into a struct.

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
+        
+        let message = value.objectForKeyedSubscript("message")
+        if (message?.isString != true) {
+            return JSValueError.unknownError(jsError: value)
+        }
+        
+        let type = value.objectForKeyedSubscript("type")
+        if (type?.isString != true) {
+            return JSValueError.simpleJsError(jsError: value, message: message!.toString()!)
+        }
```
KVSRoyal: Can you use a `guard let` here and above (for `message`)? It fits the pattern better (bail out of happy path) and allows us to remove the force-unwraps (i.e. the `!`). E.g.

```swift
guard let type = value.objectForKeyedSubscript("type"), 
    type.isString, 
    let typeAsString = type.toString() 
else {
    return .simpleJsError(jsError: value, message: message.toString())
}

// You can now use type and typeAsString safely down here. The compiler will know they are not nil
```

We should never use the force-unwrap operator in production code. If unwrapping fails, the app will crash, and that's typically undesirable. We instead use `guard let` and `if let` to confirm something is not nil.

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
+    
+    func getJSErrorObject() -> JSValue {
+        switch self {
+        case .simpleJsError(let jsError, _):
+            return jsError
+        case .errorWithMetadata(let jsError, _, _, _, _):
+            return jsError
+        case .unknownError(let jsError):
+            return jsError
+        }
```
JunDangIntuit: since you only return jsError, you could simplify it to:
func getJSErrorObject() -> JSValue {
    switch self {
    case .simpleJsError(let jsError, _),
         .errorWithMetadata(let jsError, _, _),
         .unknownError(let jsError):
        return jsError
    }
}

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
+        
+        let message = value.objectForKeyedSubscript("message")
+        if (message?.isString != true) {
+            return JSValueError.unknownError(jsError: value)
+        }
+        
+        let type = value.objectForKeyedSubscript("type")
+        if (type?.isString != true) {
+            return JSValueError.simpleJsError(jsError: value, message: message!.toString()!)
```
JunDangIntuit: let's not use force unwrap as it is fragile even though you believe it will not be nil:
guard let message = message?.toString() else {
    return JSValueError.unknownError(jsError: value)
}
return JSValueError.simpleJsError(jsError: value, message: messageString)

please use this pattern for all force unwrap in this file

ios/core/Sources/utilities/JSValue+Extensions.swift
KVSRoyal: You could maybe try this to decode directly from `JSValue`? [Like this](https://github.com/player-ui/player/blob/6222f07d6fbf274ceb6ecd11a094456524557841/ios/core/Sources/Types/Hooks/Hook.swift#L161-L171).

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
+    
+    public var metadata: [String: Any]? {
+        switch self {
+        case .errorWithMetadata(_, _, _, _, let metadata):
+            return metadata
+        default:
+            return nil
+        }
+    }
+}
```
JunDangIntuit, line 117: May you refactor all  public var hasMetadata: Bool { },  public var type: String { },  public var severity: ErrorSeverity? { } and  public var metadata: [String: Any]? after you put type: String, severity: ErrorSeverity?, metadata: [String: Any]? into a struct. It should save many lines of codes

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
+        }
+    }
+    
+    public var type: String {
+        switch self {
+        case .errorWithMetadata(_, _, let type, _, _):
+            return type
+        default:
+            return "UNKNOWN"
+        }
```
JunDangIntuit: "UNKNOWN" for non-metadata cases — should this be an ErrorTypes constant? like add ErrorTypes.unknown.

ios/core/Sources/utilities/JSValue+Extensions.swift
JunDangIntuit: this should work, but for sure without force wrapping.

JunDangIntuit: @KVSRoyal  I think the type of JSValue will not work though as it does not conform to Decodable

KVSRoyal: Typically it does not, yes, but [we actually apparently have an extension to handle decoding JSValues](https://github.com/player-ui/player/blob/6222f07d6fbf274ceb6ecd11a094456524557841/ios/core/Sources/Types/Assets/BaseAssetRegistry.swift#L290-L296).

ios/core/Tests/utilities/JSValueExtensionsTests.swift
```diff
@@ -24,11 +25,80 @@ class JSValueExtensionsTests: XCTestCase {
...
+        })
+    }
+    
+    /// Test case for errors thrown with error controller metadata. Check that metadata is preserved
+    func testTryCatchThrowsErrorWithMetadata() {
+
+        let functionReturningError = self.context
+            .evaluateScript("""
+                             (() => {
+                                const err = new Error("Fail")
```
JunDangIntuit: Nit: it'd be better not to use abbreviation. not sure how Javscript, but in swift file, a full name is preferred.

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,43 +17,128 @@ extension JSValue {
...
             self.context.evaluateScript(
             """
                (fn, args) => {
                    try {
-                       return fn(...args)
+                       return {
+                           result: fn(...args),
+                           success: true
+                       }
                    } catch(e) {
```
JunDangIntuit: Nit: since this JS is read primarily by Swift developers, catch(error) would be clearer than catch(e).

ios/core/Tests/HeadlessPlayerTests.swift
JunDangIntuit: you have a good catch

ios/core/Sources/utilities/JSUtilities.swift
tmarmer: I like this idea a lot! I'm wondering what you think of taking it maybe one step further and adding a utility to shortcut constructing known JS classes from the `JSContext`. Thinking something like:

```swift
internal enum JSClass: String {
    case Error = "Error"
}

internal extension JSContext {
    func getJSClass(_ name: JSClass) -> JSValue? {
        objectForKeyedSubscript(name)
    }

    func constructClass(_ name: JSClass, withArguments: [Any]?) -> JSValue? {
        getClass(name).construct(withArguments: withArguments)
    }
}
```

It's definitely on the simpler side of utilities but imo it helps clear up the intent of the enum since I don't think it's immediately obvious that the use is meant for `objectForKeyedSubscript` on the `JSContext` itself.

ios/core/Sources/utilities/JSValue+Extensions.swift
tmarmer: Removed the `"UNKNOWN"` case. The fallback now is to use an empty string to represent not having a known error type. This should only happen for errors that don't follow our shared interface so I think it's okay to keep this out of `ErrorTypes` enum imo.

ios/core/Sources/Types/Core/CompletedState.swift
tmarmer: I'll make the keys for this object into an enum, but could you make a github issue to follow up and remove this across the repo? I think it's a good practice we should try to keep up with

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -9,6 +9,11 @@ import Foundation
 import JavaScriptCore
 
 extension JSValue {
+    
+    public enum TryCatchResultKeys {
+        static let success = "success"
+        static let result = "result"
+    }
```
KVSRoyal: 1. Nit: Could this conform to `String` just like the `JSClass` enum. It's also weird this enum doesn't have any cases 😅 so it would be a good change in that regard too. E.g.

```swift
public enum TryCatchResultKeys: String {
    case success, result
}
```

2. Does this need to be `public`? They seem like an internal implementation detail. If they don't need to be public, can we remove the access modifier (will default to `internal`)?

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -17,128 +22,102 @@ extension JSValue {
...
+    public let metadata: [String: Any]?
+    
+    public let originalJSError: JSValue
+    
+    public enum JSKeys {
+        static let message = "message"
+        static let type = "type"
+        static let severity = "severity"
+        static let metadata = "metadata"
     }
```
KVSRoyal: Same comment here about the enum. Can this conform to `String` and have cases?

KVSRoyal (commented):
Just a few really small comments. Otherwise lgtm

ios/core/Sources/utilities/JSValue+Extensions.swift
```diff
@@ -23,37 +27,93 @@ extension JSValue {
...
+        }
+        
+        severity = ErrorSeverity(rawValue: jsErrorObject.objectForKeyedSubscript(JSKeys.severity)?.toString() ?? "")
+        metadata = jsErrorObject.objectForKeyedSubscript(JSKeys.metadata)?.toDictionary() as? [String: Any]
+    }
+}
+
+extension JSValueError: ErrorWithMetadata {
+    public var hasMetadata: Bool {
+        !type.isEmpty
```
JunDangIntuit: is it correct for hasMetadata to check type is empty?

ios/core/Sources/utilities/JSUtilities.swift
```diff
@@ -72,3 +98,10 @@ public protocol JSConvertibleError {
     /// The description to use when send to JavaScriptCore
     var jsDescription: String { get }
 }
+
+public protocol ErrorWithMetadata : Error {
+    var hasMetadata: Bool { get }
```
JunDangIntuit: hasMetadata is purely derivable from metadata != nil. It should not to be a protocol requirement

ios/core/Sources/utilities/JSValue+Extensions.swift
tmarmer: Because the `ErrorWithMetadata` requires the `type` property, it's the simplest identifier we have for errors coming from JS. I could alternatively make `hasMetadata` a property and set its value during `init` based on the existence of `type` coming from `JS`. This would allow empty strings as a valid `type` value and make the condition a little more. Would that work better here?

ios/core/Sources/utilities/JSUtilities.swift
tmarmer: Maybe the property name is incorrect. I need a way (at least on the `JSValueError` to declare whether or not the error actually conforms to the shared interface error with metadata interface. The only required property for that interface is the `type` property (which is why I check if it's empty on `JSValueError`) but the others are optional. I figured having this extra property would also allow `enum` errors to only have the additional data supported on some error types.

ios/core/Sources/Types/Hooks/Hook.swift
```diff
@@ -82,32 +82,25 @@ public class Hook<T>: BaseJSHook where T: CreatedFromJSValue {
 }
 
 /**
- This class represents a SyncBailHook in the JS runtime that can be tapped into
- to receive JS events and optionally return a value to bail/stop further execution
+ This class represents a JS Hook with the ability to return a result. This can work for a Waterfall or Bail hook.
  */
-public class BailHook<T>: BaseJSHook where T: CreatedFromJSValue {
+public class HookWithResult<T, R>: BaseJSHook where T: CreatedFromJSValue, R: Encodable {
```
JunDangIntuit: May you share the reason to add Encodable here?

ios/core/Sources/utilities/JSUtilities.swift
JunDangIntuit: I think hasMetadata should not be in the protocol. you could probably do this:
  ```
  func error<E>(for error: E) -> JSValue? where E: Error, E: JSConvertibleError {
        if let jsValueError = error as? JSValueError {
            return jsValueError.originalJSError
        }
        
        let errObj = constructClass(.Error, withArguments: [error.jsDescription])
        if let errorWithMetadata = error as? ErrorWithMetadata, let err = errObj { /// this is checking if it is nil
            if !errorWithMetadata.type.isEmpty {  /// this is checking if the type is empty
                err.setValue(errorWithMetadata.type, forProperty: JSValueError.JSKeys.type)
            }
            if let severity = errorWithMetadata.severity {
                err.setValue(severity.rawValue, forProperty: JSValueError.JSKeys.severity)
            }
            if let metadata = errorWithMetadata.metadata {
                err.setValue(metadata, forProperty: JSValueError.JSKeys.metadata)
            }
        }
```
A separate flag that can drift out of sync with the actual properties.  In another place
```
 extension AssetRenderError: ErrorWithMetadata {
    public var hasMetadata: Bool {
        true
    }
```
 it's boilerplate forced by the protocol that adds no value. The protocol shouldn't require conformers to implement a property that's always a constant or always derivable from other properties already in the protocol.

File: ios/core/Sources/Types/Hooks/Hook.swift (hunk 33 above)
Comment (tmarmer) [reply ios/core/Sources/Types/Hooks/Hook.swift]: This file is for tapable hooks that interact with the JS layer. It's possible that the result is read by some core JS code so unless I misunderstand how the integration works, it would need to be `Encodable` in order to actually work.

File: ios/core/Sources/utilities/JSUtilities.swift (hunk 32 above)
Comment (tmarmer) [reply ios/core/Sources/utilities/JSUtilities.swift]: Alright, makes sense to me. I think I could skip the `isEmpty` check within the `error` function since it really is just an edge case for JS errors coming from the core layer. In those cases the first check that returns the original object anyway means we don't even really need to care when serializing the error since anybody else conforming to that protocol should be provided some kind of recognizable string value anyway. I'll still leave a `isErrorWithMetadata` on `JSValueError` itself in case we need some way to tell whether or not there is any meaningful metadata on the error (like if somebody is tapping into the `onError` hook on the error controller)

File: ios/core/Sources/Types/Hooks/Hook.swift (hunk 33 above)
Comment (JunDangIntuit) [reply ios/core/Sources/Types/Hooks/Hook.swift]: I think you might not need Encodable. Encodable implies encode(to:) is used for JS serialization, but it isn't — JavaScriptCore bridges via Obj-C (NSNumber, NSDictionary, etc.).
aka: When a @convention(block) closure returns a value to JavaScriptCore, it doesn't know about Swift's Codable system. It uses Objective-C bridging.

Comment (tmarmer) [reply ios/core/Sources/Types/Hooks/Hook.swift]: Interesting! I can remove the `Encodable` restriction, but is there some other kind of type or protocol we can use here to ensure people using this hook only return something that works with that? Trying to avoid cases where people return some complex object that ends up throwing or losing information as it goes to JS.

File: ios/core/Sources/utilities/JSValue+Extensions.swift (hunk 29 above)
Comment (tmarmer) [reply ios/core/Sources/utilities/JSValue+Extensions.swift]: With regards to (1) I've been looking into how to better store these strings since using an a `String` `enum` comes with its own issues. It ends up being a bit error-prone in that most the functions requiring these string keys are using an `Any` for the args type, so any time I miss using `.rawValue` on an enum it causes things to break, usually with something getting constructed incorrectly. All I really care about here is having the string constants, the `enum` was just a convenient way to namespace these constants. From what I could gather online, using something like a `struct` to namespace the values instead makes the type still constructible, which a caseless `enum` wouldn't be.

If we don't want caseless enums, I could change this to a struct with constants in it or we could come up with some other pattern for constant strings so we don't need to use the `.rawValue` on enum cases that otherwise don't benefit from being treated as an `enum`.

File: jvm/j2v8/src/main/kotlin/com/intuit/playerui/j2v8/extensions/HandleValue.kt [hunk 34]
```diff
@@ -17,22 +18,26 @@ import com.intuit.playerui.j2v8.v8Array
...
+            V8.getUndefined() -> null
+            is V8Primitive -> value
+            is V8Function -> toInvokable<Any?>(format, this, format.serializer())
+            is V8Array -> deserializationStrategy?.let { format.decodeFromRuntimeValue(deserializationStrategy, this) } ?: toList(format)
+            is V8Object -> deserializationStrategy?.let { format.decodeFromRuntimeValue(deserializationStrategy, this) } ?: toNode(format)
+            else -> null
+        }
+    }
+
+internal fun Any?.handleValue(format: RuntimeFormat<V8Value>): Any? = handleValue<Any>(format, null)
```
brocollie08: you might not need this if you give the new one with the new arg a default null

jvm/j2v8/src/test/kotlin/com/intuit/playerui/j2v8/extensions/InvokeTest.kt
```diff
@@ -24,4 +34,26 @@ internal class InvokeTest : J2V8Test() {
...
     }
+
+    @Test
+    fun `test nested serialization`() {
+        v8.evaluateInJSThreadBlocking(runtime) {
+            val func = v8.executeObjectScript("""(function(){return [{ "testProp": "testValue" }]})""") as V8Function
+            val invokable = func.toInvokable(format, v8Object, ListSerializer(TestSerializableClass.serializer()))
+
+            assertNotNull(invokable)
+            assertEquals(arrayListOf(TestSerializableClass("testValue")), invokable.invoke())
```
brocollie08: nit - listOf doesn't work here?

android/player/src/main/kotlin/com/intuit/playerui/android/asset/AssetRenderException.kt
```diff
@@ -1,9 +1,17 @@
...
+import com.intuit.playerui.core.bridge.serialization.serializers.ThrowableSerializer
+import com.intuit.playerui.core.error.ErrorSeverity
+import com.intuit.playerui.core.error.ErrorTypes
 import com.intuit.playerui.core.player.PlayerException
+import com.intuit.playerui.core.player.PlayerExceptionMetadata
+import kotlinx.serialization.Serializable
 
-class AssetRenderException : PlayerException {
+@Serializable(ThrowableSerializer::class)
+class AssetRenderException :
```
brocollie08, line 12: this can't be PlayerExceptionWithMetadata?

ios/core/Sources/utilities/JSValue+Extensions.swift
KVSRoyal: Oh, I see. Okay, in that case would this work? I like `extension String`---very versatile and wherever the String type is expected you can then just do `.success` instead of `String.success`.

```swift
private extension String {
    static let success = "success"
    static let result = "result"
}

foo(someAnyParam: String.success)
bar(someStringParam: .success)
```

android/player/src/main/kotlin/com/intuit/playerui/android/asset/AssetRenderException.kt
tmarmer: It could, but I was thinking `PlayerExceptionWithMetadata` should just have an `internal constructor` on it and leave it non-`open` to prevent extension since I don't like the idea of having a version of this error constructible with any combination of type/severity/metadata. I would prefer to remove `PlayerExceptionWithMetadata` in the long term if we could add some kind of polymorphic serialization pattern for `Throwable`s.

That said, I can go either way on this implementation, so let me know what you think works best for now.

ios/core/Sources/utilities/JSValue+Extensions.swift
tmarmer: This would definitely work for my case here, tho losing the namespacing of having it contained in something like `TryCatchResultKeys` or `JSKeys` does hurt showing the intention of these constants. I can switch to this approach if it's preferred though.

I did copy the original `enum` with `static let` constants from [this comment](https://github.com/player-ui/player/pull/802#discussion_r3028454746) from @JunDangIntuit though, so I'd also like to get her opinion here in case there was a reason for suggesting this approach.

sugarmanz (approved):
Looks much better now — thanks for the extra effort to settle on strategy and fix serialization disparity.

ios/core/Sources/Types/Hooks/Hook.swift
JunDangIntuit: Yes, I think you can create a marker protocol and allow only simple types can pass

public protocol SomeName {}

extension Bool: SomeName {}
extension Int: SomeName {}
extension Double: SomeName {}
extension Float: SomeName {}
extension String: SomeName {}
extension JSValue: SomeName {}
extension NSNumber: SomeName {}
extension NSString: SomeName {}

And then replace Encodable with SomeName

JunDangIntuit: Encodable is not correct to use here as any struct in Swift can be Encodable, like below is also correct:
struct SomeResult: Encodable {
    let name: String
    let count: Int
}
so when you have HookWithResult<JSValueError, SomeResult>, it will compile as well, nothing wrong, but could produce broken data.

ios/core/Sources/utilities/JSValue+Extensions.swift
JunDangIntuit: oh, yes, sorry about my previous confusion of enum. 
either 
private enum TryCatchResultKeys: String {
    case success, result
}
or 
private struct TryCatchResultKeys {
    static let success = "success"
    static let result = "result"
}

I'd lean toward a struct over extension String here. Even with private, adding .success and .result to String muddies the namespace — anywhere Swift infers a String type in that file, these show up as valid values (e.g. let name: String = .success compiles), which can be confusing. A struct keeps the namespacing explicit.

ios/core/Sources/Types/Hooks/Hook.swift
tmarmer: Makes sense. I've removed `Encodable` for now but didn't want to apply the marker protocol as I think we'll want to standardize an approach that works for all hooks in this file and didn't want to pull that into the scope of this PR.

ios/core/Sources/utilities/JSValue+Extensions.swift
tmarmer: As per other thread, made this a struct.

cehan-Chloe (approved):

JunDangIntuit (approved):

PR #817, opened 2026-03-13
`ExternalStatePlugin` allows overrides

KVSRoyal:
This is now the jumbo PR covering all 3 platforms. (Was previously just for core.)

## What

Please review [the release notes](https://player-ui.github.io/pr/817/announcements/player-one-dot-0/#externalactionplugin-renamed-to-externalstateplugin-with-new-handler-api) that are part of the PR. (Putting them there helps me keep redundancy down and keep the docs accurate.)

## Why
**Deterministic Overrides**: Users want predictable behaviour when multiple plugins or handlers are registered. The registry ensures that:

* More specific matchers always take precedence (e.g., { ref: "x", type: "y" } beats { ref: "x" })
* Later registrations replace earlier ones with the same specificity
* Behaviour is consistent regardless of registration order

**Existing Patterns**: We chose to reuse `Registry` as this follows established patterns in Player for Asset registration. Hopefully this will make it easier for users to understand the logic.

**Visibility**: We chose to add `debug` logging when handlers are overridden to make it clear to developers what's happening during plugin composition.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A` (part of 1.0)

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

# Release Notes
### `ExternalActionPlugin` renamed to `ExternalStatePlugin` with new handler API

**The plugin has been renamed** from `ExternalActionPlugin` to `ExternalStatePlugin`. The name `ExternalState` better reflects what the plugin does — it handles `EXTERNAL` navigation states — and is now used consistently across all platforms.

**The handler API has changed.** Instead of a single function that receives all external states, you now provide an array of handlers. Each handler requires a `ref` (matching the external state's `ref` field) and a `handlerFunction`. An optional `match` object can be provided to match on additional state properties beyond `ref`. See docs site for full examples.

#### Handler matching and specificity

Handlers are matched against the external state using partial object matching on `ref` and any additional properties in `match`. When multiple handlers could match a state, the most specific one wins (the one with the most matching properties).

This allows you to register a general handler for a `ref` and a more specific one for a particular variant:

```typescript
new ExternalStatePlugin([
  // Matches any state with ref: "my-action"
  { ref: "my-action", handlerFunction: () => "default" },
  // Takes precedence when type: "special" is also present
  {
    ref: "my-action",
    match: { type: "special" },
    handlerFunction: () => "special",
  },
]);
```

#### Overriding handlers across plugins

Overriding handlers is now supported. When multiple `ExternalStatePlugin` instances register a handler for the same state, the **last registered handler wins**. A `debug` log is emitted when a handler is replaced, so you can tell when an override is happening.

To add more handlers after initial setup, register a new `ExternalStatePlugin` instance. Plugins are typically grouped together and easy to find by name. This makes it straightforward to determine the override order.

We considered allowing more handlers to be added to an `ExternalStatePlugin` after creation, however, this would make it more difficult to figure out the override order. So we do not support that.

#### Missing handlers

For now, If no handler matches an external state, the Player remains on that state and no transition occurs. Ensure every `EXTERNAL` state in your flow has a registered handler.

> [!NOTE]
> In a future release, unhandled external states will automatically report an error via the Error Controller.


KetanReddy:
Maybe out of scope for this PR but `sorted-array` is deprecated so we might want to replace it?

core/partial-match-registry/src/index.ts
```diff
@@ -4,6 +4,11 @@ import { createObjectMatcher } from "./deep-partial-matcher";
 
 export { createObjectMatcher } from "./deep-partial-matcher";
 
+/** Minimal logger interface for registry operations */
+interface Logger {
+  debug?: (...args: Array<any>) => void;
+}
```
KetanReddy: Maybe out of scope for this but we should probably centralize logger definitions for the Player ecosystem.

core/partial-match-registry/src/index.ts
```diff
@@ -26,42 +31,122 @@ interface RegistryIndex<V> {
...
+ * const registry = new Registry<string>();
+ * registry.set({ type: "action" }, "generic");
+ * registry.set({ type: "action", id: "123" }, "specific");
+ *
+ * registry.get({ type: "action" }); // Returns "generic"
+ * registry.get({ type: "action", id: "123" }); // Returns "specific" (more specific match)
+ * registry.get({ type: "action", id: "123", extra: "data" }); // Returns "specific" (fuzzy match)
+ * ```
+ *
+ * @param V - The type of values stored in the registry
```
KetanReddy: Nit: Do we need this level of docs in the docstring for the class? Can we add this to the doc site and link out to it?

core/partial-match-registry/src/index.ts
```diff
@@ -26,42 +31,122 @@ interface RegistryIndex<V> {
...
-  constructor(initialSet?: Array<[any, V]>) {
+  /**
+   * Creates a new Registry instance
+   *
+   * @param initialSet - Optional array of [key, value] tuples to populate the registry
+   * @param logger - Optional logger instance for debug logging (e.g., when entries are replaced)
+   */
+  constructor(
+    initialSet?: Array<[any, V]>,
+    private logger?: Logger | undefined,
```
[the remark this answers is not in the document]
KVSRoyal: Yeah, for sure.

core/partial-match-registry/src/index.ts
```diff
@@ -26,42 +31,122 @@ interface RegistryIndex<V> {
...
+  /**
+   * Add match -> value mapping to the registry
+   *
+   * If an entry with the same specificity and matching key already exists, it will be replaced
+   * and a debug log will be emitted (if a logger is configured).
+   *
+   * @param match - The key to match against (can be a primitive or object)
+   * @param value - The value to associate with this key
+   */
+  set(match: any, value: V): void {
```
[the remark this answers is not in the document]
KVSRoyal: Yeah, definitely a good idea. Will try. I might need to go annoy @tmarmer to get advice on it.

core/partial-match-registry/src/index.ts
```diff
@@ -27,17 +32,35 @@ interface RegistryIndex<V> {
 /**
  * A partial match registry is a map that uses an object to "match" against keys.
  * More specific matches take precedence over less specific ones.
+ *
+ * @param V - The type of values stored in the registry
```
spentacular: No need to add param since typescript handles this.

KVSRoyal:
@claude Review this PR

intuit-svc:
## Docs Preview

A preview of your PR docs was deployed by CircleCI [#34847](https://circleci.com/gh/player-ui/player/34847) on `Mon, 06 Apr 2026 22:55:51 GMT`

### 📖 Docs ([View site](https://player-ui.github.io/pr/817/))
- [announcements/player-one-dot-0](https://player-ui.github.io/pr/817/announcements/player-one-dot-0)
- [assets/reference](https://player-ui.github.io/pr/817/assets/reference)
- [content/navigation](https://player-ui.github.io/pr/817/content/navigation)
- [getting-started](https://player-ui.github.io/pr/817/getting-started)
- [guides/multi-flow-experiences](https://player-ui.github.io/pr/817/guides/multi-flow-experiences)
- [guides/writing-plugins](https://player-ui.github.io/pr/817/guides/writing-plugins)
- [platforms/android](https://player-ui.github.io/pr/817/platforms/android)
- [plugins/core/common-expressions](https://player-ui.github.io/pr/817/plugins/core/common-expressions)
- [plugins/core/common-types](https://player-ui.github.io/pr/817/plugins/core/common-types)
- [plugins/core/computed-properties](https://player-ui.github.io/pr/817/plugins/core/computed-properties)

JunDangIntuit (approved):

plugins/external-state/core/src/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    this.registerHandlers(player);
+
+    // Only the first instance should tap the hooks to avoid redundant taps
+    if (!isFirstInstance) {
+      return;
+    }
+
+    player.hooks.flowController.tap(this.name, (flowController) => {
+      flowController.hooks.flow.tap(this.name, (flow) => {
+        flow.hooks.transition.tap(this.name, (fromState, toNamedState) => {
```
spentacular: We either need to port these over or add a followup to get rid of the `setTimeout`: https://github.com/player-ui/player/blob/main/plugins/external-action/core/src/index.ts#L39

docs/site/src/content/docs/announcements/player-one-dot-0.mdx
```diff
@@ -180,9 +273,198 @@ targets: [
 ]
 ```
 
+### `ExternalActionPlugin` renamed to `ExternalStatePlugin` with new handler API
```
sugarmanz, line 276: Can these be added to the PR `# Release Notes` section so they get picked up and added to the `CHANGELOG.md`?

docs/site/src/content/docs/announcements/player-one-dot-0.mdx
```diff
@@ -180,9 +273,198 @@ targets: [
...
+
+// After — ExternalStatePlugin
+// Separate handler per ref; matched automatically
+val plugin = ExternalStatePlugin(
+  ExternalStateHandler(
+    ref = "my-action",
+    handlerFunction = ExternalStateHandler.Function { state, options, transition ->
+      transition("Next")
+    }
+  )
```
sugarmanz: Does the trailing lambda syntax work here?

```kotlin
  ExternalStateHandler(ref = "my-action") { state, options, transition ->
    transition("Next")
  }
```

docs/site/src/content/docs/plugins/multiplatform/external-state.mdx
```diff
@@ -0,0 +1,163 @@
...
+  <Fragment slot="android">
+
+```kotlin
+import com.intuit.playerui.plugins.externalState.ExternalStatePlugin
+
+val player = AndroidPlayer(
+  ExternalStatePlugin(
+    ExternalStateHandler(
+      ref = "my-action",
+      handlerFunction = ExternalStateHandler.Function { state, options, transition ->
```
sugarmanz: Ditto about using trailing lambda syntax as suggested implementation format if it works

sugarmanz (commented):
Small comments regarding documentation, otherwise looks good!

docs/site/src/content/docs/announcements/player-one-dot-0.mdx
KVSRoyal: It does. Updated!

plugins/external-state/core/src/index.ts
KVSRoyal: Update! I also tried to remove the `param` stuff from the web docs.

docs/site/src/content/docs/announcements/player-one-dot-0.mdx
KVSRoyal: Yeah, sure, adding. Although for the 1.0 stuff I think we've so far just been putting things in the migration docs file.

spentacular (approved):

sugarmanz (approved):

PR #825, opened 2026-03-17
Fix HermesNode property access on released JSI objects

kharrop:
This PR prevents a rare but high-impact Android crash in Hermes where the app can hit a native `NullPointerException` while reading JS backed objects. It does not change Player flow/state behavior; it only improves safety and diagnostics in a teardown race window.

## Context

The crash occurs when native code tries to read a property from a Hermes JSI object that has already been released (often due to timing/lifecycle during teardown or navigation).

This shows up as a fatal `NullPointerException` inside Hermes/JSI property access (getProperty), which crashes the app process.

## Testing

```sh
bazel test //jvm/hermes:test //jvm/core:test
```

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

# Release Notes

Added a defensive guard in HermesNode so property reads on a released runtime/object fail safely by throwing a PlayerRuntimeException with context instead of a fatal native NPE.

Added regression tests:
- A deterministic unit test for “released JSI object property read”
- A small concurrency test that reads player.state while player.release() is happening

<details>
  <summary>📦 Published PR as canary version: <code>0.15.2--canary.825.33564</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.15.2--canary.825.33564
</details>


jvm/graaljs/src/main/kotlin/com/intuit/playerui/graaljs/extensions/Lock.kt
```diff
@@ -33,43 +34,34 @@ internal suspend fun <T> Value.lock(timeout: Long = 5000, block: suspend Value.(
...
-internal fun <T> Value.blockingLock(block: Value.() -> T): T {
-    context.runtime.ensureNotReleased()
-    return when {
-        context.runtime.lock.isHeldByCurrentThread -> block()
-        else -> runBlocking {
-            lock { block() }
-        }
-    }
+internal fun <T> Value.blockingLock(block: Value.() -> T): T = when {
+    context.runtime.lock.isHeldByCurrentThread -> block()
```
brocollie08, line 54: these _shouldn't_ be impacted since they'll be on the same thread

jvm/hermes/src/main/kotlin/com/intuit/playerui/hermes/bridge/HermesNode.kt
```diff
@@ -104,13 +104,14 @@ public class HermesNode(
...
+                .mapUndefinedToNull()
+                ?.let { format.decodeFromRuntimeValue(deserializer, it) }
+        }
 
     override fun <T> deserialize(deserializer: DeserializationStrategy<T>): T = runtime
         .evaluateInJSThreadBlocking {
-            jsiObject.asValue(runtime)
-        }.let { format.decodeFromRuntimeValue(deserializer, it) }
+            format.decodeFromRuntimeValue(deserializer, jsiObject.asValue(runtime))
+        }
```
brocollie08: i wrapped the whole thing here and above. `mapUndefineToNull` as well as decoding can both run into similar race conditions since they need to reach the c++ layer

sugarmanz: Don't those operations re-enter `evaluateInJSThreadBlocking` though? Would we be able to remove that? Or is keeping it in the same block just activating the shortcut?

sugarmanz: Do we need to make these updates for the other node impls?

jvm/hermes/src/main/kotlin/com/intuit/playerui/hermes/extensions/Lock.kt
```diff
@@ -40,34 +41,33 @@ internal abstract class DedicatedRuntimeThreadContext internal constructor() : R
 }
 
 internal suspend fun <T> Runtime<*>.evaluateInJSThread(block: suspend RuntimeThreadContext.() -> T): T {
-    ensureNotReleased()
     val currentRuntimeThreadContext = coroutineContext[RuntimeThreadContext]
     // TODO: Put Dedicated Runtime Thread Context in the dispatcher context if we can?
     return if (currentRuntimeThreadContext != null) {
-        block(currentRuntimeThreadContext)
+        ensureNotReleased { block(currentRuntimeThreadContext) }
```
sugarmanz: Why do `ensureNotReleased` here? You're ignoring it for this case in graal:
https://github.com/player-ui/player/pull/825/changes#r3017129498

Just want to be consistent.

jvm/hermes/src/main/kotlin/com/intuit/playerui/hermes/extensions/Lock.kt
```diff
@@ -40,34 +41,33 @@ internal abstract class DedicatedRuntimeThreadContext internal constructor() : R
 }
 
 internal suspend fun <T> Runtime<*>.evaluateInJSThread(block: suspend RuntimeThreadContext.() -> T): T {
-    ensureNotReleased()
```
sugarmanz, line 43: Should this whole function be wrapped in `ensureNotReleased` to avoid needing to call it multiple times?

jvm/j2v8/src/main/kotlin/com/intuit/playerui/j2v8/extensions/Lock.kt
```diff
@@ -22,15 +19,12 @@ internal fun <Context : V8Value, T> Context.evaluateInJSThreadBlocking(
...
-    runtime.ensureNotReleased()
     // if we're already on the dispatcher thread, DON'T BLOCK
     return if (this@evaluateInJSThreadBlocking.runtime.locker.hasLock()) {
         block()
     } else {
         if (!muteLog) runtime.checkBlockingThread(Thread.currentThread())
-        runBlocking {
-            evaluateInJSThread(runtime, block)
-        }
+        runBlocking { runtime.ensureNotReleased { evaluateInJSThread(runtime, block) } }
```
sugarmanz, line 27: Do we need the check here if we're sure `evaluateInJSThread` will wrap `block` with it?

sugarmanz (commented):
Looks fine — main concern is consistency across runtime impls. Even if functionally equivalent, the more it diverges, the more difficult it is to maintain.

jvm/hermes/src/main/kotlin/com/intuit/playerui/hermes/extensions/Lock.kt
brocollie08: yea you right, this shouldn't be necessary

jvm/j2v8/src/main/kotlin/com/intuit/playerui/j2v8/extensions/Lock.kt
brocollie08: it does, though evaluateInJSThread has the guard as well

jvm/hermes/src/main/kotlin/com/intuit/playerui/hermes/bridge/HermesNode.kt
brocollie08: `mapUndefinedToNull` doesn't, decoding does, i moved decoding back out to minimize changes

sugarmanz (approved):

PR #831, opened 2026-03-24
JVM supports deterministic override API for `ExternalActionPlugin`

KVSRoyal:
## What
This is the JVM part of https://github.com/player-ui/player/pull/817. This does not include a docs update. I plan to update the multi-platform docs in a separate PR once the APIs are solidified.

This PR changes the `ExternalActionPlugin` to accept an array of match-handler pairs. These pairs are passed through to the core.

**Before:**
```kotlin
// Single handler approach - one handler per plugin
val plugin = ExternalActionPlugin()
// Register handler dynamically
player.onExternalAction { state, options, transition ->
    when (state.ref) {
        "action-1" -> {
            // Handle action 1
            transition("Next")
        }
        "action-2" -> {
            // Handle action 2
            transition("Other")
        }
    }
}
```

**After:**
```kotlin
// Array-based approach - multiple handlers with specificity
val plugin = ExternalActionPlugin(
    listOf(
        // Less specific - matches any state with ref: "action"
        ExternalActionHandler(
            match = mapOf("ref" to "action"),
            handler = { state, options, transition ->
                // Default handler
                transition("default")
            }
        ),
        // More specific - matches state with both ref and type
        ExternalActionHandler(
            match = mapOf("ref" to "action", "type" to "special"),
            handler = { state, options, transition ->
                // This handler takes precedence when type: "special"
                transition("special")
            }
        )
    )
)
val player = setupPlayer(plugin)
```

## Why

**Deterministic Overrides**: Users want predictable behaviour when multiple plugins or handlers are registered. Registering handlers by `match` ensures that:

* More specific matchers always take precedence (e.g., { ref: "x", type: "y" } beats { ref: "x" })
* Later registrations replace earlier ones with the same specificity
* Behaviour is consistent regardless of registration order

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A` (part of 1.0)

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/plugins/external-action/jvm/src/main/kotlin/com/intuit/playerui/plugins/externalAction/ExternalActionPlugin.kt
+++ b/plugins/external-action/jvm/src/main/kotlin/com/intuit/playerui/plugins/externalAction/ExternalActionPlugin.kt
@@ -8,27 +8,53 @@ import com.intuit.playerui.core.bridge.runtime.add
 import com.intuit.playerui.core.flow.state.NavigationFlowExternalState
 import com.intuit.playerui.core.flow.state.NavigationFlowState
 import com.intuit.playerui.core.player.Player
+import com.intuit.playerui.core.player.PlayerException
 import com.intuit.playerui.core.player.state.ControllerState
 import com.intuit.playerui.core.plugins.JSScriptPluginWrapper
 import com.intuit.playerui.core.plugins.PlayerPlugin
 import com.intuit.playerui.core.plugins.findPlugin
 import com.intuit.playerui.plugins.settimeout.SetTimeoutPlugin
 
-/** Function definition of an external action handler */
-public fun interface ExternalActionHandler {
-    public fun onExternalState(
-        state: NavigationFlowExternalState,
-        options: ControllerState,
-        transition: (String) -> Unit,
-    )
+public data class ExternalActionHandler(
+    val match: Map<String, Any>,
+    val handler: Handler,
+) {
+    /**
+     * Function definition of an external action handler
+     * @param state The NavigationFlowExternalState that was transitioned to
+     * @param options The ControllerState providing access to data and expression evaluation
+     * @param transition Callback to transition to the next state
+     */
+    public fun interface Handler {
+        public fun onExternalAction(
+            state: NavigationFlowExternalState,
+            options: ControllerState,
+            transition: (String) -> Unit,
+        )
+    }
+
+    init {
+        require(match.containsKey("ref")) {
+            "The match map must contain a 'ref' key. Got: $match"
+        }
+    }
 }
 
-/** Core plugin wrapper providing external action support for the JVM Player */
+/**
+ * Core plugin wrapper providing external action support for the JVM Player.
+ *
+ * This plugin uses a registry-based approach to match external states to handler functions.
+ * Multiple handlers can be registered, and handlers are matched using partial object matching
+ * with specificity ordering (more specific matches take precedence).
+ *
+ * @param handlers List of handler configurations with matchers and handler functions.
+ */
 public class ExternalActionPlugin(
-    private var handler: ExternalActionHandler? = null,
+    vararg handlers: ExternalActionHandler,
 ) : JSScriptPluginWrapper(PLUGIN_NAME, sourcePath = BUNDLED_SOURCE_PATH),
     PlayerPlugin {
     private lateinit var player: Player
+    private val handlers: List<ExternalActionHandler> = handlers.toList()
 
     override fun apply(player: Player) {
         this.player = player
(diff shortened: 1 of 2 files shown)


plugins/external-action/jvm/src/test/kotlin/com/intuit/playerui/plugins/externalAction/ExternalActionPluginTest.kt
```diff
@@ -80,33 +94,126 @@ internal class ExternalActionPluginTest : PlayerTest() {
...
+            ),
+        )
+
+        setupPlayer(plugin)
+        val result = player.start(jsonFlow).await()
+        assertEquals(result.endState.outcome, "BCK")
+    }
+
+    @TestTemplate
+    fun testExternalActionHandlingWithSpecificity() = runBlockingTest {
```
brocollie08, line 138: I'm a bit confused, is this test passing? I don't see any changes to the content, how could it map to the more specific handler

plugins/external-action/jvm/src/test/kotlin/com/intuit/playerui/plugins/externalAction/ExternalActionPluginTest.kt
```diff
@@ -80,33 +94,126 @@ internal class ExternalActionPluginTest : PlayerTest() {
...
+                    match = mapOf("ref" to "test-1"),
+                    handler = { _, options, transition ->
+                        assertEquals(options.data.get("transitionValue"), "Next")
 
-            // Test expression evaluation
-            options.expression.evaluate("{{transitionValue}} = 'Prev'")
-            assertEquals(options.data.get("transitionValue"), "Prev")
+                        // Test expression evaluation
+                        options.expression.evaluate("{{transitionValue}} = 'Prev'")
+                        assertEquals(options.data.get("transitionValue"), "Prev")
```
brocollie08: Could you please check one of the following:
- this test _can_ fail if you assert something that doesn't match, I've seen callbacks swallow assertion failures if it's not in a test thread or something like that
- write the test in a way so the assert isn't in a callback

plugins/external-action/jvm/src/main/kotlin/com/intuit/playerui/plugins/externalAction/ExternalActionPlugin.kt
```diff
@@ -14,18 +14,42 @@ import com.intuit.playerui.core.plugins.PlayerPlugin
...
 
-/** Function definition of an external action handler */
-public fun interface ExternalActionHandler {
-    public fun onExternalState(
-        state: NavigationFlowExternalState,
-        options: ControllerState,
-        transition: (String) -> Unit,
-    )
+public data class ExternalActionHandler(
+    val match: Map<String, Any>,
```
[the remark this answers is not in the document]
KVSRoyal: Resolved offline; will cover in the final jumbo PR.

plugins/external-action/jvm/src/main/kotlin/com/intuit/playerui/plugins/externalAction/ExternalActionPlugin.kt
```diff
@@ -14,18 +14,42 @@ import com.intuit.playerui.core.plugins.PlayerPlugin
...
+            state: NavigationFlowExternalState,
+            options: ControllerState,
+            transition: (String) -> Unit,
+        )
+    }
+
+    init {
+        require(match.containsKey("ref")) {
+            "The match map must contain a 'ref' key. Got: $match"
+        }
```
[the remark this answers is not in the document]
KVSRoyal: Resolved offline; will cover in the final jumbo PR. Resolution is to use the error handler to snag for now.

plugins/external-action/jvm/src/test/kotlin/com/intuit/playerui/plugins/externalAction/ExternalActionPluginTest.kt
KVSRoyal: It is passing, yeah. There is no content change. `main` actually already has an `extraProperty` in it for us to do more specific matching against: https://github.com/player-ui/player/blob/6222f07d6fbf274ceb6ecd11a094456524557841/plugins/external-action/jvm/src/test/kotlin/com/intuit/playerui/plugins/externalAction/ExternalActionPluginTest.kt#L36.

plugins/external-action/jvm/src/test/kotlin/com/intuit/playerui/plugins/externalAction/ExternalActionPluginTest.kt
KVSRoyal: Looks like it physically can fail even though the assertion is in a callback. I went ahead and moved the assertions out anyway though.

Image

PR #834, opened 2026-03-30
Align Post Initialization Plugin Registration API

### LINKED ISSUE #790: Align Post Init `registerPlugin` APIs across Platforms
Currently, Core Player supports the ability to add plugins to the instance after it is initialized. React/iOS/JVM/Android all have various levels of support for this and mixed APIs for this support. We should align this functionality across all platforms:

- React
 - Doesn't support adding Core plugins
- iOS 
 - Only supports adding Core plugins
- Android/JVM
 - Seemingly no support?

KetanReddy:
Closes #790 

All platforms now support registering plugins with a player after it has been instantiated.

## JVM (HeadlessPlayer)
Added registerPlugin(plugin: Plugin) that appends to the plugins list and applies the plugin appropriately (RuntimePlugin, JSPluginWrapper, or PlayerPlugin).

## Android (AndroidPlayer)
- Added registerPlugin(plugin: Plugin) that delegates to the underlying HeadlessPlayer for standard plugins, or stores and applies AndroidPlayerPlugin instances directly.

## iOS (HeadlessPlayer / HeadlessPlayerImpl)
- Added registerPlugin(_ plugin: NativePlugin) as a protocol requirement with a default implementation. Concrete implementations (e.g. HeadlessPlayerImpl) track registered plugins in a mutable plugins list.

## React (ReactPlayer)
- Updated registerPlugin to accept both ReactPlayerPlugin and PlayerPlugin, calling apply on the core player and applyReact on the React player as appropriate.

TODO:

- [ ] Update Migration Guide

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
+++ b/android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
@@ -306,8 +306,6 @@ public class AndroidPlayer private constructor(
         player.registerPlugin(plugin)
         if (plugin is AndroidPlayerPlugin) {
             plugin.apply(this)
-        } else {
-            player.registerPlugin(plugin)
         }
     }
 
--- a/android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
+++ b/android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
@@ -48,7 +48,7 @@ public typealias AndroidPlayerConfig = AndroidPlayer.Config
  */
 public class AndroidPlayer private constructor(
     public val player: HeadlessPlayer,
-    override val plugins: List<Plugin> = player.plugins,
+    plugins: List<Plugin> = player.plugins,
 ) : Player() {
     /** Convenience constructor to provide vararg style [plugins] parameter */
     public constructor(
@@ -89,6 +89,9 @@ public class AndroidPlayer private constructor(
         player,
         (plugins.toList() + player.plugins).distinct(),
     )
+
+    private val _plugins: MutableList<Plugin> = plugins.toMutableList()
+    override val plugins: List<Plugin> get() = _plugins
 
     override val logger: TapableLogger by player::logger
 
@@ -297,6 +300,16 @@ public class AndroidPlayer private constructor(
         clearCaches()
         player.release()
         hooks.release.call()
+    }
+
+    /** Register and apply a [Plugin] to this player after instantiation. */
+    public fun registerPlugin(plugin: Plugin) {
+        _plugins.add(plugin)
+        if (plugin is AndroidPlayerPlugin) {
+            plugin.apply(this)
+        } else {
+            player.registerPlugin(plugin)
+        }
     }
 
     // TODO: Do we even need context as a param anymore?
--- a/android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
+++ b/android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
@@ -90,7 +90,7 @@ public class AndroidPlayer private constructor(
         (plugins.toList() + player.plugins).distinct(),
     )
 
-    override val plugins: List<Plugin> by player::plugins
+    override val plugins: List<Plugin> get() = player.plugins
 
     override val logger: TapableLogger by player::logger
 
@@ -302,8 +302,8 @@ public class AndroidPlayer private constructor(
     }
 
     /** Register and apply a [Plugin] to this player after instantiation. */
-    public fun registerPlugin(plugin: Plugin) {
-        plugins.add(plugin)
+    override fun registerPlugin(plugin: Plugin) {
+        player.registerPlugin(plugin)
         if (plugin is AndroidPlayerPlugin) {
             plugin.apply(this)
         } else {
--- a/android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
+++ b/android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
@@ -90,8 +90,7 @@ public class AndroidPlayer private constructor(
         (plugins.toList() + player.plugins).distinct(),
     )
 
-    private val _plugins: MutableList<Plugin> = plugins.toMutableList()
-    override val plugins: List<Plugin> get() = _plugins
+    override val plugins: List<Plugin> by player::plugins
 
     override val logger: TapableLogger by player::logger
 
@@ -304,7 +303,7 @@ public class AndroidPlayer private constructor(
 
     /** Register and apply a [Plugin] to this player after instantiation. */
     public fun registerPlugin(plugin: Plugin) {
-        _plugins.add(plugin)
+        plugins.add(plugin)
         if (plugin is AndroidPlayerPlugin) {
             plugin.apply(this)
         } else {
--- a/ios/core/Sources/Player/HeadlessPlayer.swift
+++ b/ios/core/Sources/Player/HeadlessPlayer.swift
@@ -225,6 +225,17 @@ public extension HeadlessPlayer {
         jsPlayerReference?.invokeMethod("registerPlugin", withArguments: [plugin.pluginRef as Any])
     }
 
+    /// Registers a NativePlugin with player after instantiation
+    /// For JSBasePlugin instances, delegates to the typed overload; for other NativePlugins, calls apply directly
+    /// - Parameter plugin: The plugin to register
+    func registerPlugin(_ plugin: NativePlugin) {
+        if let jsPlugin = plugin as? JSBasePlugin {
+            registerPlugin(jsPlugin)
+        } else {
+            plugin.apply(player: self)
+        }
+    }
+
     /**
      Starts Player for the given flow
      - parameters:
--- a/jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
+++ b/jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
@@ -74,7 +74,8 @@ public class HeadlessPlayer @ExperimentalPlayerApi @JvmOverloads public construc
 
     private val player: Node
 
-    override var plugins: List<Plugin> = plugins; private set
+    override var plugins: List<Plugin> = plugins
+        private set
 
     override val node: Node by ::player
 
@@ -200,17 +201,15 @@ public class HeadlessPlayer @ExperimentalPlayerApi @JvmOverloads public construc
     }
 
     /** Register and apply a [Plugin] to this player after instantiation. */
-    public fun registerPlugin(plugin: Plugin) {
+    override fun registerPlugin(plugin: Plugin) {
         plugins = plugins + plugin
-        when (plugin) {
-            is RuntimePlugin -> {
-                plugin.apply(runtime)
-                if (plugin is JSPluginWrapper) {
-                    registerPlugin.invoke(plugin.instance)
-                }
+        if (plugin is RuntimePlugin) {
+            plugin.apply(runtime)
+            if (plugin is JSPluginWrapper) {
+                registerPlugin.invoke(plugin.instance)
             }
-            is PlayerPlugin -> plugin.apply(this)
         }
+        if (plugin is PlayerPlugin) plugin.apply(this)
     }
 
     internal companion object {
--- a/jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
+++ b/jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
@@ -1,6 +1,7 @@
 package com.intuit.playerui.core.player
 
 import com.intuit.playerui.core.bridge.Completable
+import com.intuit.playerui.core.bridge.Invokable
 import com.intuit.playerui.core.bridge.Node
 import com.intuit.playerui.core.bridge.NodeWrapper
 import com.intuit.playerui.core.bridge.Promise
@@ -13,6 +14,7 @@ import com.intuit.playerui.core.bridge.runtime.add
 import com.intuit.playerui.core.bridge.runtime.runtimeContainers
 import com.intuit.playerui.core.bridge.runtime.runtimeFactory
 import com.intuit.playerui.core.bridge.serialization.serializers.NodeSerializableField
+import com.intuit.playerui.core.bridge.serialization.serializers.NodeSerializableFunction
 import com.intuit.playerui.core.constants.ConstantsController
 import com.intuit.playerui.core.experimental.ExperimentalPlayerApi
 import com.intuit.playerui.core.logger.TapableLogger
@@ -72,8 +74,7 @@ public class HeadlessPlayer @ExperimentalPlayerApi @JvmOverloads public construc
 
     private val player: Node
 
-    private val _plugins: MutableList<Plugin> = plugins.toMutableList()
-    override val plugins: List<Plugin> get() = _plugins
+    override var plugins: List<Plugin> = plugins; private set
 
     override val node: Node by ::player
 
@@ -174,6 +175,8 @@ public class HeadlessPlayer @ExperimentalPlayerApi @JvmOverloads public construc
             .onEach { it.apply(this) }
     }
 
+    private val registerPlugin: Invokable<Unit> by NodeSerializableFunction()
+
     override fun start(flow: String): Completable<CompletedState> = try {
         start(runtime.execute("($flow)") as Node)
     } catch (exception: Exception) {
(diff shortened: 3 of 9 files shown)


react/player/src/player.tsx
```diff
@@ -154,10 +154,19 @@ export class ReactPlayer {
 
   /** Register and apply [Plugin] if one with the same symbol is not already registered. */
   public registerPlugin(plugin: ReactPlayerPlugin): void {
```
sugarmanz: Does this need to update to `plugin: ReactPlayerPlugin | PlayerPlugin`?

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
```diff
@@ -72,6 +72,9 @@ public class HeadlessPlayer @ExperimentalPlayerApi @JvmOverloads public construc
 
     private val player: Node
 
+    private val _plugins: MutableList<Plugin> = plugins.toMutableList()
+    override val plugins: List<Plugin> get() = _plugins
```
sugarmanz: Prefer immutable lists, use `var` with private `set`:
```suggestion
   override var plugins: List<Plugin> = plugins; private set
```

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
```diff
@@ -193,6 +196,20 @@ public class HeadlessPlayer @ExperimentalPlayerApi @JvmOverloads public construc
...
     }
 
+    /** Register and apply a [Plugin] to this player after instantiation. */
+    public fun registerPlugin(plugin: Plugin) {
+        _plugins.add(plugin)
+        when (plugin) {
+            is RuntimePlugin -> {
+                plugin.apply(runtime)
+                if (plugin is JSPluginWrapper) {
+                    player.getInvokable<Unit>("registerPlugin")!!.invoke(plugin.instance)
```
sugarmanz: Use `NodeSerializableFunction` to add as a private member and invoke — this way deserialization only happens once.

android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
```diff
@@ -90,6 +90,9 @@ public class AndroidPlayer private constructor(
         (plugins.toList() + player.plugins).distinct(),
     )
 
+    private val _plugins: MutableList<Plugin> = plugins.toMutableList()
+    override val plugins: List<Plugin> get() = _plugins
```
sugarmanz: Just delegate to `HeadlessPlayer.plugins` for the store:
```suggestion
    override val plugins: List<Plugin> by player::plugins
```

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
sugarmanz: One downfall to storing plugins at this level is if there are any core plugins that invoke `Player.registerPlugin`, we won't be able to find them. I'm not entirely sure how that'd work, but it was something that I was keeping in mind for this workstream. Maybe we try to store everything at the bottom-most layer and query for it each time?

react/player/src/player.tsx
KetanReddy: Oh fair. `ReactPlayerPlugin` extends `PlayerPlugin`  so _technically_ it would work but TS would complain if you just try and pass in a `PlayerPlugin`

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
KetanReddy: I think there is where my lack of knowledge of JVM Player is failing me. Would this not be the lowest level or do you mean store them in the JS runtime?

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
KetanReddy: So like add `registerPlugin` as a `NodeSerializableFunction` as a private member to `player`?

ios/test-utils-core/Sources/utilities/HeadlessPlayerImpl.swift
```diff
@@ -18,8 +20,21 @@ open class HeadlessPlayerImpl: HeadlessPlayer {
...
     }
+
+    public func registerPlugin(_ plugin: NativePlugin) {
+        plugins.append(plugin)
+        if let jsPlugin = plugin as? JSBasePlugin {
+            assert(jsPlayerReference != nil, "Cannot register plugins before setupPlayer(context:plugins:) is called")
+            jsPlugin.context = jsPlayerReference?.context
+            jsPlayerReference?.invokeMethod("registerPlugin", withArguments: [jsPlugin.pluginRef as Any])
+        } else {
+            assert(jsPlayerReference != nil, "Cannot register plugins before setupPlayer(context:plugins:) is called")
```
KVSRoyal: 1. Can we lift the assertion out of the conditional since it's in both branches?
2. Should the default implementation of registerPlugin also have the assert happen no matter which branch?

KVSRoyal (approved):
Just one small nit. Otherwise, lgtm.

Also, fwiw, I was a bit worried about making every `JSBasePlugin` a `NativePlugin` too but it looks like all of our use cases more or less support that, so I think it's okay. Might be good for @JunDangIntuit @nancywu1 to also weigh in though.

KetanReddy:
> Also, fwiw, I was a bit worried about making every `JSBasePlugin` a `NativePlugin` too but it looks like all of our use cases more or less support that, so I think it's okay. Might be good for @JunDangIntuit @nancywu1 to also weigh in though.

That was something I was going to bring up in terms of how big of a change that would be. But I think an added benefit there is that it is more in line with how JVM treats plugins for core vs itself so that should be some bonus platform alignment. 

Definitely a conversation to have though to make sure it is the best path forward. The iOS test failures look like issues related to that so I'll hold off making any more changes there until we agree on the path forward.

android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
```diff
@@ -48,7 +48,7 @@ public typealias AndroidPlayerConfig = AndroidPlayer.Config
  */
 public class AndroidPlayer private constructor(
     public val player: HeadlessPlayer,
-    override val plugins: List<Plugin> = player.plugins,
+    plugins: List<Plugin> = player.plugins,
```
sugarmanz, line 51: Can we remove this altogether?

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
```diff
@@ -72,6 +77,9 @@ public class HeadlessPlayer @ExperimentalPlayerApi @JvmOverloads public construc
 
     private val player: Node
 
+    override var plugins: List<Plugin> = plugins
+        internal set
```
sugarmanz: If we can:
```suggestion
        private set
```

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
```diff
@@ -218,3 +240,21 @@ public class HeadlessPlayer @ExperimentalPlayerApi @JvmOverloads public construc
...
+/**
+ * Find a [Plugin] by type. First checks the native [plugins] list and returns [T] if found.
+ * If not found natively, falls back to the JS player's findPlugin by evaluating the plugin's
+ * symbol expression in the runtime — using [WithSymbol.symbol] from [T]'s companion object if
+ * available, otherwise deriving it from the class name (e.g. "MyPlugin.symbol"). Returns the
+ * raw [Node] from the JS runtime if found only there, or null if not found in either location.
+ */
+public inline fun <reified T : Plugin> HeadlessPlayer.findPlugin(): Any? {
+    plugins.filterIsInstance<T>().firstOrNull()?.let { return it }
+    val symbolExpr = (T::class.companionObjectInstance as? WithSymbol)?.symbol
```
sugarmanz: I don't see anywhere that's actually implementing this pattern where the companion object of some plugin class is implementing `WithSymbol`. Instead, it looks like `JSScriptPluginWrapper` implements `WithSymbol` itself? That would require you to instantiate the class you're trying to find in order to build the name which is non-trivial given most of the plugins require the constructor params for that plugin to instantiate — which is a problem for this API as well given you couldn't just find the `Node` representing the JS instance and create a new instance (I understand that's why we're currently returning the `Node` if we find it).

We're baking some unfounded patterns into this implementation, if we're to add this functionality, we should prove it out. That said, I don't know that I love the `companionObjectInstance` reflection here.. Nor would we be able to use the `JSScriptPluginWrapper` at face value.

I've been wanting to redo our module loading system for a while, really we need a way to describe the exports of some module we're loading, which would be the way to load the symbol. Given we also need a way to instrument our plugin wrappers with an already instantiated JS `Node`, I'm pressed to say there is more work here than benefit we get for short term solutions that'd need to change when we make changes in the future.

We can talk more feedback here if we decide to continue trying to get this into this PR.

jvm/core/src/test/kotlin/com/intuit/playerui/core/player/HeadlessPlayerTest.kt
```diff
@@ -628,4 +631,65 @@ internal class HeadlessPlayerTest :
...
+
+    @TestTemplate
+    fun `findPlugin returns native plugin when registered natively`() {
+        var applied = false
+        val plugin = object : PlayerPlugin {
+            override fun apply(player: Player) {
+                applied = true
+            }
+        }
+        (player as HeadlessPlayer).registerPlugin(plugin)
```
sugarmanz: Ah, we should not be casting here — we need to add `registerPlugin` to the base `Player` class definition so that anybody could call it on any Player instance.

sugarmanz (changes requested):

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
KetanReddy: Yeah lets figure out the best path forward. If we're going to do this for 1.0 we should try and do this right unless the scope is going to massively balloon. If we need to totally redo a system to make this target state is there a way we can make the public API clean and make the underlying changes at some point down the line?

ios/core/Sources/Player/HeadlessPlayer.swift
```diff
@@ -165,6 +174,9 @@ public protocol HeadlessPlayer {
  Default Implementation for player methods
  */
 public extension HeadlessPlayer {
+    /// Default empty plugins list for conformers that do not track native plugins
+    var plugins: [NativePlugin] { [] }
```
JunDangIntuit: the computed property cannot be mutated, is it expected?

jvm/core/src/main/kotlin/com/intuit/playerui/core/player/HeadlessPlayer.kt
sugarmanz: >  is there a way we can make the public API clean and make the underlying changes at some point down the line?

The public API already exists with `Player.findPlugin` — it'd just be enhancing it to account for symbols.

If we want to make it work, we should do something like:
```kt
@ExperimentalPlayerApi
interface Module {
  val moduleName: String

  // we would want to delegate more loading responsibilities here but not necessary for now, just establishing the pattern
}

fun Module.pluginSymbol(runtime: Runtime<*>, key: String = "Symbol"): String? {
  val module = runtime.getObject(moduleName) ?: throw PlayerRuntimeException("module not loaded into runtime")
  return module.getSymbol(key)
}

class MyPlugin(someArgs: Any) : JSScriptPluginWrapper("$moduleName.$PLUGIN_NAME", sourcePath = BUNDLED_SOURCE_PATH) {

  companion object : Module {
    override val moduleName = "MyPlugin"
    
    private const val PLUGIN_NAME = "MyPlugin"
    private const val BUNDLED_SOURCE_PATH = "plugins/my-plugin/core/dist/MyPlugin.native.js"
  }
}

// Any? being a return type here does change the public API to be less strict unfortunately, maybe a findJSPlugin is warranted if you're wanting this functionality to not change the existing public API for worse
public inline fun <reified T : Plugin> Player.findPlugin(): Any? {
  plugins.filterIsInstance<T>().firstOrNull()?.let { return it }

  // honestly hate using heavy reflection for this, but i don't see another good solution
  val module = T::class.companionObjectInstance as? Module
    ?: return warnPluginNotFound<T>(T::class.java.simpleName) // no module definition to lookup
  val runtime = when (this) {
    is AndroidPlayer -> player.runtime
    is HeadlessPlayer -> runtime
    else -> return warnPluginNotFound<T>(T::class.java.simpleName) // no runtime to look up the symbol in
  }
  val symbol = module.pluginSymbol(runtime) ?: return warnPluginNotFound<T>(T::class.java.simpleName) // no symbol to use

  // somehow reference the decoded findPlugin from the Player instance, should likely be decoded/invoked the same way registerPlugin is, so moving findPlugin to Player altogether (some of the above logic would be simpler if findPlugin flows through Player instead of an extension function
  return _findPlugin(symbol) ?: warnPluginNotFound<T>(T::class.java.simpleName)
}
```

KetanReddy:
After discussions with @sugarmanz, the work required to make the changes required for `findPlugin` on JVM/Android are going to be fairly substantial and require changes to how modules are loaded. To limit the scope of this PR I've archived the old working branch on the branch `archive-find-plugin-api` and rolled back the associated commits here. The PR description has been updated to reflect these changes. 

@sugarmanz can you start a discussion on the module loading changes so we can work though exactly what that looks like and the goals of those changes?

android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
KetanReddy: Turns out we can't because `Pluggable` requires it to be defined. If there is a better way let me know

ios/core/Sources/Player/HeadlessPlayer.swift
```diff
@@ -225,6 +225,17 @@ public extension HeadlessPlayer {
         jsPlayerReference?.invokeMethod("registerPlugin", withArguments: [plugin.pluginRef as Any])
     }
 
+    /// Registers a NativePlugin with player after instantiation
+    /// For JSBasePlugin instances, delegates to the typed overload; for other NativePlugins, calls apply directly
+    /// - Parameter plugin: The plugin to register
+    func registerPlugin(_ plugin: NativePlugin) {
```
KetanReddy, line 231: Not sure if this will stop `JSBasePlugin` instances from being passed in?

KVSRoyal: As long as the plugin conforms to `NativePlugin`, it can be passed in. If it only conforms to `JSBasePlugin`, then it cannot be passed in.

I think that's fine / intended based on current architecture. Something that only conforms to `JSBasePlugin` is not something users should be using directly---is my understanding.

KVSRoyal (approved):

android/player/src/main/kotlin/com/intuit/playerui/android/AndroidPlayer.kt
```diff
@@ -299,6 +301,16 @@ public class AndroidPlayer private constructor(
...
     }
 
+    /** Register and apply a [Plugin] to this player after instantiation. */
+    override fun registerPlugin(plugin: Plugin) {
+        player.registerPlugin(plugin)
+        if (plugin is AndroidPlayerPlugin) {
+            plugin.apply(this)
+        } else {
+            player.registerPlugin(plugin)
+        }
```
sugarmanz: Remove the `else` block here — this is adding the plugin to the `plugins` list twice.

sugarmanz (changes requested):

KetanReddy:
>As long as the plugin conforms to NativePlugin, it can be passed in. If it only conforms to JSBasePlugin, then it cannot be passed in.
>
>I think that's fine / intended based on current architecture. Something that only conforms to JSBasePlugin is not something users should be using directly---is my understanding.
@KVSRoyal It looks like thats whats used to wrap any core plugin so I think that should be changed.

KetanReddy:
> > As long as the plugin conforms to NativePlugin, it can be passed in. If it only conforms to JSBasePlugin, then it cannot be passed in.
> > I think that's fine / intended based on current architecture. Something that only conforms to JSBasePlugin is not something users should be using directly---is my understanding.
> > @KVSRoyal It looks like thats whats used to wrap any core plugin so I think that should be changed.

Actually my bad, I forgot this just overloaded the existing registerPlugin function for `JSBasePlugin` to add support for native plugins. Turns out I lost some context redoing this PR a few times 😅

spentacular (approved):

ios/core/Sources/Player/HeadlessPlayer.swift
KVSRoyal: > sugarmanz: @KVSRoyal It looks like thats whats used to wrap any core plugin so I think that should be changed.
> 
> KetanReddy: my bad, I forgot this just overloaded the existing registerPlugin function for JSBasePlugin to add support for native plugins. Turns out I lost some context redoing this PR a few times 😅

To sugarmanz: is it? We can talk about it offline. We should really start logging these architecture decisions somewhere.

To KetanReddy: LOL it happens

To both: reply here please lol (if any more discussion). The quoting is tedious

sugarmanz (approved):

PR #835, opened 2026-03-31
Propagate styles in composable asset 

justin-medeiros:
### Summary

  - When a ComposableAsset renders a nested ComposableAsset with xmlStyles, those
  styles were lost — only textStyle was propagated via LocalTextStyle. This means a
  Compose → Compose → XML view chain would not apply ContextThemeWrapper to the leaf
   XML view.
  - This PR fixes that by overlaying xmlStyles onto LocalContext in the CompositionLocalProvider block
  of `RenderableAsset.compose()`, so intermediate Compose layers carry the styled
  context through to any eventual composeAndroidView call.

This PR also:
  - Adds ComposableAssetTest with 5 integration tests covering initView, full player
  flow, Compose→Compose nesting, LocalTextStyle propagation, and 3-level
  Compose→Compose→XML style propagation with ContextThemeWrapper assertion.
  - Upgrades Espresso from 3.3.0 to 3.6.1 to fix Java 21 reflection incompatibility
  with createComposeRule().

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


libs.versions.toml
```diff
@@ -51,7 +52,7 @@ logback-classic = { module = "ch.qos.logback:logback-classic", version = "1.2.11
 # Testing
 androidx-test-core = { module = "androidx.test:core", version.ref = "androidx-test" }
 androidx-test-runner = { module = "androidx.test:runner", version.ref = "androidx-test" }
-androidx-test-espresso = { module = "androidx.test.espresso:espresso-intents", version = "3.3.0" }
+androidx-test-espresso = { module = "androidx.test.espresso:espresso-intents", version = "3.6.1" }
```
justin-medeiros, line 55: Upgrade espresso as we were getting a java reflection issue with using `composeTestRule` in our tests

android/player/BUILD
```diff
@@ -56,11 +56,14 @@ instrumented_test_deps = [
     artifact("io.mockk:mockk"),
     artifact("org.jetbrains.kotlinx:kotlinx-coroutines-test"),
     artifact("androidx.test:core"),
-    "//jvm:kotlin_serialization"
+    "//jvm:kotlin_serialization",
+    artifact("androidx.compose.ui:ui-test-junit4"),
+    artifact("androidx.compose.ui:ui-test-manifest"),
```
justin-medeiros, line 61: Needed for compose tests

KetanReddy:
@justin-medeiros you may want to figure out signed commits before you add any more

android/demo/BUILD
```diff
@@ -24,6 +24,7 @@ test_deps = [
     artifact("androidx.test.espresso:espresso-intents"),
     artifact("androidx.test.ext:junit-ktx"),
     artifact("androidx.compose.ui:ui-test-junit4"),
+    artifact("com.google.guava:guava"),
```
justin-medeiros, line 27: Had to add this dependency as we were getting this error when upgrading espresso: `Didn't find class "com.google.common.util.concurrent.ListenableFuture"`

android/player/src/main/kotlin/com/intuit/playerui/android/compose/ComposableAsset.kt
brocollie08, line 90: would it be possible to pass style in here to the assetContext and drop the style arg in composeAndroidView entirely?

justin-medeiros: Made a change to try it. Can I get a canary to test?

brocollie08: i don't think this is sufficient, what about nested composables inside AndroidViews, they lose the textstyle, right?

brocollie08 (approved):

PR #836, opened 2026-04-01
Externalized Core Player from Core Plugin Bundles

brocollie08:
Currently, the plugin native bundles contain an entire copy of Player - this PR externalizes the Player from the native bundles, reducing size burden of the native bundles. Native plugins now need core player bundle to be loaded before in order to use. 

This also makes the change to iOS HeadlessPlayer to load core Player right away, before any plugin instantiation.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

<details>
  <summary>📦 Published PR as canary version: <code>0.15.3--canary.836.34893</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 0.15.3--canary.836.34893
</details>


intuit-svc:
## Build Preview

Your PR was deployed by CircleCI [#34893](https://circleci.com/gh/player-ui/player/34893) on `Tue, 07 Apr 2026 18:09:14 GMT` with this version:

```
0.15.3--canary.836.34893
```

### 📖 Docs ([View site](https://player-ui.github.io/pr/836/))

jvm/perf/src/main/kotlin/com/intuit/playerui/perf/jmh/PlayerPerf.kt
```diff
@@ -94,6 +97,11 @@ public abstract class RuntimePerformance {
         runtime = it
     }
 
+    protected fun releaseRuntime() {
+        runBlocking { runtime.scope.coroutineContext[Job]?.children?.forEach { it.join() } }
+        runtime.release()
+    }
```
sugarmanz: Nit: name this so it's clear what we're doing when we call it, `releaseWhenIdle` maybe?

plugins/metrics/core/BUILD
```diff
@@ -2,16 +2,21 @@ load("@rules_player//javascript:defs.bzl", "js_pipeline")
...
 load("//tools:defs.bzl", "NATIVE_BUILD_DEPS", "tsup_config", "vitest_config")
 
+NATIVE_BUNDLE = "MetricsPlugin"
+
 npm_link_all_packages(name = "node_modules")
 
-tsup_config(name = "tsup_config")
+tsup_config(
+    name = "tsup_config",
+    native_bundle = NATIVE_BUNDLE,
```
sugarmanz, line 11: This plugin also has a [dependency on the beacon plugin](https://github.com/player-ui/player/blob/6222f07d6fbf274ceb6ecd11a094456524557841/plugins/metrics/core/package.json#L6), which means it'll have it's own copy too. Possible to configure this to accept some extra externals? Or should we just auto-populate the native TSUP config to resolve all workspace packages?

sugarmanz:
Add some release notes and/or update the PR title to describe the resulting change/impact. I've got a feeling this could cause some issues for consumers.

If we're changing bundling strategy too, we should update our docs to say why things are external or change the template.

We should really just refactor module loading in general to be a bit more standardized

KetanReddy:
>  I've got a feeling this could cause some issues for consumers.

Do we want to consider this as part of 1.0 or post 1.0 if it could be breaking?

plugins/metrics/core/BUILD
brocollie08: Is Beacon's presence here as "given" as Player is for other plugins?

sugarmanz: We'd need to apply it [here](https://github.com/player-ui/player/blob/6222f07d6fbf274ceb6ecd11a094456524557841/plugins/metrics/jvm/src/main/kotlin/com/intuit/playerui/plugins/metrics/MetricsPlugin.kt#L29) if not found to be sure

sugarmanz:
> > I've got a feeling this could cause some issues for consumers.
> 
> Do we want to consider this as part of 1.0 or post 1.0 if it could be breaking?

Not necessarily, as long as we test in a few different places (like in any internal test framework). Just want the changelog to clearly reflect what changed in case we need to understand what happened when releasing this.

PR #847, opened 2026-04-15
Restore Xcode local development

KVSRoyal:
Some upgrades from https://github.com/player-ui/player/pull/779 break local Xcode builds (although they pass in CI/CD due to bazel weirdness). Downgrade some of our bazel deps so things work locally again. Xcode 26-26.4 should work for local development.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [x] Unable to update docs


core/player/src/controllers/data/utils.ts
```diff
@@ -8,9 +8,7 @@ import type {
 } from "../../data";
 
 /** Wrapper for the Data Controller Class that prevents writes */
-export class ReadOnlyDataController
-  implements DataModelWithParser<DataModelOptions>
-{
+export class ReadOnlyDataController implements DataModelWithParser<DataModelOptions> {
```
KVSRoyal, line 11: These are all lint fixes. I had to relint for some reason.

MODULE.bazel
```diff
@@ -110,13 +110,6 @@ register_toolchains("//jvm:kotlin_toolchain")
 
 # Android
 bazel_dep(name = "rules_android", version = "0.6.5")
-archive_override(
```
sugarmanz, line 113: Delete the patch file too if we're no longer using it

KVSRoyal: Good catch! Removed

spentacular:
@claude can you review this?

spentacular (approved):

PR #848, opened 2026-04-17
Add ios-review skill with iOS code conventions

KVSRoyal:
We're getting more ios contribution since AI has become more popular. Now that we have more PRs, we ios reviewers end up putting a lot of duplicate comments across PRs. To reduce churn in human-led reviews, this PR introduces a new Claude skill `ios-review` to handle common issues. This skill is also intended to be human-readable documentation of our design standards.

This also adds some more `swiftlint` rules but does not enforce them yet because that would add a lot of reformatting to this PR. Instead, we will revise our swiftlint config fully and reformat as part of this issue: https://github.com/player-ui/player/issues/849.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/.claude/skills/ios-review/SKILL.md
+++ b/.claude/skills/ios-review/SKILL.md
@@ -1,0 +1,177 @@
+---
+name: ios-review
+description: Review Swift/iOS code against team conventions. Use when the user asks to review Swift code, runs /ios-review, or when writing or editing Swift files in this repo. Enforces standards that linters cannot catch.
+context: fork
+---
+
+## Current diff (when reviewing a PR)
+- Diff: !`git diff HEAD~1 2>/dev/null || echo "No diff available"`
+
+# iOS Code Review — Team Conventions
+
+Review the code (or the diff above if present) against the rules below. Apply **General Rules** always. Apply **UI Rules** only if the changes include SwiftUI views (files containing `View`, `body`, `@ViewBuilder`, or SwiftUI imports).
+
+For each violation, report:
+- **Rule**: which rule was broken
+- **Location**: file + line number
+- **Suggestion**: concrete fix
+
+If no violations are found, say so clearly.
+
+---
+
+## General Rules
+
+### 1. Function length ≤ 50 lines
+Functions should be 50 lines or fewer. If a function exceeds this, suggest extracting logical chunks into named helpers.
+
+**Why:** Long functions are harder to read, test, and maintain. Shorter functions with clear names make the call site self-documenting.
+
+### 2. Avoid nested conditionals
+Nested conditionals (`if`, `guard`, `switch`, `for`, `while`, `repeat`) are a bad pattern in iOS code. Prefer:
+- `guard`/`else` to exit early instead of nesting
+- Extracting conditions into well-named boolean variables or functions
+- AI-assisted rewrites when nesting is deep
+
+Flag any conditional nested inside another conditional.
+
+**Why:** Deep nesting hides the happy path, increases cognitive load, and makes it easy to miss edge cases. Early exits keep logic linear and readable.
+
+### 3. Use async/await instead of callbacks
+All async code should use Swift's `async/await`. Callbacks (completion handlers, closures passed as the last argument to async operations) are the old pattern and should not appear in new code.
+
+**Why:** Callbacks invert control flow and make error handling inconsistent. `async/await` reads top-to-bottom, composes cleanly with `try`, and eliminates callback pyramids.
+
+### 4. Clear logic path — name functions after intent, not implementation
+Function names should describe *what* the function achieves (its intention), not *how* it works technically. For example, a function that registers a hook to make an interaction fail should be called `failInteraction()`, not `registerFailureHook()`.
+
+Also: add comments on functions you can't rename (e.g. protocol requirements, library callbacks) explaining their intent.
+
+**Why:** Implementation details change; intent rarely does. Names tied to mechanics rot as the code evolves and obscure the purpose at the call site.
+
+### 5. Public methods before private methods
+`public` and `open` methods should appear near the top of a type, just below `init`.
+
+Correct order within a type:
+1. `init`
+2. `public`/`open` methods
+3. `internal` methods
+4. `private` methods
+
+**Why:** Public APIs are hard to change post-release. Placing them at the top makes the contract immediately visible to reviewers without scrolling past implementation details.
+
+### 6. Doc comments on all public/open declarations
+Every `public` or `open` function, property, type, and initializer must have a doc comment (`///`). This includes anything a consumer of the library will interact with directly.
+
+**Why:** Doc comments are the contract between the library and its callers. They surface in Xcode "Quick help" and are the first place a consumer looks before reading source.
+
+### 7. Use shorthand `guard/if let foo` (not `guard/if let foo = foo`)
+Modern Swift (5.7+) supports shorthand optional binding. Prefer:
+```swift
+guard let foo else { return }   // ✅
+if let bar { ... }              // ✅
+```
+Over:
+```swift
+guard let foo = foo else { return }  // ❌ verbose
+if let bar = bar { ... }             // ❌ verbose
+```
+
+**Why:** The redundant `= foo` is noise. Shorthand is idiomatic Swift 5.7+ and reduces visual clutter.
+
+### 8. No force-unwrap (`!`) in production code
+Force-unwrap crashes the app if the value is nil. Use `guard let`, `if let`, or `??` with an appropriate default instead. The only acceptable use of `!` is in test code (where crashes surface bugs immediately).
+
+**Why:** Force-unwrap turns a recoverable nil into an unrecoverable crash. Production code should handle unexpected nils gracefully rather than terminating.
+
+### 9. Boolean variables should start with `is` (preferred), `has`, `should`, `can`, or similar
+Boolean variables and properties should read as a yes/no question. E.g. `isEnabled`, `hasLoaded`, `shouldRetry`. Avoid names like `enabled`, `loaded`, `retry`.
+
+**Why:** Boolean names that read as questions make conditionals (`if isEnabled`) self-explanatory. Noun/verb names (`if enabled`, `if retry`) are ambiguous and harder to parse at a glance.
+
+### 10. Use `private extension String` for string constants (not bare string literals or `enum`)
+Define file-scoped string constants as a `private extension String` at the bottom of the file:
+```swift
+// ✅ Prefer
+private extension String {
+    static let submitTitle = "Submit"
+}
+// Usage — shorthand works wherever String is accepted
+button.setTitle(.submitTitle)
+```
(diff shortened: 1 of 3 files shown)


.claude/skills/ios-review/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+### 1. Function length ≤ 50 lines
+Functions should be 50 lines or fewer. If a function exceeds this, suggest extracting logical chunks into named helpers.
+
+**Why:** Long functions are harder to read, test, and maintain. Shorter functions with clear names make the call site self-documenting.
+
+### 2. Avoid nested conditionals
+Nested conditionals (`if`, `guard`, `switch`, `for`, `while`, `repeat`) are a bad pattern in iOS code. Prefer:
+- `guard`/`else` to exit early instead of nesting
+- Extracting conditions into well-named boolean variables or functions
+- AI-assisted rewrites when nesting is deep
```
JunDangIntuit, line 40: One level of nesting is natural and readable. The real problem is deep nesting (3+ levels). Recommendation: Flag nesting at 2+ levels deep, not 1+.

.claude/skills/ios-review/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+- **Suggestion**: concrete fix
+
+If no violations are found, say so clearly.
+
+---
+
+## General Rules
+
+### 1. Function length ≤ 50 lines
+Functions should be 50 lines or fewer. If a function exceeds this, suggest extracting logical chunks into named helpers.
```
JunDangIntuit: prefer a soft guideline ("functions should generally be 50 lines or fewer")
some functions (e.g., a switch over many enum cases, or a complex configure() method) might exceed 50 lines and still be perfectly clear.

.claude/skills/ios-review/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+**Why:** Implementation details change; intent rarely does. Names tied to mechanics rot as the code evolves and obscure the purpose at the call site.
+
+### 5. Public methods before private methods
+`public` and `open` methods should appear near the top of a type, just below `init`.
+
+Correct order within a type:
+1. `init`
+2. `public`/`open` methods
+3. `internal` methods
+4. `private` methods
```
JunDangIntuit, line 65: nit:  add a note recommending // MARK: - sections to make the grouping scannable

.claude/skills/ios-review/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+
+**Why:** Public APIs are hard to change post-release. Placing them at the top makes the contract immediately visible to reviewers without scrolling past implementation details.
+
+### 6. Doc comments on all public/open declarations
+Every `public` or `open` function, property, type, and initializer must have a doc comment (`///`). This includes anything a consumer of the library will interact with directly.
+
+**Why:** Doc comments are the contract between the library and its callers. They surface in Xcode "Quick help" and are the first place a consumer looks before reading source.
+
```
JunDangIntuit: agree use /// format, may add: never /** */ block comments.

.claude/skills/ios-review/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+Modern Swift (5.7+) supports shorthand optional binding. Prefer:
+```swift
+guard let foo else { return }   // ✅
+if let bar { ... }              // ✅
+```
+Over:
+```swift
+guard let foo = foo else { return }  // ❌ verbose
+if let bar = bar { ... }             // ❌ verbose
+```
```
JunDangIntuit: suggest this would be a preference, not a hard requirement. guard let foo = foo else is not wrong -- it's just more visual noise :)

.claude/skills/ios-review/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+private extension String {
+    static let submitTitle = "Submit"
+}
+// Usage — shorthand works wherever String is accepted
+button.setTitle(.submitTitle)
+```
+Exception: use an `enum` (not `String` extension) for **closed, finite sets** where the compiler should enforce exhaustiveness — e.g. state type discriminators (`VIEW`, `ACTION`, `EXTERNAL`). Those are better modeled as enums so the compiler catches missing cases.
+
+**Why:** The `private extension String` pattern enables the clean `.foo` shorthand at every call site without polluting the global namespace. Enums are reserved for cases where exhaustiveness matters — mixing the two degrades the compiler's ability to catch missing cases.
+
```
JunDangIntuit: I recently searched a bit regarding this, so from this https://google.github.io/swift/ , 
 To define a “namespace” to group a set of related declarations, such as constants or helper functions."
Also, from google:
Image

I tend to avoid using string extensions because they can make autocomplete a bit confusing alongside the existing String API. For example, when typing . in a String context, custom constants show up mixed in with the standard methods:
button.setTitle(.sub  // autocomplete shows:
                      //   .submitTitle     ← the constant defined
                      //   .subscript(...)  ← String API
                      //   .suffix(...)     ← String API
It may lead to some semantic confusion and makes it trickier to keep things naturally grouped in a way that’s easy to follow.

KVSRoyal: Interesting. I'm good to change this to the caseless `enum`s then. Although the how to in this image is misleading.  Seems like it should be https://google.github.io/swift/#nesting-and-namespacing:

```swift
enum Dimensions {
  static let tileMargin: CGFloat = 8
  static let tilePadding: CGFloat = 4
  static let tileContentSize: CGSize(width: 80, height: 64)
}
```

Also this page is an awesome style guide. What are your thoughts on leveraging that and defining our own additional ones on top?

.claude/skills/ios-review/SKILL.md
```diff
@@ -0,0 +1,177 @@
...
+        self?.handle()
+    }
+}
+```
+
+**Why:** Forgetting `[weak self]` inside a `Task` creates a retain cycle even when the enclosing closure correctly uses `[weak self]`. The two capture lists are independent.
+
+---
+
+## UI Rules
```
JunDangIntuit, line 103: is this #14?

JunDangIntuit:
I like this initiative — it looks great overall! One suggestion: it'd be worth cross-referencing these two guides as we refine the rules, since a few of our conventions touch on areas they cover explicitly (naming, force-unwraps, namespacing constants, etc.):

https://www.swift.org/documentation/api-design-guidelines/
https://google.github.io/swift/

JunDangIntuit:
I like this initiative — it looks great overall! One suggestion: it'd be worth cross-referencing these two guides as we refine the rules, since a few of our conventions touch on areas they cover explicitly (naming, force-unwraps, namespacing constants, etc.):

https://www.swift.org/documentation/api-design-guidelines/
https://google.github.io/swift/

---- Sorry, a mistake, I accidentally closed this pr, now reopened.

.claude/skills/ios-review/SKILL.md
JunDangIntuit: Yes, I like your suggestion. I also added another comment at the end to reference this doc and another one.
By the way, the image was from a quick Google search (so, two different sources 🙂).

.claude/skills/ios-review/SKILL.md
KVSRoyal: I think it's kind of hard to enforce soft guidelines via this, but I think we can always choose to just ignore the feedback if we think it's irrelevant. 😅 

I did change the phrasing to say "generally" though.

.claude/skills/ios-review/SKILL.md
KVSRoyal: I have mixed feelings about `// MARK: -`. Most of the time I think it's unnecessary. Open to it though.

What benefit do you think we'd get? / Can you expand on "scannable"? I know it adds some little markers in the minimap but I, personally, rarely find them useful. Hence why I don't have any current recommendations related to `// MARK:` in here.

.claude/skills/ios-review/SKILL.md
KVSRoyal: I think `/** */` are okay. What's your opinion against them?

.claude/skills/ios-review/SKILL.md
KVSRoyal: Technically, nothing in here is a hard requirement. They're all just our preferences for consistency. But since it's easy to fix I think it's fine to leave it in. Would be nice if we could just have our linter handle this though since it's pretty small.

Actually, let me look into that briefly 🤔 .

.claude/skills/ios-review/SKILL.md
KVSRoyal: Oh yeah, good catch

.claude/skills/ios-review/SKILL.md
KVSRoyal: So we can definitely just enforce it through swiftlint, which is nice. I'll make an issue for us to update our SwiftLint later since I don't want to bloat this PR with a bunch of re-formatting.

.claude/skills/ios-review/SKILL.md
JunDangIntuit: It is also from https://google.github.io/swift/, in which: "Documentation comments are written using the format where each line is preceded by a triple slash (///). Javadoc-style block comments (/** ... */) are not permitted."

Also from google ai overview:
Screenshot 2026-04-21 at 12 47 08 PM

I think we might say (///) is preferred to /** ... */ ?

JunDangIntuit: like, we would like to avoid sth like this:
Screenshot 2026-04-21 at 12 58 12 PM

.claude/skills/ios-review/SKILL.md
JunDangIntuit: It is from the guide too: "When deciding on the logical order of members, it can be helpful for readers and future writers (including yourself) to use // MARK: comments to provide descriptions for that grouping. These comments are also interpreted by Xcode and provide bookmarks in the source window's navigation bar. (Likewise, // MARK: -, written with a hyphen before the description, causes Xcode to insert a divider before the menu item.)"
The scannable basically is on the jumpbar, for example, I added // MARK: - SwfitUIPlayer, // MARK: - Context, then you can navigate it with the jumpbar:

Image

JunDangIntuit: but what you said is correct when for small, well-organized types, // MARK: adds noise without much benefit. It's most valuable in large views (both SwiftUI and UIKit), coordinators

.claude/skills/ios-review/SKILL.md
KVSRoyal: Huh. Okay. Well, if it's in the guide then it'll be enforced so down to roll with it.

.claude/skills/ios-review/SKILL.md
KVSRoyal: OOh, well if it's in the guide I'll just leave it at that then, since the guide will enforce it. Does that solve this?

.claude/skills/ios-review/SKILL.md
KVSRoyal: Forgot to drop the original string rule I just realized. Dropping now.

JunDangIntuit (dismissed):

JunDangIntuit (approved):

KetanReddy (approved):

PR #850, opened 2026-04-21
Refactor stateType from String to NavigationFlowStateType enum

JunDangIntuit:
- Add `NavigationFlowStateType` enum in its own file (`NavigationFlowStateType.swift`), matching the JVM project structure (`NavigationFlowStateType.kt`).
- Use an `unknown(String)` case with associated value for forward-compatibility: unrecognized state types from JS are preserved rather than discarded, with an assertionFailure in debug builds to catch contract mismatches early.
- added unit tests

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/ios/core/Sources/Types/Core/NavigationFlowStateType.swift
+++ b/ios/core/Sources/Types/Core/NavigationFlowStateType.swift
@@ -1,0 +1,53 @@
+import Foundation
+
+/// All possible enumerations of navigation flow states.
+///
+/// Maps to the `state_type` field in the JSON flow definition (e.g. `"state_type": "VIEW"`).
+/// Mirrors `NavigationFlowStateType` on the JVM/Android side.
+public enum NavigationFlowStateType: Equatable {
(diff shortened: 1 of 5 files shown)


KVSRoyal (dismissed):

ios/core/Sources/Types/Core/NavigationFlowStateType.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
```
KVSRoyal, line 1: Changes look good. Seems like some of the tests are failing CI/CD though?

JunDangIntuit: The tests passed, may you review again, thanks

KVSRoyal (approved):

PR #853, opened 2026-04-23
guard against async released exceptions when retrieving state

sugarmanz:
We can guard against released exceptions during state retrieval since we can just return `ReleasedState`. We might need a better solve for other APIs, since invoking them has the potential to crash user applications if not handled by the consumers, but that's a tougher problem.

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

# Release Notes

Guard against released exceptions when retrieving state since we can turn that into a `ReleasedState`.

### DIFF
--- a/jvm/hermes/src/main/jni/BUILD
+++ b/jvm/hermes/src/main/jni/BUILD
@@ -35,7 +35,7 @@ cc_library(
     copts = CMAKE_BUILD_TYPE_COPTS + select({
         # TODO: Using React Natives JSI doesn't contain microtask - come back to this if we build it ourselves
         ":should_include_jni": [
-            "-DJSI_MICROTASK"
+            "-DJSI_MICROTASK",
         ],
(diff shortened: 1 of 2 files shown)


KetanReddy (approved):

jvm/hermes/src/main/jni/BUILD
```diff
@@ -45,9 +45,12 @@ cc_library(
         # Android build
         "//conditions:default": [
             "-Wl,-z,max-page-size=16384",
-            "-lunwind"
+            "-lunwind",
         ],
     }),
+    tags = [
+        "manual",
+    ],
```
sugarmanz, line 53: This "fixes" `bazel test //jvm/...`

brocollie08 (approved):

PR #854, opened 2026-04-27
Automatically lint swift files in CI/CD, format on pre-commit

### LINKED ISSUE #849: More robust SwiftLint
**Is your feature request related to a problem? Please describe.**

Our SwiftLint settings are currently pretty minimal and not particularly helpful.

**Describe the solution you'd like**

We should make them more robust by introducing appropriate rules and enforcing them by setting the severity to `error`.

We may also want to add automatic formatting to commits or in CI/CD so instead of reacting to lint errors in CI/CD, we proactively fix them before they get to CI/CD to reduce PR churn.

**Additional context**

https://github.com/player-ui/player/pull/848 introduces some SwiftLint rules. This issue and that PR are part of the effort to reduce some of the manual review work required since AI has led to more PRs and more daily reviews required.

KVSRoyal:
## What

> [!NOTE]
> Update: this now uses `SwiftLint` and `SwiftFormat` directly via Swift Package Manager.

Resolves https://github.com/player-ui/player/issues/849.

Automatic formatting will be enforced for `.swift` files using a combination of `SwiftLint` and `SwiftFormat`. Here is the developer flow:

1. Local: Run formatting locally with `just format-ios`
2. Pre-commit: When a dev commits any Swift files, formatting will be run via a husky pre-commit hook
3. CI/CD: CI/CD will run a lint check and fail if there are any issues

## Why

We would like to enable automatic formatting to catch some basic ios issues before review. This will reduce PR review burden.

Specifically, we chose to go with a mix of `SwiftLint` and `SwiftFormat` to leverage each one's specialties. `SwiftLint` is better at catching logic issues that cannot be automatically fixed, like force unwrapping, while `SwiftFormat` is better at fixing things automatically.

## Other Options Considered

- [`rules_swiftformat`](https://github.com/cgrindel/rules_swiftformat):
    - Pros: This is hermetic (does not require SwiftFormat on the machine). This is actually the recommended way according to the SwiftFormat github repo with Bazel, [link](https://github.com/nicklockwood/SwiftFormat#bazel-build)
    - Cons: has not been maintained by a human since 2022. It does get some dependabot updates though
    - Blocking Problem: not in the bazel registry, doesn't appear to work with new MODULE-based Bazel
- Fork `rules_swiftformat`:
    - Pros: hermetic
    - Cons: We would have to maintain this, including adding it to the bazel registry
- [`rules_lint`](https://github.com/aspect-build/rules_lint) (replacement for [`bazel-super-formatter`](https://github.com/aspect-build/bazel-super-formatter):
    - Pros: Actively maintained (last commit 3 weeks ago).
    - Cons: non-hermetic (depends on local SwiftFormat). Using local SwiftFormat means we'll all probably have different linters running
- Use the `SwiftFormat` Swift package ([link](https://github.com/nicklockwood/SwiftFormat#swift-package-manager-plugin)) and have a rules_player rule leverage it:
    - Pros: Easy way to align on the same version across machines
    - Cons: non-hermetic.
    - Blocking Problem: We apparently might need to reference the binary directly. Referencing the binary directly increases the rule complexity a lot
- Use `swift-format` via Xcode
    - Pros: Built into Xcode, is a standard
    - Cons: Not-hermetic. Formatting will vary across different Xcode versions, even minor versions. Hard to align on versions across devs---doesn't fit our current development style well.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,9 +164,7 @@ jobs:
               $(bazel query "kind(ios_unit_test, //ios/...)") \
               $(bazel query "kind(ios_unit_test, //plugins/...)") \
               $(bazel query "kind(ios_ui_test, //ios/...)") \
-              $(bazel query "kind(ios_ui_test, //plugins/...)") \
-              $(bazel query "attr(name, '.*SwiftLint', //ios/...)") \
-              $(bazel query "attr(name, '.*SwiftLint', //plugins/...)")
+              $(bazel query "kind(ios_ui_test, //plugins/...)")
       - run:
           name: Tests
           command: |
@@ -181,18 +179,11 @@ jobs:
               $(bazel query "kind(ios_ui_test, //ios/...)") \
               $(bazel query "kind(ios_ui_test, //plugins/...)")
       - run:
-          name: Lint
-          command: |
-            bazel test \
-              --config=ci-mac \
-              $(bazel query "attr(name, '.*SwiftLint', //ios/...)") \
-              $(bazel query "attr(name, '.*SwiftLint', //plugins/...)")
-
-      - run:
-          name: SwiftFormat
+          name: Swift Lint and Format
           command: |
             mint bootstrap
-            just swift-format-lint
+            mint run swiftlint lint --strict
+            mint run swiftformat ios plugins --lint
 
       - run:
           when: always
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,7 +147,7 @@ jobs:
       - run:
           name: Homebrew Dependencies
           command: |
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install bazelisk lcov
+            HOMEBREW_NO_AUTO_UPDATE=1 brew install bazelisk lcov mint
 
       - macos/preboot-simulator:
           version: "26.4"
@@ -187,6 +187,12 @@ jobs:
               --config=ci-mac \
               $(bazel query "attr(name, '.*SwiftLint', //ios/...)") \
               $(bazel query "attr(name, '.*SwiftLint', //plugins/...)")
+
+      - run:
+          name: SwiftFormat
+          command: |
+            mint bootstrap
+            just swift-format-lint
 
       - run:
           when: always
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,7 +147,7 @@ jobs:
       - run:
           name: Homebrew Dependencies
           command: |
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install bazelisk lcov mint
+            HOMEBREW_NO_AUTO_UPDATE=1 brew install bazelisk lcov
 
       - macos/preboot-simulator:
           version: "26.4"
@@ -179,13 +179,6 @@ jobs:
               $(bazel query "kind(ios_ui_test, //ios/...)") \
               $(bazel query "kind(ios_ui_test, //plugins/...)")
       - run:
-          name: Swift Lint and Format
-          command: |
-            mint bootstrap
-            mint run swiftlint lint --strict
-            mint run swiftformat ios plugins --lint
-
-      - run:
           when: always
           command: |
             RESULTS_DIR=_test_results
@@ -209,6 +202,25 @@ jobs:
 
       - store_test_results:
           path: _test_results
+
+  # We run this stage in parallel with build-ios-trunk so it doesn't slow down the build
+  lint_ios:
+    executor: ios
+    steps:
+      - attach_workspace:
+          at: ~/player
+
+      - run:
+          name: Homebrew Dependencies
+          command: |
+            HOMEBREW_NO_AUTO_UPDATE=1 brew install mint
+
+      - run:
+          name: Swift Lint and Format
+          command: |
+            mint bootstrap
+            mint run swiftlint lint --strict
+            mint run swiftformat ios plugins --lint
 
   test:
     executor: base
@@ -443,6 +455,26 @@ workflows:
             - build-fork
             - bazelrc_fork
 
+      - lint_ios:
+          name: lint-ios-trunk
+          filters:
+            branches:
+              ignore:
+                - /pull\/.*/
+          requires:
+            - bazelrc
+            - build-trunk
+
+      - lint_ios:
+          name: lint-ios-fork
+          filters:
+            branches:
+              only:
+                - /pull\/.*/
+          requires:
+            - build-fork
+            - bazelrc_fork
+
       - maybe_release:
           filters:
             branches:
@@ -453,6 +485,7 @@ workflows:
           requires:
             - test-trunk
             - build-ios-trunk
+            - lint-ios-trunk
             - android-test-trunk
 
       - test:
@@ -577,6 +610,10 @@ workflows:
           requires:
             - build
 
+      - lint_ios:
+          requires:
+            - build
+
       - android_test:
           context:
             - Build
@@ -611,6 +648,7 @@ workflows:
             - test
             - bundle-analyze
             - build_and_test_ios
+            - lint_ios
             - build_docs
 
   full_release:
@@ -630,6 +668,10 @@ workflows:
             - setup
 
       - build_and_test_ios:
+          requires:
+            - bazelrc
+
+      - lint_ios:
           requires:
             - bazelrc
 
@@ -666,4 +708,5 @@ workflows:
             - test
             - bundle_analyze
             - build_and_test_ios
+            - lint_ios
             - build_docs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,9 +162,7 @@ jobs:
             bazel build \
               --config=ci-mac \
               $(bazel query "kind('ios_unit_test|ios_ui_test', //ios/...)") \
-              $(bazel query "kind('ios_unit_test|ios_ui_test', //plugins/...)") \
-              $(bazel query "attr(name, '.*SwiftLint', //ios/...)") \
-              $(bazel query "attr(name, '.*SwiftLint', //plugins/...)")
+              $(bazel query "kind('ios_unit_test|ios_ui_test', //plugins/...)")
       - run:
           name: Tests
           command: |
@@ -176,14 +174,6 @@ jobs:
               --test_timeout=1800 \
               $(bazel query "kind('ios_unit_test|ios_ui_test', //ios/...)") \
               $(bazel query "kind('ios_unit_test|ios_ui_test', //plugins/...)")
-      - run:
-          name: Lint
-          command: |
-            bazel test \
-              --config=ci-mac \
-              $(bazel query "attr(name, '.*SwiftLint', //ios/...)") \
-              $(bazel query "attr(name, '.*SwiftLint', //plugins/...)")
-
       - run:
           when: always
           command: |
@@ -208,6 +198,25 @@ jobs:
 
       - store_test_results:
           path: _test_results
+
+  # We run this stage in parallel with build-ios-trunk so it doesn't slow down the build
+  lint_ios:
+    executor: ios
+    steps:
+      - attach_workspace:
+          at: ~/player
+
+      - run:
+          name: Homebrew Dependencies
+          command: |
+            HOMEBREW_NO_AUTO_UPDATE=1 brew install mint
+
+      - run:
+          name: Swift Lint and Format
+          command: |
+            mint bootstrap
+            mint run swiftlint lint --strict
+            mint run swiftformat ios plugins --lint
 
   test:
     executor: base
@@ -469,6 +478,26 @@ workflows:
             - build-fork
             - bazelrc_fork
 
+      - lint_ios:
+          name: lint-ios-trunk
+          filters:
+            branches:
+              ignore:
+                - /pull\/.*/
+          requires:
+            - bazelrc
+            - build-trunk
+
+      - lint_ios:
+          name: lint-ios-fork
+          filters:
+            branches:
+              only:
+                - /pull\/.*/
+          requires:
+            - build-fork
+            - bazelrc_fork
+
       - maybe_release:
           filters:
             branches:
@@ -479,6 +508,7 @@ workflows:
           requires:
             - test-trunk
             - build-ios-trunk
+            - lint-ios-trunk
             - android-test-trunk
 
       - test:
@@ -606,6 +636,10 @@ workflows:
             - build
             - test
 
+      - lint_ios:
+          requires:
+            - build
+
       - android_test:
           context:
             - Build
@@ -647,6 +681,7 @@ workflows:
             - test
             - bundle-analyze
             - build_and_test_ios
+            - lint_ios
             - build_docs
 
   full_release:
@@ -666,6 +701,10 @@ workflows:
             - setup
 
       - build_and_test_ios:
+          requires:
+            - bazelrc
+
+      - lint_ios:
           requires:
             - bazelrc
 
@@ -702,4 +741,5 @@ workflows:
             - test
             - bundle_analyze
             - build_and_test_ios
+            - lint_ios
             - build_docs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -462,6 +462,7 @@ workflows:
           requires:
             - bazelrc
             - build-trunk
+            - lint-ios-trunk
 
       - build_and_test_ios:
           name: build-ios-fork
@@ -472,6 +473,7 @@ workflows:
           requires:
             - build-fork
             - bazelrc_fork
+            - lint-ios-fork
 
       - lint_ios:
           name: lint-ios-trunk
@@ -501,7 +503,6 @@ workflows:
           requires:
             - test-trunk
             - build-ios-trunk
-            - lint-ios-trunk
             - android-test-trunk
 
       - test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,7 +209,7 @@ jobs:
       - run:
           name: Homebrew Dependencies
           command: |
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install mint
+            HOMEBREW_NO_AUTO_UPDATE=1 brew install mint@0.18.0
 
       - run:
           name: Swift Lint and Format
@@ -486,7 +486,6 @@ workflows:
                 - /pull\/.*/
           requires:
             - bazelrc
-            - build-trunk
 
       - lint_ios:
           name: lint-ios-fork
@@ -495,7 +494,6 @@ workflows:
               only:
                 - /pull\/.*/
           requires:
-            - build-fork
             - bazelrc_fork
 
       - maybe_release:
@@ -638,7 +636,7 @@ workflows:
 
       - lint_ios:
           requires:
-            - build
+            - bazelrc
 
       - android_test:
           context:
@@ -681,7 +679,6 @@ workflows:
             - test
             - bundle-analyze
             - build_and_test_ios
-            - lint_ios
             - build_docs
 
   full_release:
@@ -741,5 +738,4 @@ workflows:
             - test
             - bundle_analyze
             - build_and_test_ios
-            - lint_ios
             - build_docs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -207,18 +207,11 @@ jobs:
           at: ~/player
 
       - run:
-          name: Homebrew Dependencies
-          # mint has no versions in homebrew, so pin the homebrew version with it
-          # This is mint 0.18.0
-          command: |
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/f682012776a07b711896a7294bf262534f0dfc98/Formula/m/mint.rb
-
-      - run:
           name: Swift Lint and Format
           command: |
-            mint bootstrap
-            mint run swiftlint lint --strict
-            mint run swiftformat ios plugins --lint
+            swift run --package-path xcode mint bootstrap
+            swift run --package-path xcode mint run swiftlint lint --strict ios plugins
+            swift run --package-path xcode mint run swiftformat ios plugins --lint
 
   test:
     executor: base
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,9 +209,8 @@ jobs:
       - run:
           name: Swift Lint and Format
           command: |
-            swift run --package-path xcode mint bootstrap
-            swift run --package-path xcode mint run swiftlint lint --strict ios plugins
-            swift run --package-path xcode mint run swiftformat ios plugins --lint
+            swift run --package-path xcode swiftlint lint --strict ios plugins
+            swift run --package-path xcode swiftformat ios plugins --lint
 
   test:
     executor: base
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,8 +208,10 @@ jobs:
 
       - run:
           name: Homebrew Dependencies
+          # mint has no versions in homebrew, so pin the homebrew version with it
+          # This is mint 0.18.0
           command: |
-            HOMEBREW_NO_AUTO_UPDATE=1 brew install mint@0.18.0
+            HOMEBREW_NO_AUTO_UPDATE=1 brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/f682012776a07b711896a7294bf262534f0dfc98/Formula/m/mint.rb
 
       - run:
           name: Swift Lint and Format
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -462,7 +462,6 @@ workflows:
           requires:
             - bazelrc
             - build-trunk
-            - lint-ios-trunk
 
       - build_and_test_ios:
           name: build-ios-fork
@@ -473,7 +472,6 @@ workflows:
           requires:
             - build-fork
             - bazelrc_fork
-            - lint-ios-fork
 
       - lint_ios:
           name: lint-ios-trunk
--- a/.lintstagedrc.cjs
+++ b/.lintstagedrc.cjs
@@ -6,4 +6,8 @@ module.exports = {
   '*.md': (filenames) => {
     return [`prettier --write ${filenames.join(' ')}`];
   },
+  '*.swift': (filenames) => {
+    const files = filenames.join(' ');
+    return [`mint run swiftformat ${files}`];
+  },
 };
--- a/.lintstagedrc.cjs
+++ b/.lintstagedrc.cjs
@@ -6,4 +6,8 @@ module.exports = {
   '*.md': (filenames) => {
     return [`prettier --write ${filenames.join(' ')}`];
   },
+  '*.swift': (filenames) => {
+    const files = filenames.join(' ');
+    return [`mint run swiftformat ${files}`, `mint run swiftlint --fix`];
+  },
 };
--- a/.lintstagedrc.cjs
+++ b/.lintstagedrc.cjs
@@ -8,6 +8,6 @@ module.exports = {
   },
   '*.swift': (filenames) => {
     const files = filenames.join(' ');
-    return [`mint run swiftformat ${files}`];
+    return [`mint run swiftformat ${files}`, `mint run swiftlint --fix`];
   },
 };
--- a/.lintstagedrc.cjs
+++ b/.lintstagedrc.cjs
@@ -8,6 +8,9 @@ module.exports = {
   },
   '*.swift': (filenames) => {
     const files = filenames.join(' ');
-    return [`mint run swiftformat ${files}`, `mint run swiftlint --fix ${files}`];
+    return [
+      `swift run --package-path xcode mint run swiftformat ${files}`,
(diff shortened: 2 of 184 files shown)


tmarmer:
Regarding `rules_swiftformat`:
> Blocking Problem: not in the bazel registry so it's really hard, if not impossible to use

I thought we could target by a commit for rules repos not registered on the central bazel rules repository. The quickstart guide on the rules repo lists this [as a way to install it](https://github.com/cgrindel/rules_swiftformat#1-configure-your-workspace-to-use-rules_swiftformat) too. If it's too much to manage, I think `mint` looks like the best alternative but want to make sure I understand exactly what's stopping us from using the recommended approach.

.circleci/config.yml
```diff
@@ -147,7 +147,7 @@ jobs:
       - run:
```
KVSRoyal, line 147: > Regarding rules_swiftformat:
> 
> Blocking Problem: not in the bazel registry so it's really hard, if not impossible to use
> 
> I thought we could target by a commit for rules repos not registered on the central bazel rules repository. The quickstart guide on the rules repo lists this [as a way to install it](https://github.com/cgrindel/rules_swiftformat#1-configure-your-workspace-to-use-rules_swiftformat) too. If it's too much to manage, I think mint looks like the best alternative but want to make sure I understand exactly what's stopping us from using the recommended approach.

@tmarmer Yeah, it does list that but:
1. I tried it and it didn't work
2. I don't think this supports the new Bazel format. I got some errors I thought might be to that effect when trying
3. The example says to use the last release, which is 0.4.1, which is from 2022. This is out of date and not really maintained

tmarmer: Well if it doesn't work, that's a good enough reason to do something else XD
If mint works and it's easy to keep everything up to date than I think that's a good approach.

cehan-Chloe:
My testing observation: I had two swift files with formatting issue, one staged one unstaged. running commit, the expected behavior is the commit failed with fixed version unstaged and an error about the linting and get it fixed since we have `mint run swiftlint --fix`. now formatting issue is still there and the error is 
```
➜  player git:(issue-849) ✗ git commit -m "test"
✔ Preparing...
✔ Running tasks...
✖ Prevented an empty git commit!
✔ Reverting to original state because of errors...
✔ Cleaning up...

  ⚠ lint-staged prevented an empty git commit.
  Use the --allow-empty option to continue, or check your task configuration

husky - pre-commit hook exited with code 1 (error) 
```
i think it's because of the formatting issue is already fixed and no file change to commit then that leaves it an empty git commit. But it looks confusing since no linting error is pointed out or fixed and I still have staged change. Recording:
https://drive.google.com/file/d/1Vkkf8u2lIk9DJda6t_-GHtzkDtnnlOeg/view?usp=drive_link

.circleci/config.yml
KVSRoyal, line 144: Unrelated to this line, just starting a reply-able thread:

https://github.com/player-ui/player/pull/854#issuecomment-4371840081

@cehan-Chloe I think that's okay in this case? Because presumably irl we won't be committing a file with no meaningful changes.

This example is: 
1. Break format in 1 file by adding a space or something. Make no meaningful changes
2. Try to commit
3. This will break

But if we did something closer to actual development (tested locally):
1. Break format in 1 file and make a meaningful change in the same file or another file
2. Try to commit
3. Works ✅

cehan-Chloe: ya. that's fair point.

cehan-Chloe (approved):

cehan-Chloe (dismissed):

JunDangIntuit (dismissed):

intuit-svc:
## Benchmark Results

_Comparison against baseline from `main`. ⚠️ = regression (>10% slower), ✅ = improvement (>5% faster)_

### `core/player` ⚠️

| Benchmark | Current | Baseline | Change |
|-----------|---------|----------|--------|
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.bar` | 703.07K ops/s | 785.81K ops/s | -10.5% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets.1.name` | 487.92K ops/s | 546.75K ops/s | -10.8% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets.01.name` | 501.03K ops/s | 516.41K ops/s | -3.0% |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets['01'].name` | 436.88K ops/s | 436.51K ops/s | +0.1% |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets[01].name` | 499.93K ops/s | 443.38K ops/s | +12.8% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets[name = "frodo"].type` | 235.22K ops/s | 236.75K ops/s | -0.6% |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets["name" = "sprinkles"].type` | 189.58K ops/s | 173.33K ops/s | +9.4% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets["isDog" = false].type` | 248.86K ops/s | 207.50K ops/s | +19.9% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets["isDog" = true].type` | 248.90K ops/s | 224.09K ops/s | +11.1% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.bar` | 498.86K ops/s | 408.04K ops/s | +22.3% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets.1.name` | 358.49K ops/s | 290.35K ops/s | +23.5% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets.01.name` | 310.09K ops/s | 251.35K ops/s | +23.4% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets['01'].name` | 327.59K ops/s | 243.43K ops/s | +34.6% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets[01].name` | 380.07K ops/s | 254.35K ops/s | +49.4% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets[name = "frodo"].type` | 220.78K ops/s | 163.06K ops/s | +35.4% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets["name" = "sprinkles"].type` | 169.81K ops/s | 142.23K ops/s | +19.4% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets["isDog" = false].type` | 229.50K ops/s | 189.05K ops/s | +21.4% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets["isDog" = true].type` | 220.39K ops/s | 245.64K ops/s | -10.3% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = 1 + 3 (sync)` | 378.42K ops/s | 433.05K ops/s | -12.6% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = 1 + 3 (async)` | 300.15K ops/s | 324.02K ops/s | -7.4% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: conditional(true, true, false) (sync)` | 425.83K ops/s | 438.21K ops/s | -2.8% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: conditional(true, true, false) (async)` | 354.68K ops/s | 372.14K ops/s | -4.7% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional({{bar}} > 0, true, false) (sync)` | 161.66K ops/s | 154.75K ops/s | +4.5% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional({{bar}} > 0, true, false) (async)` | 138.19K ops/s | 166.54K ops/s | -17.0% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional(conditional(true = false, false, true), conditional(false = false, true, false), conditional(true = true, false, true)) (sync)` | 123.87K ops/s | 150.81K ops/s | -17.9% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional(conditional(true = false, false, true), conditional(false = false, true, false), conditional(true = true, false, true)) (async)` | 123.59K ops/s | 133.70K ops/s | -7.6% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = await(asyncTestFunction(1)) (sync)` | N/A | N/A | N/A |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = await(asyncTestFunction(1)) (async)` | 243.76K ops/s | 240.17K ops/s | +1.5% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = asyncTestFunction(1) (sync)` | 249.89K ops/s | 264.63K ops/s | -5.6% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = asyncTestFunction(1) (async)` | 210.43K ops/s | 265.11K ops/s | -20.6% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: asyncTestFunction(1) (sync)` | 616.23K ops/s | 785.23K ops/s | -21.5% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: asyncTestFunction(1) (async)` | 484.55K ops/s | 684.41K ops/s | -29.2% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional(!{{bar}} == false, await(asyncTestFunction(1)), false) (sync)` | 113.10K ops/s | 178.69K ops/s | -36.7% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional(!{{bar}} == false, await(asyncTestFunction(1)), false) (async)` | 90.53K ops/s | 167.22K ops/s | -45.9% ⚠️ |
| `core/player/src/view/resolver/__tests__/index.bench.ts > resolver benchmarks > initial resolve` | 475.98 ops/s | 582.99 ops/s | -18.4% ⚠️ |
| `core/player/src/view/resolver/__tests__/index.bench.ts > resolver benchmarks > Resolving from cache` | 12.55K ops/s | 18.49K ops/s | -32.1% ⚠️ |
| `core/player/src/view/resolver/__tests__/index.bench.ts > resolver benchmarks > data changes` | 2.03K ops/s | 1.86K ops/s | +9.6% ✅ |
| `core/player/src/view/resolver/__tests__/index.bench.ts > resolver benchmarks > data changes slow` | 531.73 ops/s | 630.50 ops/s | -15.7% ⚠️ |

### `plugins/async-node/core` ⚠️

| Benchmark | Current | Baseline | Change |
|-----------|---------|----------|--------|
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 1 times` | 7.92K ops/s | 14.45K ops/s | -45.2% ⚠️ |
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 5 times` | 9.18K ops/s | 13.02K ops/s | -29.5% ⚠️ |
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 10 times` | 8.50K ops/s | 9.51K ops/s | -10.6% ⚠️ |
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 50 times` | 2.74K ops/s | 2.98K ops/s | -8.1% |
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 100 times` | 1.61K ops/s | 1.70K ops/s | -5.3% |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 1 times` | 5.42K ops/s | 7.61K ops/s | -28.8% ⚠️ |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 5 times` | 5.34K ops/s | 7.99K ops/s | -33.2% ⚠️ |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 10 times` | 5.10K ops/s | 6.13K ops/s | -16.8% ⚠️ |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 50 times` | 2.05K ops/s | 2.30K ops/s | -10.7% ⚠️ |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 100 times` | 1.27K ops/s | 1.38K ops/s | -7.8% |

### `react/player` ⚠️

| Benchmark | Current | Baseline | Change |
|-----------|---------|----------|--------|
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 1 ReactAssets` | 554.30 ops/s | 624.44 ops/s | -11.2% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 1 ReactAssets` | 811.03 ops/s | 1.13K ops/s | -28.4% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 5 ReactAssets` | 551.92 ops/s | 604.13 ops/s | -8.6% |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 5 ReactAssets` | 691.12 ops/s | 1.04K ops/s | -33.2% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 10 ReactAssets` | 563.62 ops/s | 621.51 ops/s | -9.3% |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 10 ReactAssets` | 759.85 ops/s | 907.58 ops/s | -16.3% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 50 ReactAssets` | 459.02 ops/s | 504.89 ops/s | -9.1% |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 50 ReactAssets` | 192.41 ops/s | 279.20 ops/s | -31.1% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 100 ReactAssets` | 357.52 ops/s | 411.71 ops/s | -13.2% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 100 ReactAssets` | 106.90 ops/s | 123.20 ops/s | -13.2% ⚠️ |

package.json
```diff
@@ -13,7 +13,8 @@
   },
   "packageManager": "pnpm@10.18.1",
   "scripts": {
-    "fmt": "prettier --write '**/*.{ts,tsx}'"
+    "fmt": "prettier --write '**/*.{ts,tsx}'",
+    "prepare": "husky install"
```
KetanReddy, line 17: Whats the goal of running husky during the install and packaging?

KVSRoyal: [It's just the recommendation](https://typicode.github.io/husky/how-to.html#manual-setup).

KetanReddy:
> - Cons: non-hermetic (depends on local SwiftFormat). Using local SwiftFormat means we'll all probably have different linters running
>
> Use the SwiftFormat Swift package ([link](https://github.com/nicklockwood/SwiftFormat#swift-package-manager-plugin)) and have a rules_player rule leverage it:
> - Pros: Easy way to align on the same version across machines

Sorry, a bit confused here. Is aligning the version of `SwiftFormat` a problem or not? I might be missing something with how `rules_lint` works...

In general though, while the concerns around the alternate solution's not being hermetic are valid, the biggest pro of some of those are that everything stays within the Bazel ecosystem. My general concern with moving tooling out of bazel is that is starts to further complicate the DevEx of working in a large repository. Having `bazel build` and `bazel test` be the entrypoint for building or testing any package in the repository, regardless of language, is a smoother experience than having to read docs/files to figure out how to run something. Requiring `mint` to be installed is also another step that further complicates the experience as well. _Edit:  also, since the command to install mint via brew doesn't pin a version, this approach also isn't hermetic._

Thats not to say that we shouldn't choose the best tool if its inconvenient, we just need to be intentional about it and understand the broader ramifications for it.

CONTRIBUTING.md
```diff
@@ -159,6 +158,29 @@ iOS Development requires a bit more set-up.
    1. Close and re-open the terminal.
    1. Run `which rsync` to make sure you’re using the homebrew one. If it’s not the homebrew rsync, something is wrong and needs to be fixed.
 
+1. Install git hooks:
+
+   ```bash
+   pnpm run prepare
```
AngelaVilladiego: On my end the command `pnpm run prepare` failed with
```
sh: husky: command not found
 ELIFECYCLE  Command failed.
 WARN   Local package.json exists, but node_modules missing, did you mean to install?
```

A `pnpm install` fixes it but there's no mention in the docs to run that and additionally it creates a tracked file `Package.resolved` so I'm not sure if that's correct. If a `pnpm install` or other command is required could you add those instructions to the file please?

AngelaVilladiego: While you're at it, could you also add `brew install just` to the list of installs needed?

AngelaVilladiego (changes requested):

.swiftlint.yml
```diff
@@ -37,14 +77,11 @@ opt_in_rules:
...
-# # Don't add unnecessary parentheses to control statements
-# control_statement:
-#     severity: error
+shorthand_optional_binding:
+    severity: error
+force_unwrapping:
+    severity: error
+# Don't add unnecessary parentheses to control statements
+control_statement:
+    severity: error
```
AngelaVilladiego, line 63: Claude noted that the "nesting" override was not added here and remains a warning (though under `--strict` it will throw an error). Is there a reason this was removed since the TODO was created?

.lintstagedrc.cjs
```diff
@@ -6,4 +6,8 @@ module.exports = {
   '*.md': (filenames) => {
     return [`prettier --write ${filenames.join(' ')}`];
   },
+  '*.swift': (filenames) => {
+    const files = filenames.join(' ');
+    return [`mint run swiftformat ${files}`, `mint run swiftlint --fix`];
+  },
```
AngelaVilladiego, line 15: Claude suggested adding the `${files}` argument to the swiftlint command as well to avoid running it on unstaged files, is there a reason you're doing so for `swiftformat` and not `swiftlint`?

AngelaVilladiego:
Quick note, there's no version change type in the PR comment selected :)

CONTRIBUTING.md
KVSRoyal: Great suggestions. Not sure what's going on with the `Package.resolved` file though.

`just` is already included in [here](https://github.com/player-ui/player/blob/main/CONTRIBUTING.md#building-and-testing-locally-all-platforms). It's technically optional

.swiftlint.yml
KVSRoyal: Just missed it. Good catch, adding

KVSRoyal: Oh, wait, actually `nesting` doesn't support severity:
```bash
warning: Configuration for 'nesting' rule contains the invalid key(s) 'severity'.
```

.lintstagedrc.cjs
KVSRoyal: I don't remember if there was a reason not to so I've added it 👍🏽 . Good catch.

package.json
KVSRoyal: Reusing this thread to reply to [this](https://github.com/player-ui/player/pull/854#issuecomment-4909698481):

### `rules_lint` being non-hermetic
>Is aligning the version of `SwiftFormat` a problem or not? I might be missing something with how `rules_lint` works...

`rules_lint` uses your _local_ version of `swift-format` (the one that ships with Xcode, which is different from `SwiftFormat`. I typoed in my intial notes.). It does not actually ship the tool, so it's not hermetic. Xcode's `swift-format` changes even with minor versions, so we'd all be running different linters. All of us having the exact same Xcode version is pretty restrictive so I don't think we should do that either.

Screenshot 2026-07-08 at 11 18 30 AM

### Should we pin `mint`?
>Edit: also, since the command to install mint via brew doesn't pin a version, this approach also isn't hermetic.

I didn't pin a version for this because I think the `mint` version doesn't really matter---and newer is probably better. Just like `brew` version doesn't really matter. The thing that actually needs to be pinned is the formatter versions, and those are pinned.

### Overall Thoughts

Regarding the other concerns, my thought is that since the hermetic solutions aren't viable, then the next best is the thing that is:
1. Actively maintained
2. Bazel-integrated if possible
3. Lightest weight

Considering all these things, I think `mint` is the best even though it's not through bazel. I did integrate it in our existing `just` framework so it never has to be used directly.

CONTRIBUTING.md
KVSRoyal: I added `pnpm i`. I didn't do anything about the `Package.resolved` which doesn't appear for me.

KVSRoyal: Just remembered `pnpm i` runs `prepare` as part of it. Updating.

package.json
KVSRoyal: Open to suggestions though.

KetanReddy: IMO we should pin `mint`. In theory it could pull in a new major version with breaking changes opaquely and fail in CI. The brew version should be dictated by the underlying CI image/metal so we don't have to worry about it changing as much. 

Do we have any numbers around the impact to CI times with running this as a separate job?

KVSRoyal: I think it's unlikely that major version changes will be an issue with mint BUT I agree it doesn't hurt to be safe. I'll look into how we can pin it.

For the CI numbers, the step appears to take about 6mins (way slower than running local), but since it's in parallel with the other steps it won't typically slow the pipeline down at all. If everything else is cached, it does appear to push our max from 5 to 6 mins. From what I can tell, most of that time is spent on `mint bootstrap`. Might help to put that in our docker image instead? What do you think?

Screenshot 2026-07-08 at 4 42 34 PM

KetanReddy: Yeah lets give that a shot. If we can bring that down that would be great because that would increase our max build time by 20% (usually runs in around 25m). Is there any cashing mint can do?

KVSRoyal: Can you explain the 20% calculation? That implies this adds 5 mins to the overall build time. I think in general it adds 0-2 min? 0 on most runs and 2 if everything else is cached, including ios.

KVSRoyal: Update: cannot go in the docker image. I remembered that ios actualy does not run on the docker image because it needs to run on a mac environment.

.swiftformat
```diff
@@ -0,0 +1,28 @@
+--swiftversion 5.5
```
JunDangIntuit: do we want to make the swiftVersion to 5.7 as well since in the .swiftlint.yml file you wrote  # Enforces Swift 5.7+ shorthand

plugins/reference-assets/swiftui/Sources/SwiftUI/TextAsset.swift
```diff
@@ -16,50 +13,49 @@ struct TextData: AssetData, Equatable {
...
-        - destination: A string URL destination
-     */
+    let destination: URL?
+
+    /// Constructs a `LinkModifier` with a string destination
+    /// - parameters:
+    ///   - destination: A string URL destination
     init(_ destination: String) {
-        self.destination = URL(string: destination)!
+        self.destination = URL(string: destination)
```
JunDangIntuit, line 50: it is a correct fix to avoid crash with force unwrap. just one suggestion: the failure is now silent. Could we log the offending string when parsing fails? like
init(_ destination: String) {
    self.destination = URL(string: destination)
    if self.destination == nil {
       /// log sth
    }
}

KVSRoyal: Actually, `self.destination` is optional now which should mean that instead of crashing at all, the value will just be nil.

https://github.com/player-ui/player/pull/854/changes/BASE..c8614482f4c17e9bfc06f060ca5c42d4bbaa5440#diff-60096c64102aa723724569fc27682fb7a073c0cd714c61da4eb57f8f9b622e25R44-R50

JunDangIntuit: Right, I saw that. But would it be useful to log some debug information when the parsed URL is nil? Right now, an invalid or empty link parses to nil, and tapping it silently does nothing. Logging something would make it much easier to diagnose malformed URLs.

KVSRoyal: Oooh, I see. Good catch. Yeah, I'll add that

KetanReddy (dismissed):

JunDangIntuit (dismissed):

CONTRIBUTING.md
```diff
@@ -159,6 +158,22 @@ iOS Development requires a bit more set-up.
...
+   pnpm i
+   ```
+
+   These hooks include auto-formatting and will run for all staged files pre-commit.
+
+1. Our linters, ([SwiftLint](https://github.com/realm/SwiftLint) and [SwiftFormat](https://github.com/nicklockwood/SwiftFormat)), are pinned via SPM in `xcode/Package.swift` and run with `swift run --package-path xcode`, so no separate install is needed. Use the `just` commands below.
+
+   | Command                      | Description                                  |
+   | ---------------------------- | -------------------------------------------- |
+   | `just format-ios`            | Auto-fix all lint and format errors in-place |
```
AngelaVilladiego, line 173: Running `just format-ios` on my device results in error:

```
Reading config file at /Users/avilladiego/Documents/GitHub/player/.swiftformat
error: organizeDeclarations rule timed out in /Users/avilladiego/Documents/GitHub/player/ios/test-utils/Sources/unit-test/SwiftUIAssetUnitTestCase.swift.
error: organizeDeclarations rule timed out in /Users/avilladiego/Documents/GitHub/player/ios/core/Sources/CorePlugins/PartialMatchFingerprintPlugin.swift.
error: organizeDeclarations rule timed out in /Users/avilladiego/Documents/GitHub/player/plugins/check-path/ios/Sources/CheckPathPlugin.swift.
error: organizeDeclarations rule timed out in /Users/avilladiego/Documents/GitHub/player/ios/core/Sources/CorePlugins/JSBasePlugin.swift.
error: organizeDeclarations rule timed out in /Users/avilladiego/Documents/GitHub/player/ios/test-utils/Sources/ui-test/AssetUITestCase.swift.
error: organizeDeclarations rule timed out in /Users/avilladiego/Documents/GitHub/player/ios/core/Sources/Types/Beacon.swift.
error: organizeDeclarations rule timed out in /Users/avilladiego/Documents/GitHub/player/ios/logger/Sources/TapableLogger.swift.
error: organizeDeclarations rule timed out in /Users/avilladiego/Documents/GitHub/player/ios/swiftui/Sources/SwiftUIPlayer.swift.
error: recipe `format-ios` failed on line 135 with exit code 70
```

Device specs: Macbook M3 Pro, 36GB RAM, Sequoia 15.7.5

KVSRoyal: We had a convo offline and landed on removing `organizeDeclarations`. It would be nice to have it because it organizes things automatically for us, but if it's not going to work, there's no point. Seems to timeout at random even on machines with pretty good specs.

AngelaVilladiego (approved):
Working and looks good to me!

KetanReddy (approved):

JunDangIntuit (approved):

PR #855, opened 2026-04-27
Manage memory leak

JunDangIntuit:
Fix JSContext memory leak caused by cross-runtime retain cycles between Swift closures and JavaScriptCore.

When navigating between flows in ManagedPlayer, each load() creates a new JSContext on the shared JSVirtualMachine. The old JSContext was never deallocated because @convention(block) closures registered via hook taps, the exception handler, and plugin references created strong reference paths back to the context.
## Changes
- **SwiftUIPlayer.swift**
  - Enhanced `unload()` to fully break references to the old `JSContext`
  - Clears `exceptionHandler`, `setTimeout` polyfill, plugin context/pluginRef
  - Forces `JSGarbageCollect`
  - Wrapped cleanup in `autoreleasepool` to immediately drain ObjC autoreleased temporaries (e.g., `JSValue.context`)

- **HeadlessPlayer.swift**
  - Fixed implicit `self` capture in `attachExceptionHandler`
  - Captures `logger` locally to break retain cycle: `Player → JSContext → exceptionHandler closure → self`

- **Hook.swift**
  - Minor cleanup (`let self = self` → `let self`)

- **Tests**
  - Added ManagedPlayerJSContextLifecycleTests
### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [x] No Update needed
- [ ] Unable to update docs

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
</table>


KVSRoyal:
@claude Review this ios pr

KVSRoyal (approved):
Lgtm

ios/swiftui/Sources/ManagedPlayer/ManagedPlayer.swift
```diff
@@ -150,7 +150,7 @@ internal struct ManagedPlayer14<Loading: View, Fallback: View>: View {
     public var body: some View {
         bodyContent(viewModel.stateTransition.call() ?? .identity)
             .onDisappear {
-                context.clearExceptionHandler()
+                context.unload()
```
nancywu1, line 153: I think we still need to keep context.clearExceptionHandler() as well
because it removes the reference to the logger

I added it in a while back in https://github.com/player-ui/player/pull/486/changes 
unless you are no longer seeing multiple references to the longer anymore after opening and exiting multiple player screens?

JunDangIntuit: yes, the logic is still there in the unload() method:  ctx.exceptionHandler = nil

nancywu1: oh I missed that one, then we can remove the clearExceptionHandler() func

nancywu1:
Also just saw the Claude review talking about the setObject( s `player` key possibly needing to call `Player.Player` (case sensitive?) is that a valid concern?

PR #857, opened 2026-04-27
Remove unnecessary ios wrappers

KVSRoyal:
During the recent plugin audit by player devs, we discovered that the following plugins have unnecessary ios and/or jvm wrappers around core-only functionality:

1. `common-expressions`
2. `common-types`
3. `computed-properties`
4. `stage-revert-data`

This PR removes the unnecessary wrappers. A note has been added to the plugins' docs to document this context and hopefully remind us not to re-add these unnecessary wrappers in the future.

Plugins 1-3 are also now loaded with the core reference assets plugin. `stage-revert-data` is not. That plugin requires `attributes.commitTransitions` to be present (i.e. errors without it). Our demo isn't set up like that and I didn't want to change it.

ADDITIONALLY:

1. Looks like `fast-sort` got dropped from the `package.json` when I merged from `main`, so this restores it.
2. There was also some required linting (no changes to the pnpm-lock file other than `fast-sort`) so that got added to this.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A` Part of 1.0

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


JunDangIntuit (approved):

core/partial-match-registry/package.json
```diff
@@ -1,5 +1,8 @@
 {
   "name": "@player-ui/partial-match-registry",
   "version": "0.0.0-PLACEHOLDER",
-  "main": "src/index.ts"
+  "main": "src/index.ts",
+  "dependencies": {
+    "fast-sort": "^3.4.1"
```
sugarmanz: We shouldn't list dependencies in the `package.json`s since they'll get populated by the Bazel build for publishing.

plugins/computed-properties/core/BUILD
```diff
@@ -11,7 +11,6 @@ vitest_config(name = "vitest_config")
 js_pipeline(
     package_name = "@player-ui/computed-properties-plugin",
     build_deps = NATIVE_BUILD_DEPS,
```
sugarmanz: We should probably not rely on the `NATIVE_BUILD_DEPS` either if we're not bundling for mobile. I think you can just omit `build_deps` and let it use the default.

plugins/common-expressions/core/BUILD
```diff
@@ -11,7 +11,6 @@ vitest_config(name = "vitest_config")
 js_xlr_pipeline(
     package_name = "@player-ui/common-expressions-plugin",
     build_deps = NATIVE_BUILD_DEPS,
```
sugarmanz: We should probably not rely on the NATIVE_BUILD_DEPS either if we're not bundling for mobile. I think you can just omit build_deps and let it use the default.

package.json
```diff
@@ -89,6 +89,7 @@
     "dequal": "^2.0.2",
     "devalue": "^5.4.2",
     "dlv": "^1.1.3",
+    "fast-sort": "^3.4.1",
```
sugarmanz, line 92: Why is this added in this PR?

core/partial-match-registry/package.json
KVSRoyal: Good catch. My bad

package.json
KVSRoyal: I think it got removed during the last merge commit, which is my bad. [It was actually added in this PR](https://github.com/player-ui/player/pull/818)

KVSRoyal: Yup, got [undone here](https://github.com/player-ui/player/commit/ce58a774b65445d529041f634f05720a86771b9f).

KVSRoyal: I double-checked and it looks like all the other files look good. I think I just accidentally took the `package.json` from master instead of resolving the diffs for some reason.

docs/site/src/content/docs/plugins/core/common-expressions.mdx
```diff
@@ -2,19 +2,17 @@
 title: Common Expressions
 ---
 
-import PlatformTabs from "../../../../components/PlatformTabs.astro";
+import { Aside } from "@astrojs/starlight/components";
 import { PackageManagers } from 'starlight-package-managers'
-import SPM from "../../../../components/SPM.astro";
+
+<Aside title="Core-only Plugin">This plugin should be added to the shared core and is not intended to be re-registered across platforms. There are no separate iOS or Android implementations — add it once in your core Player setup.</Aside>
```
sugarmanz: Nit: It'd be great to import this from a shared component if it's duplicated across docs

ios/demo/Sources/App.swift
```diff
@@ -44,7 +43,6 @@ struct MainView: View {
             ReferenceAssetsPlugin(),
             CommonTypesPlugin(),
```
sugarmanz: Oh, wait — isn't `common-types` the same use case as `common-expressions`? We should probably remove that too and surface through `ReferenceAssetsPlugin` since that's the meta-plugin approach.

sugarmanz (approved):

PR #866, opened 2026-05-12
Input Agnostic Player and Native A2UI Support

KetanReddy:
## What and Why

With the emergence of A2UI and other agent-driven standards, Player — in its mission to serve the broader SDUI space — should expand support for more formats beyond its own in order to broaden the scope of what Player solves for. Furthermore, this lets capabilities be built against Player itself, regardless of input format, allowing for the decoupling of implementation from the server driving it.

This PR adds the generic "unknown format" entrypoint to **every** platform (web, JVM, Android, iOS) and ships A2UI as the first non-Player format end-to-end.

## Core Player Changes

- New `transformContent` bail hook on Player's hooks. Fires at the top of `start()`, after plugins are registered and before `resolveFlowContent`. Plugins inspect a `ContentMeta` `{ format, version }` and either convert the payload or pass it through.
- `Player.start()` signature widened from `(payload: Flow)` to `(payload: unknown, options?: StartOptions)`. Default format is `"player"`, which preserves existing behavior (a plain Flow flows through untouched). `version` is free-form, so a single format plugin can dispatch across versions.

### Platform Entrypoints

The hook and signature widening live in the core JS bundle every platform loads; each platform entrypoint forwards the `format`/`version` through. Default `"player"` keeps all existing call sites untouched.

- **React** — `ReactPlayer.start()` mirrors the same signature (`react/player/src/player.tsx`) and forwards options to the underlying `Player.start()`.
- **JVM** — `Player.start()` gains `start(flow: String, format: String, version: String? = null)` (plus a matching `URL` overload). `HeadlessPlayer` implements it and forwards `{ format, version }` to the JS `player.start(payload, options)` as a bridge-encoded `Map`.
- **Android** — `AndroidPlayer.start(flow, format, version)` delegates to the wrapped `HeadlessPlayer`.
- **iOS** — new `StartOptions { format, version }` value type; `HeadlessPlayer.start(flow:options:completion:)` appends the options object to the JS `start` args (`[String: Any]` → JS object via JavaScriptCore). Threaded through `SwiftUIPlayer.init` / `Context.load` / `ManagedPlayer` via a defaulted `startOptions:` param (source-compatible). Use `StartOptions.a2ui`.

### Core: `@player-ui/a2ui-plugin`

Includes three sub-plugins:
- `A2UIContentPlugin`: Only activates when `meta.format === "a2ui"`. Calls `adaptA2UIToFlow(snapshot)` to transform content.
- `A2UITransformPlugin`: Per-asset transforms (Button/TextField/CheckBox/Slider/DateTimeInput/ChoicePicker/Text) that attach `run()`/`set()`/`value` helpers consumed by the rendering layer.
- `A2UIExpressionsPlugin`: Registers the A2UI v0.9.1 standard library: validation (`required`, `regex`, `length`, `numeric`, `email`), formatters (`formatString`, `formatNumber`, `formatCurrency`, `formatDate`, `pluralize`), `openUrl`, and logic (`and`/`or`/`not`).

This package emits a native JS bundle (`A2UIPlugin.native.js`), which the JVM and iOS wrappers below load directly — so the adapter, transforms, and expression std-lib are shared across all platforms with zero reimplementation.

#### Adapter Logic

Walks the flat `components[]` list from `id: "root"`, inlines child references into a nested asset tree matching Player's `{asset: ...}` shape, and produces a Flow with a single VIEW state plus one END per unique event name encountered. Translation rules:

- `{path: "/x/y"}` → `"x.y"` (Player binding)
- `formatString(...)` → `"Hello, {{x.y}}!"` template
- Other `{call, args}` → `@[fn(...)]@` expression
- `checks: [...]` on inputs → lifted into a synthesized `Flow.schema` via `synthesizeSchema` so Player's existing SchemaController/ValidationController pick them up unchanged
- Templated children: `{path, componentId}` blocks become indexed paths scoped to `<scope>._index_`
- Cycle detection throws on non-tree component graphs

### Platform Renders 

The A2UI v0.9.1 reference catalog — **16 assets** — implemented per platform: Row, Column, List, Text, Image, Icon, Divider, Card, Modal, Tabs, Button, TextField, CheckBox, Slider, DateTimeInput, ChoicePicker. Asset `type` strings are PascalCase to match the adapter output; transformed assets consume the shared `run()`/`set()`/`currentValue` helpers.

- **React: `@player-ui/a2ui-plugin-react`** — the full catalog as React components.
- **Android: `plugins/a2ui/android`** — `A2UIPlugin : AndroidPlayerPlugin, JSPluginWrapper by <jvm wrapper>` registering the catalog as Jetpack Compose renderers. Function helpers decode as Kotlin function types.
- **iOS: `plugins/a2ui/swiftui`** — `A2UIPlugin : JSBasePlugin, NativePlugin` that loads the bundle and registers the catalog as SwiftUI renderers. Function helpers decode as Swift `WrappedFunction`s.
- **JVM: `plugins/a2ui/jvm`** — generated Kotlin `JSPluginWrapper` (`A2UIPlugin`) that loads the A2UI JS bundle (mirrors `reference-assets/jvm`). Headless-capable, no UI layer.

## Packages

A new folder in the repo! The goal for this folder is to ship preconfigured Player entrypoints so consumers don't have to assemble plugins themselves. The `plugins/` directory exports building blocks; `packages/` exports ready-to-use Players for specific content formats.

Each preset comes with the A2UI plugin pre-added and appends any consumer-supplied plugins after it. Since `HeadlessPlayer`/`AndroidPlayer` are `final`, the JVM/Android entries are idiomatic factory functions; iOS is a thin `View` wrapper that also defaults `startOptions: .a2ui`.

### React: `@player-ui/a2ui`
```typescript
import { A2UIReactPlayer } from "@player-ui/a2ui";
```

### JVM: `packages/a2ui/jvm`
```kotlin
val player = A2UIHeadlessPlayer()
player.start(snapshot, "a2ui")
```

### Android: `packages/a2ui/android`
```kotlin
val player = A2UIAndroidPlayer()
player.start(snapshot, "a2ui")
```

### iOS: `packages/a2ui/swiftui`
```swift
A2UISwiftUIPlayer(flow: snapshot, result: $result)
```

## Mocks

The canonical catalog is `plugins/a2ui/mocks` (**21 snapshots**). JS consumes it via `@player-ui/a2ui-plugin-mocks`; JVM/Android via `//tools/mocks:jar`. iOS mirrors the full set as faithful inline copies (the existing iOS demo/test convention, there is no runtime mechanism to read the canonical JSON on iOS today).

## Storybook

New A2UI stories with one story per asset, driven by a `createA2UIStory` helper in the Player storybook extension that can render A2UI content.

## Demo Apps

- **iOS** — an "A2UI" screen renders the full canonical snapshot catalog, each started with `.a2ui`.
- **Android** — the demo `PlayerViewModel` registers `A2UIPlugin` alongside the reference assets (disjoint type namespaces).

## Notes

- Android `Image`/`Icon` (no Coil / `material-icons-core` only) and iOS `Image` (deployment target iOS 14 predates `AsyncImage`) render labelled placeholders; iOS `Icon` uses SF Symbols.

### Overall Approach
- Do we want to have this switch inside of Player or do we want to offer two exports from core, one per input format, to minimize bundle size

### Change Type (required)

- [ ] `patch`
- [x] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

## Release Notes
 TBD

<details>
  <summary>📦 Published PR as canary version: <code>1.1.0--canary.866.38904</code></summary>
  <br />

  Try this version out locally by upgrading relevant packages to 1.1.0--canary.866.38904
</details>


intuit-svc:
## Docs Preview

A preview of your PR docs was deployed by CircleCI [#36492](https://circleci.com/gh/player-ui/player/36492) on `Tue, 12 May 2026 07:39:49 GMT`

### 📖 Docs ([View site](https://player-ui.github.io/pr/866/))

intuit-svc:
## Build Preview

Your PR was deployed by CircleCI [#38904](https://circleci.com/gh/player-ui/player/38904) on `Thu, 09 Jul 2026 01:07:21 GMT` with this version:

```
1.1.0--canary.866.38904
```

### 📖 Docs ([View site](https://player-ui.github.io/pr/866/))

intuit-svc:
## Benchmark Results

_Comparison against baseline from `main`. ⚠️ = regression (>10% slower), ✅ = improvement (>5% faster)_

### `core/player` ⚠️

| Benchmark | Current | Baseline | Change |
|-----------|---------|----------|--------|
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.bar` | 642.36K ops/s | 588.82K ops/s | +9.1% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets.1.name` | 432.37K ops/s | 405.31K ops/s | +6.7% ✅ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets.01.name` | 376.93K ops/s | 471.42K ops/s | -20.0% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets['01'].name` | 384.67K ops/s | 470.61K ops/s | -18.3% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets[01].name` | 348.26K ops/s | 445.26K ops/s | -21.8% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets[name = "frodo"].type` | 218.48K ops/s | 243.29K ops/s | -10.2% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets["name" = "sprinkles"].type` | 187.90K ops/s | 221.16K ops/s | -15.0% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets["isDog" = false].type` | 216.52K ops/s | 295.56K ops/s | -26.7% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > parser benchmarks > Resolving binding: foo.pets["isDog" = true].type` | 224.13K ops/s | 308.99K ops/s | -27.5% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.bar` | 419.89K ops/s | 610.26K ops/s | -31.2% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets.1.name` | 312.43K ops/s | 405.96K ops/s | -23.0% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets.01.name` | 239.24K ops/s | 393.83K ops/s | -39.3% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets['01'].name` | 264.01K ops/s | 341.83K ops/s | -22.8% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets[01].name` | 266.15K ops/s | 314.81K ops/s | -15.5% ⚠️ |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets[name = "frodo"].type` | 232.64K ops/s | 234.36K ops/s | -0.7% |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets["name" = "sprinkles"].type` | 171.20K ops/s | 189.09K ops/s | -9.5% |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets["isDog" = false].type` | 216.68K ops/s | 240.22K ops/s | -9.8% |
| `core/player/src/binding/__tests__/parser.bench.ts > binding creation benchmarks > Resolving binding: foo.pets["isDog" = true].type` | 239.39K ops/s | 246.26K ops/s | -2.8% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = 1 + 3 (sync)` | 426.02K ops/s | 449.48K ops/s | -5.2% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = 1 + 3 (async)` | 305.90K ops/s | 362.75K ops/s | -15.7% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: conditional(true, true, false) (sync)` | 372.92K ops/s | 439.45K ops/s | -15.1% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: conditional(true, true, false) (async)` | 414.39K ops/s | 308.90K ops/s | +34.2% ✅ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional({{bar}} > 0, true, false) (sync)` | 197.04K ops/s | 171.19K ops/s | +15.1% ✅ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional({{bar}} > 0, true, false) (async)` | 175.58K ops/s | 172.45K ops/s | +1.8% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional(conditional(true = false, false, true), conditional(false = false, true, false), conditional(true = true, false, true)) (sync)` | 141.79K ops/s | 154.78K ops/s | -8.4% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional(conditional(true = false, false, true), conditional(false = false, true, false), conditional(true = true, false, true)) (async)` | 152.34K ops/s | 106.34K ops/s | +43.3% ✅ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = await(asyncTestFunction(1)) (sync)` | N/A | N/A | N/A |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = await(asyncTestFunction(1)) (async)` | 288.36K ops/s | 243.43K ops/s | +18.5% ✅ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = asyncTestFunction(1) (sync)` | 322.27K ops/s | 227.29K ops/s | +41.8% ✅ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = asyncTestFunction(1) (async)` | 207.69K ops/s | 242.15K ops/s | -14.2% ⚠️ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: asyncTestFunction(1) (sync)` | 815.81K ops/s | 698.81K ops/s | +16.7% ✅ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: asyncTestFunction(1) (async)` | 513.46K ops/s | 532.75K ops/s | -3.6% |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional(!{{bar}} == false, await(asyncTestFunction(1)), false) (sync)` | 183.16K ops/s | 168.02K ops/s | +9.0% ✅ |
| `core/player/src/expressions/__tests__/performance.bench.ts > Expression Parsing/Execution Benchmark > Parsing: {{foo}} = conditional(!{{bar}} == false, await(asyncTestFunction(1)), false) (async)` | 113.92K ops/s | 153.84K ops/s | -26.0% ⚠️ |
| `core/player/src/view/resolver/__tests__/index.bench.ts > resolver benchmarks > initial resolve` | 543.81 ops/s | 556.80 ops/s | -2.3% |
| `core/player/src/view/resolver/__tests__/index.bench.ts > resolver benchmarks > Resolving from cache` | 13.51K ops/s | 13.26K ops/s | +1.8% |
| `core/player/src/view/resolver/__tests__/index.bench.ts > resolver benchmarks > data changes` | 2.21K ops/s | 2.33K ops/s | -5.1% |
| `core/player/src/view/resolver/__tests__/index.bench.ts > resolver benchmarks > data changes slow` | 587.28 ops/s | 465.52 ops/s | +26.2% ✅ |

### `plugins/async-node/core` ⚠️

| Benchmark | Current | Baseline | Change |
|-----------|---------|----------|--------|
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 1 times` | 13.41K ops/s | 13.42K ops/s | -0.0% |
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 5 times` | 13.27K ops/s | 13.20K ops/s | +0.5% |
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 10 times` | 9.65K ops/s | 9.64K ops/s | +0.1% |
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 50 times` | 2.82K ops/s | 3.19K ops/s | -11.5% ⚠️ |
| `plugins/async-node/core/src/__tests__/index.bench.ts > async node benchmarks > Resolve Async Node 100 times` | 1.66K ops/s | 1.78K ops/s | -6.8% |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 1 times` | 6.00K ops/s | 6.13K ops/s | -2.1% |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 5 times` | 7.16K ops/s | 6.49K ops/s | +10.4% ✅ |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 10 times` | 5.86K ops/s | 6.34K ops/s | -7.5% |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 50 times` | 2.30K ops/s | 2.60K ops/s | -11.8% ⚠️ |
| `plugins/async-node/core/src/__tests__/transform.bench.ts > async transform benchmarks > Resolve Async Node 100 times` | 1.48K ops/s | 1.60K ops/s | -7.1% |

### `react/player` ⚠️

| Benchmark | Current | Baseline | Change |
|-----------|---------|----------|--------|
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 1 ReactAssets` | 543.91 ops/s | 631.66 ops/s | -13.9% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 1 ReactAssets` | 787.06 ops/s | 1.12K ops/s | -30.0% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 5 ReactAssets` | 575.28 ops/s | 634.01 ops/s | -9.3% |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 5 ReactAssets` | 867.34 ops/s | 1.03K ops/s | -16.1% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 10 ReactAssets` | 504.97 ops/s | 640.33 ops/s | -21.1% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 10 ReactAssets` | 707.88 ops/s | 908.95 ops/s | -22.1% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 50 ReactAssets` | 413.47 ops/s | 521.80 ops/s | -20.8% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 50 ReactAssets` | 211.57 ops/s | 271.20 ops/s | -22.0% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Render asset nested in 100 ReactAssets` | 328.12 ops/s | 422.33 ops/s | -22.3% ⚠️ |
| `react/player/src/asset/__tests__/index.bench.tsx > ReactAsset benchmarks > Bubble errors nested in 100 ReactAssets` | 89.75 ops/s | 120.76 ops/s | -25.7% ⚠️ |

MODULE.bazel
```diff
@@ -75,8 +75,7 @@ npm.npm_translate_lock(
     ],
     npm_package_target_name = "{dirname}",
     npmrc = "//:.npmrc",
-    pnpm_lock = "//:pnpm-lock.yaml",
-    verify_node_modules_ignored = "//:.bazelignore",
```
KVSRoyal, line 83: What does removing `verify_node_modules_ignored` do?

plugins/a2ui/swiftui/Sources/A2UIPlugin.swift
```diff
@@ -0,0 +1,70 @@
+import SwiftUI
+
+#if SWIFT_PACKAGE
+import PlayerUI
+import PlayerUISwiftUI
+#endif
```
KVSRoyal: We can drop the conditional wrapper nowadays since we only release through Swift packages now. Can you make that change? Same for the other files like this.

```swift
// Just this now
import SwiftUI
import PlayerUI
import PlayerUISwiftUI
```

plugins/a2ui/swiftui/Sources/A2UIPlugin.swift
```diff
@@ -0,0 +1,70 @@
...
+        ResourceUtilities.urlForFile(name: fileName, ext: "js", bundle: Bundle.module)
+        #else
+        ResourceUtilities.urlForFile(
+            name: fileName,
+            ext: "js",
+            bundle: Bundle(for: A2UIPlugin.self),
+            pathComponent: "PlayerUI_A2UI.bundle"
+        )
+        #endif
+    }
```
KVSRoyal: Same as above. We can just keep the if part now:

```swift
override open func getUrlForFile(fileName: String) -> URL? {
    ResourceUtilities.urlForFile(name: fileName, ext: "js", bundle: Bundle.module)
}
```

KVSRoyal (commented):
Just some small questions. But otherwise this looks good from an ios perspective.

MODULE.bazel
KetanReddy: So before, we had to manually specify every `node_module` folder for Bazel to ignore to prevent loops in the dependency tree. This PR migrates away from that pattern to specify to just ignore all `node_modules` in the `REPO.bazel` flle. That way when we make a new JS based package we don't have to add it to that list.

packages/a2ui/react/src/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+ *
+ *   const player = new A2UIReactPlayer();
+ *   await player.start(snapshot, { format: "a2ui", version: "0.9" });
+ */
+export class A2UIReactPlayer extends ReactPlayer {
+  constructor(options: A2UIReactPlayerOptions = {}) {
+    super({
+      ...options,
+      plugins: [
+        new A2UIPlugin() as unknown as ReactPlayerPlugin,
```
tmarmer: This type cast shouldn't be needed afaik. If there is an issue where it's required here I think we should try to tackle that. All `PlayerPlugin`s should be acceptable here.

plugins/a2ui/core/src/plugins/a2ui-content-plugin.ts
```diff
@@ -0,0 +1,27 @@
...
+export class A2UIContentPlugin implements PlayerPlugin {
+  name = "a2ui-content";
+
+  apply(player: Player): void {
+    player.hooks.transformContent.tap("a2ui-content", (content, meta) => {
+      if (meta.format !== "a2ui") return content;
+      return adaptA2UIToFlow(
+        content as A2UISnapshot,
+        (player as unknown as { logger: Parameters<typeof adaptA2UIToFlow>[1] })
+          .logger,
```
tmarmer: Could we just use `player.logger` here?

tools/storybook/src/a2ui/createA2UIStory.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * a separate build step.
+ */
+export function createA2UIStory(loader: A2UIStoryLoader, options?: any) {
+  const Comp = () => (
+    <PlayerStory
+      flow={async () => {
+        const mod = await loader();
+        const snap =
+          (mod as { default?: A2UISnapshot }).default ?? (mod as A2UISnapshot);
+        return { default: snap as unknown as Record<string, unknown> };
```
tmarmer:
```suggestion
        const snap = "default" in mod ? mod.default : mod;
        return { default: snap };
```
This change will introduce a type issue because `snap` as an `A2UISnapshot` doesn't match the `Mock` type, but I think if mocks can be anything at this point we may just want to make that type `object`, `unknown`, or `Record<string, unknown>`.

tools/storybook/src/a2ui/createA2UIStory.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import type { A2UISnapshot } from "@player-ui/a2ui-plugin";
```
tmarmer, line 2: This dependency has to be added to the `package.json` or `BUILD` for this package.

tools/storybook/src/player/PlayerStory.tsx
```diff
@@ -103,9 +114,12 @@ const PlayerJsonEditorStory = () => {
...
     setTrackedBeacons([]);
 
-    setKey(jsonEditorValue.value.id);
+    setKey(
+      (jsonEditorValue.value as { id?: string; surfaceId?: string }).id ??
+        (jsonEditorValue.value as { surfaceId?: string }).surfaceId,
+    );
 
-    wp.start(jsonEditorValue.value)
+    wp.start(jsonEditorValue.value as never, format ? { format } : undefined)
```
tmarmer:
```suggestion
    wp.start(jsonEditorValue.value, format ? { format } : undefined)
```

there are a few `as never` in `adapter.ts` as well that need to be removed

core/player/src/player.ts
```diff
@@ -513,8 +516,16 @@ export class Player {
...
-    const ref = Symbol(payload?.id ?? "payload");
+  public async start(
+    payload: unknown,
+    options?: StartOptions,
+  ): Promise<CompletedState> {
+    const meta: ContentMeta = {
+      format: options?.format ?? "player",
+      version: options?.version,
+    };
+    const flow = this.hooks.transformContent.call(payload, meta) as Flow;
```
tmarmer: If I understand correctly, the reason for transforming the content at this level is to ensure we have all the properties that `Flow` has (namely things like `id` and `navigation`) but is this appropriate for all parts of the object?

One consideration I have here is that support for different formats for something like the `views` should happen at the parser level, since all that really matters for the view/assets is the AST that gets produced. This is also something we need to think about with how this will interact with the `AsyncNodePlugin` since the A2UI format supports dynamic content updates. We need to parse new content into an AST but if we have to follow the same step of A2UI > `Flow` > AST it feels like we'll just end up with extra processing when we could just go straight to the AST.

core/player/src/types.ts
```diff
@@ -47,6 +77,17 @@ export interface PlayerHooks {
...
+   * Transform raw input content into a Player `Flow` before any state is set
+   * up. Fires at the top of `Player.start()` — after plugins are applied,
+   * before `resolveFlowContent`. Plugins tap this and inspect `meta.format`
+   * (and optionally `meta.version`) to decide whether to convert. Plugins
+   * that don't handle the format pass the content through unchanged.
+   */
+  transformContent: SyncWaterfallHook<
+    [unknown, ContentMeta],
+    Record<string, any>
+  >;
```
tmarmer: I think a `Bail` hook here is more appropriate than a `Waterfall` hook as it would let us require the output type to be a `Flow`. We could then tap in ourselves and for `player` format we can just return the object as it is. This would let us remove the type-casting from the `Player` `start` function which imo is error-prone.

tmarmer (changes requested):

ios/core/Sources/Player/StartOptions.swift
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ Options forwarded to the core JS `Player.start(payload, options)` describing the
+ incoming content's format. Mirrors the JS `StartOptions`/`ContentMeta` shape.
+
+ The default format is `"player"`, meaning the payload is already a Player `Flow`
+ and needs no conversion. Any other `format` is recognized by a plugin tapping the
+ core `transformContent` hook (e.g. the A2UI plugin claims `"a2ui"`).
+ */
+public struct StartOptions {
```
JunDangIntuit: Could we make StartOptions conform to Equatable and Sendable?  Equatable allows us to write sth like ==, and Sendable for the thread safety

ios/swiftui/Sources/ManagedPlayer/ManagedPlayer.swift
```diff
@@ -30,21 +30,28 @@ public struct ManagedPlayer<Loading: View, Fallback: View>: View {
...
+     Creates a `ManagedPlayer`
+     - parameters:
+        - plugins: The plugins to use for the `SwiftUIPlayer`
+        - viewModel: The `ManagedPlayerViewModel` to use for fetching flows
+        - handleScroll: Whether or not the `ManagedPlayer` should wrap content in a `ScrollView`
+        - startOptions: Describes the content `format`/`version` for every flow in the session
+          (defaults to `nil`, i.e. Player `Flow`s). Pass e.g. `.a2ui` for an A2UI experience.
+        - onError: A handler for when the `SwiftUIPlayer` encounters an error
+        - loading: A closure providing a `View` to display while the `FlowManager` fetches flows
+     */
```
JunDangIntuit: May we use /// instead of /** */ for iOS doc purpose. had a discussion previously and made an agreement to follow this guide https://google.github.io/swift/

Image

ios/swiftui/Sources/ManagedPlayer/ManagedPlayer.swift
```diff
@@ -110,18 +121,24 @@ internal struct ManagedPlayer14<Loading: View, Fallback: View>: View {
...
+    /**
+     Creates a `ManagedPlayer`
+     - parameters:
+        - viewModel: The `ManagedPlayerViewModel` to use for fetching flows
+        - plugins: The plugins to use for the `SwiftUIPlayer`
+        - handleScroll: Whether or not the `ManagedPlayer` should wrap content in a `ScrollView`
+        - startOptions: Describes the content `format`/`version` for every flow in the session
+        - onError: A handler for when the `SwiftUIPlayer` encounters an error
+        - loading: A closure providing a `View` to display while the `FlowManager` fetches flows
+     */
```
JunDangIntuit: same here to replace /** */ with ///

ios/swiftui/Sources/ManagedPlayer/ManagedPlayer.swift
```diff
@@ -110,18 +121,24 @@ internal struct ManagedPlayer14<Loading: View, Fallback: View>: View {
...
     public init(
```
JunDangIntuit: this should not be public since ManagedPlayer14 is internal. It was made previously though.

ios/swiftui/Sources/ManagedPlayer/ManagedPlayer.swift
```diff
@@ -110,18 +121,24 @@ internal struct ManagedPlayer14<Loading: View, Fallback: View>: View {
...
+    private var startOptions: StartOptions?
+
+    /**
+     Creates a `ManagedPlayer`
+     - parameters:
+        - viewModel: The `ManagedPlayerViewModel` to use for fetching flows
+        - plugins: The plugins to use for the `SwiftUIPlayer`
+        - handleScroll: Whether or not the `ManagedPlayer` should wrap content in a `ScrollView`
+        - startOptions: Describes the content `format`/`version` for every flow in the session
+        - onError: A handler for when the `SwiftUIPlayer` encounters an error
```
JunDangIntuit: not from this pr, but seems there is no onError parameter in the init

KVSRoyal (dismissed):
Lgtm from an ios perspective once it passes the pipeline

ios/core/Sources/Player/HeadlessPlayer.swift
```diff
@@ -246,6 +246,18 @@ public extension HeadlessPlayer {
...
+    }
+
+    /**
+     Starts Player for the given flow, declaring its content format/version
+     - parameters:
+        - flow: The content payload (a Player `Flow`, or another format claimed by a plugin)
+        - options: Describes the content `format`/`version`. When `nil` (or `format == "player"`)
+          the payload is treated as a Player `Flow`. Forwarded to the JS `start(payload, options)`.
+        - completion: A completion handler for when the flow has completed
+     */
```
JunDangIntuit: oh, i might forgot to comment on all places. but may we change all /** */ style to /// for all iOS file in this pr?

KetanReddy: Surprised this is passing swift lint. Are we missing some rules?

KVSRoyal: Yeah. I can actually start working on that again now! https://github.com/player-ui/player/pull/854

plugins/a2ui/core/src/a2ui/adapter.ts
```diff
@@ -214,7 +221,18 @@ function translatePropValue(
...
+  scope: string,
+): A2UIFunctionCall;
+function rewritePathsInCall(
+  value: A2UIDynamicValue,
+  scope: string,
+): A2UIDynamicValue;
+function rewritePathsInCall(
+  value: A2UIDynamicValue,
+  scope: string,
+): A2UIDynamicValue {
```
tmarmer: Not a fan of using function overloading here since the TS compiler doesn't enforce that the implementation actually matches the signatures here. Can we separate the functions such that `rewritePathsInCall` covers `A2UIDynamicValue` as it already does, and have another function for just the `A2UIFunctionCall` that is used in this one?

tmarmer (changes requested):

plugins/a2ui/android/src/main/kotlin/com/intuit/playerui/android/a2ui/A2UIPlugin.kt
```diff
@@ -0,0 +1,45 @@
...
+ * and registers the matching Compose assets onto the [AndroidPlayer].
+ *
+ * Start A2UI content with `player.start(snapshot, "a2ui")`.
+ */
+public class A2UIPlugin :
+    AndroidPlayerPlugin,
+    JSPluginWrapper by Transforms() {
+    /** register the A2UI assets — type strings are PascalCase to match the adapter output */
+    override fun apply(androidPlayer: AndroidPlayer) {
+        androidPlayer.registerAsset("Row", ::Row)
```
brocollie08, line 24: Do we need new assets for it? Is the purpose to give a meaningful demo? These look usable for reference assets too

packages/a2ui/jvm/src/main/kotlin/com/intuit/playerui/a2ui/A2UIHeadlessPlayer.kt
```diff
@@ -0,0 +1,36 @@
...
+ * @example
+ * ```kotlin
+ * val player = A2UIHeadlessPlayer()
+ * player.start(snapshot, "a2ui").await()
+ * ```
+ */
+@OptIn(ExperimentalPlayerApi::class)
+@JvmOverloads
+@Suppress("ktlint:standard:function-naming")
+public fun A2UIHeadlessPlayer(
```
brocollie08, line 27: hmmm where is this going to get used, is there a config that says when to use this vs when to use regular `HeadlessPlayer` when creating an Android Player?

KetanReddy: The idea here is to start offering more off the shelf configured versions of Player. With this, if you know you want to use A2UI with player you should just be able to start with this package and have something production grade.

plugins/a2ui/swiftui/Sources/SwiftUI/A2UIIconAsset.swift
```diff
@@ -0,0 +1,29 @@
...
+    public override var view: AnyView { AnyView(A2UIIconAssetView(model: model)) }
+}
+
+struct A2UIIconAssetView: View {
+    @ObservedObject var model: AssetViewModel<A2UIIconData>
+
+    var body: some View {
+        let name = model.data.name?.stringValue
+        // Pass the name through as an SF Symbol; fall back to a neutral glyph.
+        Image(systemName: (name?.isEmpty == false ? name! : "questionmark.circle"))
```
JunDangIntuit: Found another one, may we not to use force unwrap? 
change it to
    let name = model.data.name?.stringValue ?? ""
        // Pass the name through as an SF Symbol; fall back to a neutral glyph.
        Image(systemName: name.isEmpty ? "questionmark.circle" : name)

plugins/a2ui/swiftui/Sources/SwiftUI/A2UIDividerAsset.swift
```diff
@@ -0,0 +1,31 @@
...
+
+    var body: some View {
+        if model.data.axis == "vertical" {
+            Divider()
+                .frame(maxHeight: .infinity)
+                .accessibility(identifier: model.data.id)
+        } else {
+            Divider()
+                .accessibility(identifier: model.data.id)
+        }
```
JunDangIntuit: nit: nothing wrong, just for convention consistency, may we write it as
 Divider()
            .frame(maxHeight: model.data.axis == "vertical" ? .infinity : nil)
            .accessibility(identifier: model.data.id)

plugins/a2ui/android/src/main/kotlin/com/intuit/playerui/android/a2ui/A2UIPlugin.kt
KetanReddy: Basically yeah, so that there is an off the shelf way for you to use A2UI with Player. Most other renderers have similar OTS assets.

JunDangIntuit (dismissed):

plugins/a2ui/swiftui/Sources/SwiftUI/A2UIDividerAsset.swift
JunDangIntuit: Hi @KetanReddy, I approved the pr. but may you also address this since it aligns better with iOS best practice. thanks!

tmarmer (dismissed):

JunDangIntuit (dismissed):

core/player/src/player.ts
KetanReddy: Correct, right now we are just introducing a non-breaking way to support multiple different types of input formats. To your second point its definitely not perfect right now in terms of fully leveraging all of Player but I think we'll have to do a bit more exploration on to figure out what the target state is for this. Specifically for A2UI I don't know if async node can fully cover the operations required because UI elements might be removed in an update.

brocollie08:
i'd be okay with this PR assuming 2 points

1. The A2UIPlayer is the pattern we want to proceed with "packaged x-player"
2. There's a plan going forward for asset organization (OOB A2UI assets vs reference assets)

KetanReddy:
> i'd be okay with this PR assuming 2 points
> 
>     1. The A2UIPlayer is the pattern we want to proceed with "packaged x-player"
> 
>     2. There's a plan going forward for asset organization (OOB A2UI assets vs reference assets)

I'll leave it up to @player-ui/reviewers, does anyone feel strongly about another naming scheme? For the Assets, let me start a discussion post about it so we can figure out what we want to do. 

https://github.com/orgs/player-ui/discussions/903

JunDangIntuit (approved):

tmarmer (approved):

brocollie08 (approved):

PR #868, opened 2026-05-16
Handle `external-state` plugin errors with `ErrorController`

KVSRoyal:
Now that `ErrorController` is available, we use that to handle errors from the `external-state` plugin instead of letting it fall through. See included docs updates for more details.

### Change Type (required)

- [ ] `patch`
- [ ] `minor`
- [ ] `major`
- [x] `N/A`. Part of 1.0

- [x] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs


plugins/external-state/core/src/__tests__/index.test.ts
```diff
@@ -439,8 +471,29 @@ describe("edge cases", () => {
     });
   });
 
-  test("no handler registered for external state - no transition occurs", async () => {
-    // Create a player with NO handlers registered
+  test("no handler registered for external state - calls captureError with ExternalStateError", async () => {
```
sugarmanz, line 593: Can we get some tests to validate how you'd handle this error by the error controller? My understanding is there are two ways, one via content and another by tapping `onError`. I think these tests would be valid for each platform, at least the `onError` ones since external actions are closer to integrations.

plugins/external-state/core/src/index.ts
```diff
@@ -88,26 +102,39 @@ export class ExternalStatePlugin implements PlayerPlugin {
...
-                } else {
-                  player.logger.warn(
-                    `External state resolved with [${transitionValue}], but Player already navigated away from [${toState.name}]`,
-                  );
-                }
+              if (!transitionValue) {
+                this.errorController?.captureError(
+                  ExternalStateError.missingTransitionValue(toState.value.ref),
+                );
+                return;
```
sugarmanz: How should the error be handled if we there isn't an `errorController`? Right now it'd fail silently potentially leaving the user stuck with a blank or stale screen. It'd likely be worth logging up front so that integrators know when this happens, or maybe a check at the start to ensure we have what we need to continue, like the `InProgress` state check below.

plugins/external-state/core/src/ExternalStateError.ts
```diff
@@ -0,0 +1,45 @@
...
+  ref: string;
+  /** Which failure mode this error represents */
+  reason: ExternalStateErrorReason;
+}
+
+export class ExternalStateError
+  extends Error
+  implements PlayerErrorMetadata<ExternalStateErrorMetadata>
+{
+  readonly type: string = "EXTERNAL-STATE";
```
sugarmanz: Can we follow the same convention for types as the [primary type definition](https://github.com/player-ui/player/blob/d09c97276ded80ee1b5d5ed0ff3378630707f3f2/core/player/src/controllers/error/types.ts#L8-L21), i.e. lowercase?

plugins/external-state/core/src/index.ts
KVSRoyal: Good call. Adding

plugins/external-state/core/src/ExternalStateError.ts
```diff
@@ -0,0 +1,45 @@
...
+  ref: string;
+  /** Which failure mode this error represents */
+  reason: ExternalStateErrorReason;
+}
+
+export class ExternalStateError
+  extends Error
+  implements PlayerErrorMetadata<ExternalStateErrorMetadata>
+{
+  readonly type: string = "externalState";
```
JunDangIntuit: May we add this to ErrorTypes so it's the single source of truth. for example: export const ErrorTypes = {
  EXTERNAL_STATE:  "externalState",
}
Drop the : string annotation: readonly type = ErrorTypes.EXTERNAL_STATE. Since ErrorTypes is as const, the field infers the literal "external-state".  Because string widens it and breaks narrowing — TS can't tell this error apart from others whose type is also typed as string.

KVSRoyal: Will do! Good call

KVSRoyal: Er, except I probably will keep string since [that's the pattern used in other places](https://github.com/player-ui/player/blob/866f2361d8f4cc301eae89b8e4379a0a857575da/core/player/src/view/resolver/ResolverError.ts#L11).

.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ orbs:
 executors:
   base:
     docker:
-      - image: docker.io/playerui/bazel-docker:13
+      - image: docker.io/playerui/bazel-docker:20
```
brocollie08: this can probably use the latest? i think the PR got merged

plugins/external-state/jvm/src/test/kotlin/com/intuit/playerui/plugins/externalState/ExternalStatePluginTest.kt
```diff
@@ -169,4 +174,116 @@ internal class ExternalStatePluginTest : PlayerTest() {
...
+      }
+    }
+  }
+}
+        """
+
+    @TestTemplate
+    fun testMissingHandlerNavigatesViaContentErrorTransitions() = runBlockingTest {
+        val plugin = ExternalStatePlugin(
+            ExternalStateHandler(ref = "different-ref") { _, _, transition ->
```
brocollie08, line 213: hmmmm is having no handler expected to go into an error state? Do we not get a bunch of no-ops if there's no handler found for the ref in the registry? Could you show me where it errors out

KVSRoyal: Hi! It goes into an error state, yeah. Here is where [the core code errors out](https://github.com/player-ui/player/pull/868/changes#diff-607fda439a5957eb93b627af1fec24fa45e7c65024e900071483691e2a722b94R123-R129).

JunDangIntuit (approved):

.circleci/config.yml
KVSRoyal: Done

brocollie08 (approved):

PR #907, opened 2026-07-20
Fix cancellation swallowed as AssetRenderException in RenderableAsset render()

A1shK:
**AI generated summary**

render()'s outer catch didn't special-case CancellationException the way[ doRender()'s inner catch already does](https://github.com/player-ui/player/pull/907/changes#diff-a7eb693581e56fa61f8e3612a94fd2e2881e2dd1482858ea81d128e21e0fd785R161). When a parent rehydrates while an in-flight inflateChild() render() call for a child asset is still suspended, the parent's renewHydrationScope() cancellation gets wrapped into AssetRenderException("Failed to render asset") instead of propagating as-is, escaping inflateChild()'s unguarded launch and failing the whole flow.

Includes an isolated repro test demonstrating the race and confirming the fix.

**Human explanation**

We are seeing a few flakes in our tests with errors saying " failed to render asset" and this is one theory that  i want to test out.  locally it has reduced some flakes but wanted to test it in a canary to actually confirm that. 

### Change Type (required)

- [x] `patch`
- [ ] `minor`
- [ ] `major`
- [ ] `N/A`

- [ ] Updated docs
- [ ] No Update needed
- [ ] Unable to update docs

### DIFF
--- a/android/player/src/main/kotlin/com/intuit/playerui/android/asset/RenderableAsset.kt
+++ b/android/player/src/main/kotlin/com/intuit/playerui/android/asset/RenderableAsset.kt
@@ -121,12 +121,12 @@ public abstract class RenderableAsset<Data>(
                         player.asyncHydrationTrackerPlugin?.trackHydration(this@RenderableAsset)
                         try {
                             rehydrate(cachedView)
-                            player.asyncHydrationTrackerPlugin?.renderingComplete(this@RenderableAsset)
                             cachedView
                         } catch (exception: StaleViewException) {
-                            player.asyncHydrationTrackerPlugin?.renderingComplete(this@RenderableAsset)
                             renewHydrationScope("recreating after stale rehydrate")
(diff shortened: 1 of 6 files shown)


android/player/src/main/kotlin/com/intuit/playerui/android/asset/RenderableAsset.kt
```diff
@@ -138,6 +138,7 @@ public abstract class RenderableAsset<Data>(
                 }
             }.also { player.cacheAssetView(assetContext, it) }
     } catch (exception: Throwable) {
+        if (exception is CancellationException) throw exception
```
A1shK, line 141: copying [this](https://github.com/player-ui/player/pull/907/changes#diff-a7eb693581e56fa61f8e3612a94fd2e2881e2dd1482858ea81d128e21e0fd785R161)

